PR #51, opened 2023-08-09
Add configurable storage and token endpoints

judemars:
Make storage and token server endpoints configurable

This involves a breaking change that disallows the user to configure the "endpoint" flag in gcsfuse. It can only be configured by the driver (via a driver flag)


judemars:
Ran `make e2e-test REGISTRY=judemars E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=true BUILD_GCSFUSE_FROM_SOURCE=true STAGINGVERSION=v999.999.999` locally and all tests pass :)

cmd/sidecar_mounter/main.go
```diff
@@ -35,9 +35,10 @@ import (
 )
 
 var (
-	gcsfusePath    = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
-	volumeBasePath = flag.String("volume-base-path", "/gcsfuse-tmp/.volumes", "volume base path")
-	gracePeriod    = flag.Int("grace-period", 30, "grace period for gcsfuse termination")
+	gcsfusePath     = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
+	volumeBasePath  = flag.String("volume-base-path", "/gcsfuse-tmp/.volumes", "volume base path")
+	gracePeriod     = flag.Int("grace-period", 30, "grace period for gcsfuse termination")
+	storageEndpoint = flag.String("storage-endpoint", "", "If set, used as the endpoint for the GCS API.")
```
songjiaxun: Because the storage endpoint will be passed from the CSI node server to the sidecar container, I don't think we need to pass this info via a flag to the sidecar mounter here.

Overall, I think the storage endpoint and token server endpoint information should be passed via the following path: GKE cluster -> CSI node server -> CSI mounter -> `sidecarmounter.MountConfig` -> sidecar mounter -> gcsfuse binaries. Unless we want the sidecar mounter to overwrite the values passed from the CSI node server, it is unnecessary to include this flag here.

cmd/sidecar_mounter/main.go
```diff
@@ -159,8 +160,9 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
 	dir := filepath.Dir(sp)
 	volumeName := filepath.Base(dir)
 	mc := sidecarmounter.MountConfig{
-		VolumeName: volumeName,
-		TempDir:    filepath.Join(dir, "temp-dir"),
+		VolumeName:      volumeName,
+		TempDir:         filepath.Join(dir, "temp-dir"),
+		StorageEndpoint: *storageEndpoint,
```
songjiaxun: Similar to the previous comment, the `sidecarmounter.MountConfig` is passed from the CSI mounter, and the storage endpoint info is already included in the `MountConfig`, I don't think we need this field added here.

cmd/csi_driver/main.go
```diff
@@ -32,14 +32,16 @@ import (
...
+	endpoint            = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	nodeID              = flag.String("nodeid", "", "node id")
+	runController       = flag.Bool("controller", false, "run controller service")
+	runNode             = flag.Bool("node", false, "run node service")
+	kubeconfigPath      = flag.String("kubeconfig-path", "", "The kubeconfig path.")
+	sidecarImage        = flag.String("sidecar-image", "", "The gcsfuse sidecar container image.")
+	identityPool        = flag.String("identity-pool", "", "The Identity Pool to authenticate with GCS API.")
+	identityProvider    = flag.String("identity-provider", "", "The Identity Provider to authenticate with GCS API.")
+	storageEndpoint     = flag.String("storage-endpoint", "", "If set, used as the endpoint for the GCS API.")
+	tokenServerEndpoint = flag.String("token-server-endpoint", "", "If set, used as the endpoint for the Token Server API.")
```
songjiaxun, line 44: If the token server endpoint is different, do we also want to pass the token server endpoint to gcsfuse binaries? The gcsfuse cli has a flag `--token-url` to accept different token server endpoints, see https://cloud.google.com/storage/docs/gcsfuse-cli

songjiaxun (commented):
Overall LGTM, but there are some details I'd like to discuss more.

cmd/csi_driver/main.go
judemars: Great point! I added that in pkg/sidecar_mounter/sidecar_mounter.go.

cmd/sidecar_mounter/main.go
judemars: Thanks for catching this! I've removed it from the flags (see cmd/sidecar_mounter/main.go is no longer in the PR)

cmd/sidecar_mounter/main.go
judemars: Thanks for catching this! I've removed it from the flags (see cmd/sidecar_mounter/main.go is no longer in the PR)

judemars:
Converting to draft for now as we've received new info that the gcsfuse compute endpoint isn't prod-ready, so deferring until we can get one in place.

Issue #57, opened 2023-08-24
emphermal storage for very large tar ~100GB


ashish01987:
I have a local folder (backup) of ~100GB files , if i directly tar the folder onto bucket e.g tar -cf /tmp/bucketmount/backup.tar /backup/ , will there be any issues with csi driver ? I see that gcsfuse csi depends on tempDir{} or some temp directore for staging files before they are uploaded to bucket.

ashish01987:
I have a local folder (backup) of  ~100GB files , if i directly tar the folder onto bucket e.g tar -cf /tmp/bucketmount/backup.tar /backup/ , will there be any issues with csi driver ? I see that gcsfuse csi depends on tempDir{} or some temp directore for staging files before they are uploaded to bucket.

songjiaxun:
I think CSI driver should work in this use case. @ashish01987 do you see any errors?

As you mentioned, gcsfuse uses a temp directory for staging files, as a result, please consider increasing the sidecar container ephemeral-storage-limit so that gcsfuse has enough space for staging the files.

See the [GKE documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/cloud-storage-fuse-csi-driver#sidecar-container) for more information.

ashish01987:
Thanks for the quick response. I tried out with 30GB files and did not see any issues with default allocated ephermeral storage to side.
Will probably try out with some more large values 60GB and 100GB of files archived into single tar on gcs bucket

ashish01987:
Thanks for the quick response. I created a tar file "backup.tar" of 30GB directly on the GCS mounted bucket (mounted by csi side car). And did not find any issues with it. Just one question here
When "backup.tar" (30GB) size is being created on the mounted bucket, will the csi side car wait for complete "backup.tar" (30GB) file being created on ephemeral storage (emptyDir: {}) and then copy it to actual bucket on cloud storage ?

If yes, i am bit concern about case where "backup.tar" size go on increasing (maybe 100GB or more due to regular backups) and sufficient node ephermeral storage is not available. In this case, one may have increase the nodes ephermeral storage manually which might cause downtime for cluster (probably ?) 

I see that csi side car uses this "gke-gcsfuse-tmp" mount point from emptyDir{} for staging  files before uploading

- emptyDir: {}
    name: gke-gcsfuse-tmp
    
[...]

PR #58, opened 2023-09-02
Add basic experimental cache support

judemars:
This CL adds simple support for utilizing the GCSFuse experimental cache. To do this, it detects whether the user has passed in experimental-local-file-cache to their mount options, and if so, adds an emptydir volume mount pointing to ephemeral local ssds (that have to have been previously set up by the user) and points the cache at them.

@songjiaxun please let me know if there's a suitable readme for me to add structions on how to add a cluster/node pool w/ local ssds.

### DIFF
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -31,6 +31,8 @@ import (
 	"time"
 
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
+	sidecarspec "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
+
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"k8s.io/klog/v2"
 )
@@ -192,7 +194,8 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
 
 	for _, opt := range mc.Options {
 		if strings.Contains(opt, "experimental-local-file-cache") {
-			mc.TempDir = "/cache/gcsfuse-tmp"
+
+			mc.TempDir = filepath.Join(sidecarspec.CacheVolumeMountPath, "gcsfuse-tmp")
 		}
 	}
 
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -190,7 +190,7 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
 		return nil, fmt.Errorf("failed to fetch bucket name from CSI driver")
 	}
 
-	for _, opt := range mc.Options{
+	for _, opt := range mc.Options {
 		if strings.Contains(opt, "experimental-local-file-cache") {
 			mc.TempDir = "/cache/gcsfuse-tmp"
 		}
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -25,13 +25,11 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strings"
 	"sync"
 	"syscall"
 	"time"
 
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
-	sidecarspec "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"k8s.io/klog/v2"
@@ -190,13 +188,6 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
 
 	if mc.BucketName == "" {
 		return nil, fmt.Errorf("failed to fetch bucket name from CSI driver")
-	}
-
-	for _, opt := range mc.Options {
-		if strings.Contains(opt, "experimental-local-file-cache") {
-
-			mc.TempDir = filepath.Join(sidecarspec.CacheVolumeMountPath, "gcsfuse-tmp")
-		}
 	}
 
 	return &mc, nil
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -30,7 +30,6 @@ import (
 	"time"
 
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
-
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"k8s.io/klog/v2"
 )
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -187,6 +188,12 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
 
 	if mc.BucketName == "" {
 		return nil, fmt.Errorf("failed to fetch bucket name from CSI driver")
+	}
+
+	for _, opt := range mc.Options{
+		if strings.Contains(opt, "experimental-local-file-cache") {
+			mc.TempDir = "/cache/gcsfuse-tmp"
+		}
 	}
 
 	return &mc, nil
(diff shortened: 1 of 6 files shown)


pkg/webhook/sidecar_spec.go
```diff
@@ -28,16 +28,18 @@ const (
 	SidecarContainerName            = "gke-gcsfuse-sidecar"
 	SidecarContainerVolumeName      = "gke-gcsfuse-tmp"
 	SidecarContainerVolumeMountPath = "/gcsfuse-tmp"
+	CacheVolumeName                 = "cache-volume"
```
songjiaxun: Can we rename the volume name to `gke-gcsfuse-cache-volume`, and change the mount path to `/gcsfuse-cache`? This is because we have to document this reserved volume so that users won't use them in other containers. I think "cache-volume" is very general and users may want to use this volume name for other containers.

pkg/webhook/sidecar_spec.go
```diff
@@ -74,15 +76,36 @@ func GetSidecarContainerSpec(c *Config) v1.Container {
...
+				EmptyDir: &v1.EmptyDirVolumeSource{},
+			},
 		},
 	}
+	if addCacheVolume {
+		toReturn = append(toReturn, v1.Volume{
+			Name: CacheVolumeName,
+			VolumeSource: v1.VolumeSource{
+				HostPath: &v1.HostPathVolumeSource{
+					Path: "/mnt/stateful_partition/kube-ephemeral-ssd",
```
songjiaxun: Could you add a comment here to mention that we assume the local ssd is enabled on the node, and this change is currently for testing?

pkg/webhook/mutatingwebhook.go
```diff
@@ -109,8 +109,25 @@ func (si *SidecarInjector) Handle(_ context.Context, req admission.Request) admi
...
 	// the gcsfuse sidecar container has to before the containers that consume the gcsfuse volume
-	pod.Spec.Containers = append([]corev1.Container{GetSidecarContainerSpec(configCopy)}, pod.Spec.Containers...)
-	pod.Spec.Volumes = append([]corev1.Volume{GetSidecarContainerVolumeSpec()}, pod.Spec.Volumes...)
+	useExperimentalLocalFileCache := false
+	hasCacheVolume := false
+	for _, v := range pod.Spec.Volumes {
+		if v.Name == CacheVolumeName {
+			hasCacheVolume = false
+		}
+		if v.CSI == nil || v.CSI.VolumeAttributes == nil {
```
songjiaxun: Does this mean for the read cache test, we only focus on the CSI inline volume?

I guess for PV/PVC usage, we will rely on the Pod annotations in the final design, is that correct?

songjiaxun: If the answers are yes, could you add comments as well? Or a TODO to remove this logic in the future?

pkg/webhook/mutatingwebhook.go
```diff
@@ -109,8 +109,25 @@ func (si *SidecarInjector) Handle(_ context.Context, req admission.Request) admi
 
 	klog.Infof("mutating Pod: Name %q, GenerateName %q, Namespace %q, CPU limit %q, memory limit %q, ephemeral storage limit %q", pod.Name, pod.GenerateName, pod.Namespace, configCopy.CPULimit.String(), configCopy.MemoryLimit.String(), configCopy.EphemeralStorageLimit.String())
 	// the gcsfuse sidecar container has to before the containers that consume the gcsfuse volume
-	pod.Spec.Containers = append([]corev1.Container{GetSidecarContainerSpec(configCopy)}, pod.Spec.Containers...)
-	pod.Spec.Volumes = append([]corev1.Volume{GetSidecarContainerVolumeSpec()}, pod.Spec.Volumes...)
+	useExperimentalLocalFileCache := false
+	hasCacheVolume := false
+	for _, v := range pod.Spec.Volumes {
+		if v.Name == CacheVolumeName {
+			hasCacheVolume = false
```
songjiaxun: I guess you mean `true` here?

cmd/sidecar_mounter/main.go
```diff
@@ -189,5 +192,12 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
 		return nil, fmt.Errorf("failed to fetch bucket name from CSI driver")
 	}
 
+	for _, opt := range mc.Options {
+		if strings.Contains(opt, "experimental-local-file-cache") {
```
songjiaxun: If we only change the temp path when the `experimental-local-file-cache` is passed, it will not help solving https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/issues/57, because for manual sidecar injection, the second cache volume may exist, we also want to change the temp dir value even without the `experimental-local-file-cache` option.

cmd/sidecar_mounter/main.go
```diff
@@ -189,5 +192,12 @@ func prepareMountConfig(sp string) (*sidecarmounter.MountConfig, error) {
...
+
+			mc.TempDir = filepath.Join(sidecarspec.CacheVolumeMountPath, "gcsfuse-tmp")
```
songjiaxun: One concern: does this work when there are multiple gcsfuse instances?

The previous design is that, for each gcsfuse process, it has a different temp dir. But with this change, I believe all the gcsfuse processes will share the same temp dir. I think this may introduce potential issues.

examples/README.md
```diff
@@ -186,3 +186,32 @@ kubectl port-forward jupyter-notebook-server 8888:8888
 # clean up
 kubectl delete -f ./examples/jupyter/jupyter-notebook-server.yaml
 ```
+
+## Jupyter Notebook Example (no experimental read cache)
```
songjiaxun: Do you mean "(with experimental read cache)"?

examples/README.md
```diff
@@ -186,3 +186,32 @@ kubectl port-forward jupyter-notebook-server 8888:8888
...
+
+```bash
+# 1. replace <bucket-name> with your pre-provisioned GCS bucket name
+GCS_BUCKET_NAME=your-bucket-name
+sed -i "s/<bucket-name>/$GCS_BUCKET_NAME/g" ./examples/jupyter/jupyter-experimental-readcache.yaml
+
+# 2. install a Jupyter Notebook server using experimental gcsfuse read cache
+kubectl apply -f ./examples/jupyter/jupyter-experimental-readcache.yaml
+
+# get service IPs
```
songjiaxun: Could you also add the sequence number for this and the following steps? Or just remove the numbers 1, 2 from the above steps?

examples/README.md
```diff
@@ -186,3 +186,32 @@ kubectl port-forward jupyter-notebook-server 8888:8888
 # clean up
 kubectl delete -f ./examples/jupyter/jupyter-notebook-server.yaml
 ```
+
+## Jupyter Notebook Example (no experimental read cache)
+
+#### Prerequisite
```
songjiaxun: The example uses GPU, so you will also need to enable GPU on the node pool if you are using a standard cluster.

You can refer to https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/tree/main/examples#prerequisites here.

examples/README.md
judemars: Woops, yup, fixed! Thanks

examples/README.md
judemars: Good idea! Added

examples/README.md
judemars: Fixed! Thanks

cmd/sidecar_mounter/main.go
judemars: Resolving as discussed offline. This feature no longer necessitates driver changes.

cmd/sidecar_mounter/main.go
judemars: Resolving as discussed offline. This feature no longer necessitates driver changes.

pkg/webhook/mutatingwebhook.go
judemars: Resolving as discussed offline. This feature no longer necessitates driver changes.

pkg/webhook/mutatingwebhook.go
judemars: Resolving as discussed offline. This feature no longer necessitates driver changes.

pkg/webhook/sidecar_spec.go
judemars: Resolving as discussed offline. This feature no longer necessitates driver changes.

pkg/webhook/sidecar_spec.go
judemars: Resolving as discussed offline. This feature no longer necessitates driver changes.

PR #114, opened 2023-11-08
Added unit tests for mutatingwebhook.go

hime:
Created unit tests for the mutatingwebhook.go, which houses the core logic for admission request handling.

### DIFF
--- a/pkg/webhook/mutatingwebhook_test.go
+++ b/pkg/webhook/mutatingwebhook_test.go
@@ -1,0 +1,171 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Copyright 2022 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+
+	v1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+func TestValidateMutatingWebhookResponse(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name             string
+		inputPod         *corev1.Pod
+		operation        v1.Operation
+		expectedResponse admission.Response
+	}{
+		{
+			name:             "Empty request test.",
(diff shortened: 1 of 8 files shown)


pkg/webhook/mutatingwebhook_test.go
```diff
@@ -0,0 +1,171 @@
...
+						"gke-gcsfuse/volumes": "true",
+					},
+				},
+			},
+			expectedResponse: admission.Allowed("The sidecar container was injected, no injection required."),
+		},
+		{
+			name:             "Injection successful test.",
+			operation:        v1.Create,
+			inputPod:         InputPodTest5(),
```
songjiaxun: Can we change these helper function names to more descriptive ones, for example, `validInputPod`. And also please make these functions private -- I don't think they will be used anywhere else.

pkg/webhook/mutatingwebhook_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		return nil
+	}
+
+	return raw
+}
+
+func InputPodTest5() *corev1.Pod {
+	return &corev1.Pod{
+		Spec: corev1.PodSpec{
```
songjiaxun: Please add a container and a volume in the input Pod, and in the expected pod spec, make sure the sidecar container and the tmp volume are added at `0` position.

pkg/webhook/mutatingwebhook_test.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{
+				"gke-gcsfuse/volumes": "true",
+			},
+		},
+	}
+}
+
+func ExpectedPodTest5() *corev1.Pod {
```
songjiaxun: If this is only used by `ResponseTest5`, probably you can combine them together.

pkg/webhook/mutatingwebhook_test.go
hime: Added a new unit test to cover this case. The `compareResponses()` function checks for sidecar order to match with expected result.

pkg/webhook/mutatingwebhook_test.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	if len(wantResponse.Patches) != len(gotResponse.Patches) {
+		return fmt.Errorf("expecting %d patches, got %d patches", len(gotResponse.Patches), len(gotResponse.Patches))
+	}
+	var wantPaths, gotPaths []string
+	for i := 0; i < len(wantResponse.Patches); i++ {
+		wantPaths = append(wantPaths, wantResponse.Patches[i].Path)
+		gotPaths = append(gotPaths, gotResponse.Patches[i].Path)
+	}
+	sort.Strings(wantPaths)
```
songjiaxun: Because you are using the cmp package, I guess you don't need to sort the array. Instead you can do something like this:

```
// Compare arrays regardless of order
    isEqual := cmp.Equal(x, y, cmpopts.SortSlices(nil))
```

pkg/webhook/mutatingwebhook_test.go
songjiaxun: The newly added test case looks good, and we can remove this test case and rename the newly added case as `validInputPod` -- we expect the input Pod spec should always already have some workload containers.

hime: Sounds good.

pkg/webhook/mutatingwebhook_test.go
hime: When they are not equal, the current way allows us to diff the contents to find out what specifically differs and print it as part of the error string.

PR #183, opened 2024-02-22
Container sidecar suppport for k8s >= 1.29

hime:
Init container sidecar suppport for k8s >= 1.29.


cmd/webhook/main.go
```diff
@@ -65,9 +68,28 @@ func main() {
...
+	kubeConfig := config.GetConfigOrDie()
+
+	// Setup client
+	client, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		klog.Fatalf("Unable to get clientset: %v", err)
+	}
+
+	// Setup stop channel
+	stopCh := signals.SetupSignalHandler()
```
hime: We are using this channel for both the informer and the manage. I'm wondering if this is okay, since there are cases in which sharing this is not recommended due to race conditions, but I'm not too familiar.

cmd/webhook/main.go
```diff
@@ -65,9 +68,28 @@ func main() {
...
+
+	// Setup Informer
+	informerFactory := informers.NewSharedInformerFactory(client, time.Duration(1))
+	nodeLister := informerFactory.Core().V1().Nodes().Lister()
+
+	informerFactory.Start(stopCh.Done())
+	informerFactory.WaitForCacheSync(stopCh.Done())
```
hime: I wonder if we need the WaitForCacheSync, some other code snippets I've seen do not use this. However, My unit tests failed when this was missing, so it might not hurt to include.

pkg/webhook/mutatingwebhook.go
```diff
@@ -174,3 +191,53 @@ func parseSidecarContainerImage(pod *corev1.Pod) (string, error) {
...
+func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
+	list, err := si.NodeLister.List(labels.Everything())
+	if err != nil {
+		return false, fmt.Errorf("failed to get cluster nodes: %w", err)
+	}
+
+	supportsNativeSidecar := true
+	for _, node := range list {
+		version, err := getRelease(node.Status.NodeInfo.KubeletVersion)
+		if version < "1.29" || err != nil {
```
hime: Using string comparison to compare major versions. Unit tests work fine, but I'm wondering if there's a better way...

pkg/webhook/mutatingwebhook.go
```diff
@@ -174,3 +191,53 @@ func parseSidecarContainerImage(pod *corev1.Pod) (string, error) {
 
 	return image, nil
 }
+
+func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
+	list, err := si.NodeLister.List(labels.Everything())
+	if err != nil {
+		return false, fmt.Errorf("failed to get cluster nodes: %w", err)
```
hime, line 207: ## Logging vs returning error
This failure seems important enough to reject admission of pod. However, an alternative is to log an error and use regular sidecar.

pkg/webhook/mutatingwebhook.go
```diff
@@ -174,3 +191,53 @@ func parseSidecarContainerImage(pod *corev1.Pod) (string, error) {
...
+	}
+
+	supportsNativeSidecar := true
+	for _, node := range list {
+		version, err := getRelease(node.Status.NodeInfo.KubeletVersion)
+		if version < "1.29" || err != nil {
+			if err != nil {
+				klog.Errorf(`invalid node gke version: could not get node "%s" k8s release from version "%s": "%v"`, node.Name, version, err)
```
hime: ### Logging vs returning error
Same thought here, we could return an error here. However, this is less severe so we can just disable native sidecar support in this case.

pkg/webhook/mutatingwebhook.go
```diff
@@ -174,3 +191,53 @@ func parseSidecarContainerImage(pod *corev1.Pod) (string, error) {
...
+			}
+			supportsNativeSidecar = false
+			break
+		}
+	}
+
+	return supportsNativeSidecar, nil
+}
+
+func getRelease(version string) (string, error) {
```
hime: This method plays a big role in deciding enablement of the feature. It would be good to make sure there's no errors here, as there isn't anything similar to use externally. May want to add more test cases to cover more scenarios.

cmd/webhook/main.go
songjiaxun: Yeah I think it makes sense to have this wait to populate the informer before the webhook starts to accept requests.

deploy/base/webhook/kustomization.yaml
```diff
@@ -18,4 +18,5 @@ kind: Kustomization
 namespace: gcs-fuse-csi-driver
 resources:
 - deployment.yaml
-- mutatingwebhook.yaml
\ No newline at end of file
+- mutatingwebhook.yaml
+- webhook_setup.yaml
```
songjiaxun: let's add an extra line after.

deploy/base/webhook/webhook_setup.yaml
```diff
@@ -0,0 +1,50 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: gcs-fuse-csi-webhook-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: gcs-fuse-csi-webhook-role
+subjects:
+  - kind: ServiceAccount
+    name: gcsfusecsi-webhook-sa
```
songjiaxun: extra line after.

pkg/csi_driver/node.go
```diff
@@ -211,8 +213,16 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		return nil, status.Errorf(code, "the sidecar container failed with error: %v", errMsgStr)
 	}
 
+	var list []corev1.ContainerStatus
```
songjiaxun: let's rename it to `containerStatusList`?

Issue #42, opened 2023-06-29
Flexible Authentication Support


everpeace:
Hi, Thank you very much for the great project! I'm really surprised that FUSE can run in the sidecar container _without_ any privileges!

As kubernetes platform admin point of view, supporting FUSE WAS difficult because we have to give privilege to application developers. But, this project proved it can breaks the limitation (thanks to "file descriptor passing" between CSI driver and FUSE sidecar).

## Context/Scenario

- I(platform admin) develops a in-house kubernetes platform for internal application developers
- I would like to support gcs-fuse-csi-driver in our clusters(multiple clusters)
- GCP project of the kubernetes clusters are managed by us(platform admin)
- But, each GCP project for applications is fully owned by the application developers

## The Problem

Current, `gcs-fuse-csi-driver` implementation depends on [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity).

However, if I understood correctly, if the application runs in multiple kubernetes clusters, application developer has to create [iam-policy-binding for each k8s cluster(k8s service account)](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to:~:text=Allow%20the%20Kubernetes%20service%20account%20to%20impersonate%20the%20IAM%20service%20account%20by%20adding%20an%20IAM%20policy%20binding%20between%20the%20two%20service%20accounts.%20This%20binding%20allows%20the%20Kubernetes%20service%20account%20to%20act%20as%20the%20IAM%20service%20account.). That also means the application developer will need to update iam-policy-binding whenever our cluster is added/removed.

As a platform admin, the UX is not so convenient. I would like to reduce this toils on the application developer side.

### Option 1. Supporting GCP Service Account's Private Key in Kubernetes Secret

This would be handy. Of course, I understand Workload Identity is more secure than long lived(never expired) secret key file. 

Our platform can provide a feature which syncs the secret across our clusters. In this case, application developers need nothing when the cluster which the application runs on is added/reduced. What the application developers need is only to specify the secret name in their manifest.

[...]

PR #214, opened 2024-04-10
Add Fleet Workload Identity support for GKE on-prem environments (#42)

sshcherbakov:
Support Fleet Workload Identity authentication in Kubernetes clusters configured with Workload Identity Federation running outside of Google Cloud (e.g. GKE on Bare Metal).

This PR adds support for the `external_account` type of credentials passed via `GOOGLE_APPLICATION_CREDENTIALS` environment variable to the driver DaemonSet pods and reuses the Google Cloud Storage API authentication on behalf of the pod identity that mounts the GCS bucket into the CSI volume.

This PR follows and replicates the Workload Identity Federation [support](https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/pull/190) in the Google Secret Manager Provider for Secret Store CSI Driver.

### DIFF
--- a/docs/fleet-wif-notes.md
+++ b/docs/fleet-wif-notes.md
@@ -1,0 +1,102 @@
+# Fleet Workload Identity Authentication
+
+This page contains example configuration to configure the `gcs-fuse-csi-driver`  
+with [Fleet Workload Identity](https://cloud.google.com/anthos/fleet-management/docs/use-workload-identity) 
+authentication in environments configured for 
+[Workload Identity Federation](https://cloud.google.com/iam/docs/workload-identity-federation) 
+outside of the Google Cloud.
+
+## `external_account` Credentials
+
+Instead of the Google Service Account key file, it is possible to pass a Fleet Workload Identity configuration
+JSON file to the process that needs authenticating to the Google API in a Kubernetes cluster configured for 
+the Workload Identity Federation. The `gcs-fuse-csi-driver` pods are such processes
+that need to authenticate to the Google Cloud Storage service API to provide access to the data stores in the 
+Google Cloud Storage buckets.
+
+Such configuration file contains `external_account` type of credential that does not contain any secrets similar 
+to the Google Service Account key. The configuration should be passed via the `GOOGLE_APPLICATION_CREDENTIALS` 
+environment variable, which requires the file name of the file containing the configuration on 
+the pod's local file system.
+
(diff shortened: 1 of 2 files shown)


songjiaxun:
Hi @sshcherbakov , thanks for the PR!

It makes sense to me to add `external_account` type of credentials in the CSI node server. However, I am curious how it works in the sidecar container. The metadata server you are changing is only for credential validation in the CSI node server, and I don't think the sidecar container can get the credential populated from the `ConfigMap`. In your test, did you manually mount the `ConfigMap` to the GCSFuse sidecar container and set `GOOGLE_APPLICATION_CREDENTIALS`?

Also, can we use `Secret` instead of `ConfigMap` to store the credentials?

sshcherbakov:
Hello @songjiaxun, thank you for looking into this!

If I take the standard [example deployment](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/examples/ephemeral/deployment.yaml)
and grant "Storage Admin" IAM role to the `FLEET_PROJECT_ID.goog[gcs-csi-example/gcs-csi]` principal (I'm granting permissions to the WIF identity directly) that the example pods are running with, they get full access to the bucket. It gets mounted, the writer writes, the bucket contents is readable in the pod's containers at `/data` directory.
Hard to tell directly why).

I didn't manipulate the test deployment. It was automatically instrumented by the driver with `gke-gcsfuse-sidecar` sidecar following `gke-gcsfuse/volumes: "true"` annotation.

The content of the ConfigMap needs to be available via environment variable.
Hence we can use a Kubernetes Secret instead as well. 
The ConfigMap/Secret does not contain any secrets to hide, hence I picked ConfigMap for the example.
Please let me know if I should change the ConfigMap in the document file example to Secret.

songjiaxun:
Thanks @sshcherbakov , i am testing on my end. Will get back to you later.

songjiaxun:
Hi @sshcherbakov , can you share your cluster setup steps?

According to https://cloud.google.com/iam/docs/workload-identity-federation-with-kubernetes and https://cloud.google.com/kubernetes-engine/fleet-management/docs/use-workload-identity, there should be one step required to mount a [projected token volume](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection) to the workload Pod. I am not sure why it's working in your step without this step.

sshcherbakov:
This PR is unfortunately not sufficient, it relies on (finds) the metadata server when running on GCP (GCE VMs). More integration is needed to complete this functionality.

PR #268, opened 2024-05-24
introduce a new PV volume attribute  to skip bucket access check in CSI

saikat-royc:
This is a followup to #255  and #259. This PR introduces support for a new `csi.VolumeAttributes.skipCSIBucketAccessCheck` flag. The default value of this flag is false. It can be enabled, to skip the CSI bucket access control checks during NodePublishVolume calls.

Sample csi spec looks as follows:
```
  csi:
    volumeAttributes:
      skipCSIBucketAccessCheck: "true"
```

Manual tests:

1) Successful mounts:
Tested ephemeral and PV based volumes with the attribute set to true/false respectively

2) Manual tests of failed scenarios
2.1) permission denied errors -> ensure PermissionDenied error code returned by CSI
```
I0524 17:30:20.681081       1 node.go:118] NodePublishVolume on volume "csi-4f65eb5233adfe57354603c3c022d40600d12dbbc7332d1d84c9ac2eb9614eb8" has skipBucketAccessCheck true
E0524 17:30:20.690302       1 utils.go:103] /csi.v1.Node/NodePublishVolume failed with error: rpc error: code = PermissionDenied desc = the sidecar container failed with error: Error while mounting gcsfuse: mountWithArgs: mountWithStorageHandle: fs.NewServer: create file system: SetUpBucket: Error in iterating through objects: googleapi: Error 403: Caller does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist)., forbidden
gcsfuse exited with error: exit status 
```
2.2) missing bucket errors -> ensure NotFound error code returned by CSI

```
I0524 17:33:28.4560281 node.go:118] NodePublishVolume on volume "csi-9ba09bcedc51ab2208f51f44781e124303b1f87767d3c9aff1ab2ed21e56ad14" has skipBucketAccessCheck true
E0524 17:33:28.4678561 utils.go:103] /csi.v1.Node/NodePublishVolume failed with error: rpc error: code = NotFound desc = the sidecar container failed with error: Error while mounting gcsfuse: mountWithArgs: mountWithStorageHandle: fs.NewServer: create file system: SetUpBucket: Error in iterating through objects: storage: bucket doesn't exist
gcsfuse exited with error: exit status 1
```

tested failedMount e2e testcases on a locally deployed CSI driver
```
make e2e-test E2E_TEST_FOCUS="failedMount"

Ran 14 of 199 Specs in 68.272 seconds
[...]

PR #269, opened 2024-05-24
update failed_mount e2e testcases for skip bucket access knob

saikat-royc:
[Part 2/3]: Followup to the #268 PR, this updates the failed mount tests for the CSI Bucket Access Skip config enabled


test/e2e/testsuites/failed_mount.go
```diff
@@ -152,13 +174,26 @@ func (t *gcsFuseCSIFailedMountTestSuite) DefineTests(driver storageframework.Tes
...
 		tPod.WaitForFailedMountError(ctx, "storage service manager failed to setup service: context deadline exceeded")
+	}
+
+	ginkgo.It("should fail when the specified service account does not have access to the GCS bucket", func() {
+		testCaseSAInsufficientAccess("")
 	})
 
-	ginkgo.It("should fail when the sidecar container is not injected", func() {
-		init()
+	ginkgo.It("[csi-skip-bucket-access-check] should fail when the specified service account does not have access to the GCS bucket", func() {
```
msau42, line 191: do you want to add a test case where it starts without access, and then access gets added?

msau42:
lgtm

test/e2e/testsuites/failed_mount.go
songjiaxun: Good point! The tricky part is how to give the SA access after it's created. We can reuse the following code: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6c96f53e2567fb3ea9a3fa9372aea03ef306b62b/test/e2e/testdriver.go#L354-L360

songjiaxun (commented):
LGTM

test/e2e/testsuites/failed_mount.go
saikat-royc: Created an issue to track the test case and unblock this PR
https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/issues/271

PR #282, opened 2024-06-03
support equal sign for gcsfuse config file mount options

songjiaxun:
With this change, the GCSFuse configuration file can be configured via CSI mount options using `=` equal sign. The existing logic is not changed. For example, mount option `file-cache:max-size-mb:100`, or `file-cache:max-size-mb=100` will be translated to yaml file:

```yaml
file-cache:
  max-size-mb:100
```


songjiaxun:
This change is problematic when the config file field only has one level and does not contain `:`. The change will wrongly pass the one-level config file field to CLI flag.

songjiaxun:
Closing this PR as GCSFuse added CLI/Config file party feature.

PR #284, opened 2024-06-09
[WIP] retry gcsfuse launch in sidecar container

songjiaxun:
This is a POC PR showing the feasibility of relaunching gcsfuse in sidecar container. The ultimate goal is to remove the bucket access logic from the CSI `NodePublishVolume` call.


songjiaxun:
Achieved the same goal using a simpler design: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/306

PR #293, opened 2024-06-16
use Pod informer to avoid Pod GET API calls

songjiaxun:
This change refactors `GetPod` function to use Pod informer and lister to get Pod spec. This change avoids direct Pod GET API calls to increase the CSI node driver scalability.

### DIFF
--- a/deploy/base/node/node_setup.yaml
+++ b/deploy/base/node/node_setup.yaml
@@ -35,7 +35,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["pods"]
-    verbs: ["get"]
+    verbs: ["get", "watch", "list"]
   - apiGroups: [""]
     resources: ["serviceaccounts"]
     verbs: ["get"]
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -19,19 +19,24 @@ package clientset
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"time"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
 
 type Interface interface {
-	GetPod(ctx context.Context, namespace, name string) (*corev1.Pod, error)
+	ConfigurePodLister(nodeName string)
+	GetPod(namespace, name string) (*corev1.Pod, error)
 	CreateServiceAccountToken(ctx context.Context, namespace, name string, tokenRequest *authenticationv1.TokenRequest) (*authenticationv1.TokenRequest, error)
 	GetGCPServiceAccountName(ctx context.Context, namespace, name string) (string, error)
 }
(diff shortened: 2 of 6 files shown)


pkg/cloud_provider/clientset/clientset.go
```diff
@@ -43,9 +45,10 @@ type PodInfo struct {
...
 type Clientset struct {
 	k8sClients kubernetes.Interface
+	podLister  listersv1.PodLister
 }
 
-func New(kubeconfigPath string) (Interface, error) {
+func New(kubeconfigPath, nodeName string) (Interface, error) {
 	var err error
 	var rc *rest.Config
 	if kubeconfigPath != "" {
```
saikat-royc, line 58: should we also return an error  if node name is an empty string?

also we already fatal in L78 in main.go, so we can return error from here

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -64,11 +67,27 @@ func New(kubeconfigPath string) (Interface, error) {
 		klog.Fatal("failed to configure k8s client")
 	}
 
-	return &Clientset{k8sClients: clientset}, nil
+	// Resync period of 0 means no resync
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
+		clientset,
+		0,
```
saikat-royc: why is the resync 0?

songjiaxun: Updated the change using a 30-minute resync. This means the informer will do a Pod re-list call every 30 mins to resolve any inconsistency caused by Pod Watch errors.

pkg/cloud_provider/clientset/clientset.go
songjiaxun: If we return an error when the node name is empty, it will always crash in the CSI controller because controller does not pass a node name, although the CSI controller is not in production.

I refactored the code to only prepare the pod informer in CSI Node server. CSI controller does not use pod informer at all.

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -19,19 +19,28 @@ package clientset
...
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
 
+const (
+	resyncDuration = time.Minute * 30
```
saikat-royc: can we make this a configurable flag (with default 30)?

deploy/base/node/node_setup.yaml
```diff
@@ -35,7 +35,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["pods"]
-    verbs: ["get"]
+    verbs: ["get", "watch", "list"]
```
saikat-royc, line 38: please make an internal bug so that we track adding this rbac as part of the release where the change goes in

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -56,19 +66,40 @@ func New(kubeconfigPath string) (Interface, error) {
...
+	ctx := context.Background()
+	informerFactory.Start(ctx.Done())
+	informerFactory.WaitForCacheSync(ctx.Done())
+
+	c.podLister = podLister
+}
+
+func (c *Clientset) GetPod(namespace, name string) (*corev1.Pod, error) {
+	if c.podLister == nil {
+		return nil, errors.New("GetPod is called before ConfigurePodLister")
```
saikat-royc: how about "pod informer not ready" error string?

saikat-royc:
looks good after the minor comments are addressed

@hime is this tested?

pkg/cloud_provider/clientset/clientset.go
songjiaxun: Done.

pkg/cloud_provider/clientset/clientset.go
songjiaxun: Done.

deploy/base/node/node_setup.yaml
songjiaxun: Done.

hime:
> looks good after the minor comments are addressed
> 
> Jaime is this tested?

Hi @saikat-royc, we have tested this PR with informer. We saw that the pod get calls are eliminated and the nodeselector is able to only track the pods that are assigned to that specific node.

PR #294, opened 2024-06-16
cache token sources

songjiaxun:
This change caches service account tokens to avoid frequent STS and IAM API calls.

### DIFF
--- a/pkg/cloud_provider/auth/token_manager.go
+++ b/pkg/cloud_provider/auth/token_manager.go
@@ -18,6 +18,8 @@ limitations under the License.
 package auth
 
 import (
+	"time"
+
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/metadata"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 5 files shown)


pkg/cloud_provider/auth/token_manager.go
```diff
@@ -34,25 +36,34 @@ type TokenManager interface {
...
+	tokenSourceStore *TokenSourceStore
 }
 
 func NewTokenManager(meta metadata.Service, clientset clientset.Interface) TokenManager {
 	tm := tokenManager{
-		meta:       meta,
-		k8sClients: clientset,
+		meta:             meta,
+		k8sClients:       clientset,
+		tokenSourceStore: NewTokenSourceStore(2*time.Minute, 30*time.Second),
```
saikat-royc: let's make these configurable

songjiaxun: Done.

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -85,6 +89,10 @@ func (ts *GCPTokenSource) fetchK8sSAToken(ctx context.Context) (*oauth2.Token, e
 		return nil, fmt.Errorf("could not find token for the identity pool %q", ts.meta.GetIdentityPool())
```
hime, line 89: Is there a reason we don't call the api when the token is not found?

PR #305, opened 2024-07-04
Remove gcsfuse arm binary dependency for building csi driver.

hime:


Makefile
```diff
@@ -81,6 +81,7 @@ ifeq (${BUILD_GCSFUSE_FROM_SOURCE}, true)
 		cp /gcsfuse_${GCSFUSE_VERSION}_amd64/usr/bin/gcsfuse /release
 
 ifeq (${BUILD_ARM}, true)
+    docker buildx create --name arm-builder --bootstrap --use
```
songjiaxun: Can you explain why this is needed? And please make sure the indentation is consistent.

hime: Remove and add & review in future commit.

PR #306, opened 2024-07-05
cache NodePublishVolume response state for republish

songjiaxun:
This change caches NodePublishVolume response state in memory. In the following republish calls, the bucket access check will be skipped if it has ever succeeded. The volume state is cleaned up when the NodeUnpublishVolume is called against the same target path.

This change will completely remove STS API requests, Kube API SA API GET requests, and GCS API Object List requests in the node republish calls.

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -50,6 +50,11 @@ type nodeServer struct {
 	volumeLocks           *util.VolumeLocks
 	k8sClients            clientset.Interface
 	limiter               rate.Limiter
+	volumeStateStore      map[string]*volumeState
+}
+
+type volumeState struct {
+	bucketAccessCheckPassed bool
 }
 
 func newNodeServer(driver *GCSDriver, mounter mount.Interface) csi.NodeServer {
@@ -60,6 +65,7 @@ func newNodeServer(driver *GCSDriver, mounter mount.Interface) csi.NodeServer {
 		volumeLocks:           util.NewVolumeLocks(),
 		k8sClients:            driver.config.K8sClients,
 		limiter:               *rate.NewLimiter(rate.Every(time.Second), 10),
+		volumeStateStore:      make(map[string]*volumeState),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/csi_driver/node.go
```diff
@@ -100,10 +106,17 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 	defer s.volumeLocks.Release(targetPath)
 
+	vs, ok := s.volumeStateStore[targetPath]
```
saikat-royc: maybe put a comment about why we chose targetPath as the key ? target path corresponds to the pods and pods may belong to different namespace and would need its own access check.

saikat-royc: for the two modes:
1. bucketName "_"
2. skipBucketAccessCheck=true

we do not need to bookkeep an entry in the map an optimize of the map mem usage. 
We can move L109-L113 inside the code block where we actually are about to perform the check (L119)
Thoughts?

pkg/csi_driver/node.go
```diff
@@ -114,6 +127,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 			return nil, status.Errorf(storage.ParseErrCode(err), "failed to get GCS bucket %q: %v", bucketName, err)
 		}
 	}
+	vs.bucketAccessCheckPassed = true
```
saikat-royc: based on the comment https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/306/files#r1688774935, this can move after L128

songjiaxun: Done.

pkg/csi_driver/node.go
songjiaxun: Done.

songjiaxun: Done.

test/e2e/testsuites/failed_mount.go
```diff
@@ -192,6 +212,56 @@ func (t *gcsFuseCSIFailedMountTestSuite) DefineTests(driver storageframework.Tes
...
+
+		ginkgo.By("Checking that the pod has failed mount error PermissionDenied")
+		tPod.WaitForFailedMountError(ctx, codes.PermissionDenied.String())
+		tPod.WaitForFailedMountError(ctx, "does not have storage.objects.list access to the Google Cloud Storage bucket.")
+
+		ginkgo.By("Setting up SA IAM policy")
+		setupIAMPolicy(bucketName, f.Namespace.Name, saName)
+
+		ginkgo.By("Checking that the pod is running")
+		tPod.WaitForRunning(ctx)
```
saikat-royc, line 244: I am not able to fully understand how would the pod recover gracefully here for spec.SkipCSIBucketAccessCheckPrefix config. I was expecting the gcsfuse process to exit after SA check failure.

songjiaxun: Yes, this test should not run with the SkipCSIBucketAccessCheck volume attribute. Removed the test case.

saikat-royc:
lgtm

PR #310, opened 2024-07-11
Add Prometheus metrics support

songjiaxun:
Google Cloud Storage FUSE offers client-side Prometheus metrics for monitoring file system operations and performance. With this change, these metrics, gathered by the sidecar container, are forwarded to the CSI driver node servers. You can access these metrics via a Prometheus server. This guide explains how to enable client-side metrics, set up a Prometheus server, and query the metrics for insights.

### DIFF
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/storage"
 	driver "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/csi_driver"
 	csimounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/csi_mounter"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/metrics"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 )
@@ -45,6 +46,7 @@ var (
 	enableProfiling           = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
 	informerResyncDurationSec = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
 	fuseSocketDir             = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
+	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 
 	// These are set at compile time.
 	version = "unknown"
@@ -92,6 +94,7 @@ func main() {
 	}
 
 	var mounter mount.Interface
+	var mm metrics.Manager
 	if *runNode {
 		if *nodeID == "" {
 			klog.Fatalf("NodeID cannot be empty for node service")
@@ -102,6 +105,11 @@ func main() {
 		mounter, err = csimounter.New("", *fuseSocketDir)
 		if err != nil {
 			klog.Fatalf("Failed to prepare CSI mounter: %v", err)
+		}
+
+		if *metricsEndpoint != "" {
+			mm = metrics.NewMetricsManager(*metricsEndpoint)
+			mm.InitializeHTTPHandler()
 		}
 	}
 
@@ -115,6 +123,7 @@ func main() {
 		TokenManager:          tm,
 		Mounter:               mounter,
 		K8sClients:            clientset,
+		MetricsManager:        mm,
 	}
 
 	gcfsDriver, err := driver.NewGCSDriver(config)
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -108,7 +108,7 @@ func main() {
 		}
 
 		if *metricsEndpoint != "" {
-			mm = metrics.NewMetricsManager(*metricsEndpoint)
+			mm = metrics.NewMetricsManager(*metricsEndpoint, *fuseSocketDir)
 			mm.InitializeHTTPHandler()
 		}
 	}
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -46,7 +46,7 @@ var (
 	enableProfiling           = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
 	informerResyncDurationSec = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
 	fuseSocketDir             = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
-	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
+	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
 
 	// These are set at compile time.
 	version = "unknown"
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -34,10 +34,9 @@ import (
 )
 
 var (
-	gcsfusePath           = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
-	volumeBasePath        = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
-	metricsScrapeInterval = flag.Int("metrics-scrape-interval", 10, "Scrape interval in seconds for gcsfuse metrics endpoint.")
-	_                     = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
+	gcsfusePath    = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
+	volumeBasePath = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
+	_              = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
 	// This is set at compile time.
 	version = "unknown"
 )
@@ -63,7 +62,7 @@ func main() {
 		time.Sleep(1500 * time.Millisecond)
 		mc := sidecarmounter.NewMountConfig(sp)
 		if mc != nil {
-			if err := mounter.Mount(ctx, mc, *metricsScrapeInterval); err != nil {
+			if err := mounter.Mount(ctx, mc); err != nil {
 				mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to mount bucket %q for volume %q: %v\n", mc.BucketName, mc.VolumeName, err))
 			}
 		}
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -34,9 +34,10 @@ import (
 )
 
 var (
-	gcsfusePath    = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
-	volumeBasePath = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
-	_              = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
+	gcsfusePath           = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
+	volumeBasePath        = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
+	metricsScrapeInterval = flag.Int("metrics-scrape-interval", 10, "Scrape interval in seconds for gcsfuse metrics endpoint.")
+	_                     = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
 	// This is set at compile time.
 	version = "unknown"
 )
@@ -62,7 +63,7 @@ func main() {
 		time.Sleep(1500 * time.Millisecond)
 		mc := sidecarmounter.NewMountConfig(sp)
 		if mc != nil {
-			if err := mounter.Mount(ctx, mc); err != nil {
+			if err := mounter.Mount(ctx, mc, *metricsScrapeInterval); err != nil {
 				mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to mount bucket %q for volume %q: %v\n", mc.BucketName, mc.VolumeName, err))
 			}
 		}
--- a/deploy/base/node/node.yaml
+++ b/deploy/base/node/node.yaml
@@ -52,6 +52,10 @@ spec:
             - --nodeid=$(KUBE_NODE_NAME)
             - --node=true
             - --identity-provider=$(IDENTITY_PROVIDER)
+            - --metrics-endpoint=:9920
+          ports:
+          - containerPort: 9920
+            name: metrics
           resources:
             limits:
               cpu: 200m
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -89,7 +89,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	// Validate arguments
-	targetPath, bucketName, fuseMountOptions, skipBucketAccessCheck, err := parseRequestArguments(req)
+	targetPath, bucketName, fuseMountOptions, skipBucketAccessCheck, disableMetricsCollection, err := parseRequestArguments(req)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
@@ -149,6 +149,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 
 		return nil, status.Error(codes.FailedPrecondition, "failed to find the sidecar container in Pod spec")
+	}
+
+	// Register metrics collecter.
+	// It is idempotent to register the same collector in node republish calls.
+	if s.driver.config.MetricsManager != nil && !disableMetricsCollection {
+		klog.V(6).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
+		s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, bucketName)
 	}
 
 	// Check if the sidecar container is still required,
@@ -219,6 +226,12 @@ func (s *nodeServer) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpubli
 		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, targetPath)
 	}
 	defer s.volumeLocks.Release(targetPath)
+
+	// Unregister metrics collecter.
+	// It is idempotent to unregister the same collector.
+	if s.driver.config.MetricsManager != nil {
+		s.driver.config.MetricsManager.UnregisterMetricsCollector(targetPath)
+	}
 
 	delete(s.volumeStateStore, targetPath)
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -18,13 +18,13 @@ limitations under the License.
 package metrics
 
 import (
-	"errors"
+	"context"
 	"fmt"
+	"io"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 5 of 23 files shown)


cmd/sidecar_mounter/gcsfuse_binary
```diff
@@ -1 +1 @@
-gs://gke-release-staging/gcsfuse/v2.3.1-gke.0/gcsfuse_bin
+gs://gke-gcsfuse-test-binary
```
songjiaxun: This will be removed when the new gcsfuse release includes the prometheus changes.

deploy/base/node/node.yaml
```diff
@@ -52,6 +55,10 @@ spec:
             - --nodeid=$(KUBE_NODE_NAME)
             - --node=true
             - --identity-provider=$(IDENTITY_PROVIDER)
+            - --metrics-endpoint=:8080
```
songjiaxun: The CSI driver enables the metrics collection by default, but users still need to explicitly enable the metrics on their volume spec.

test/e2e/testsuites/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+var expectedMetricNames = []string{
+	"fs_ops_count",
+	"fs_ops_error_count",
+	"fs_ops_latency",
+	"gcs_download_bytes_count",
+	"gcs_read_count",
+	// "gcs_read_bytes_count", // TODO: cannot find these metrics
```
songjiaxun: Hi @sethiay, in my e2e tests, I don't see these metrics, so what operation will trigger these metrics?

sethiay: reading a file should trigger these metrics.

songjiaxun: @sethiay yes, in this e2e test I am doing `ls`, `touch`, `cat`, and `rm` commands against the mounted volume, but I still don't see these metrics. Line 140 - 146 are the test steps. Could you help check if I am missing anything?

sethiay: @songjiaxun I checked for one of the commented out metrics i.e. gcs_reader_count and I am able to see the reader_count exported via stackdriver on console but not when I do `curl http://localhost:8080/metrics` as mentioned in description of https://github.com/GoogleCloudPlatform/gcsfuse/pull/1859, then I didn't get it.

CC' @kislaykishore

songjiaxun: I missed one line change to enable these metrics in the gcsfuse PR, modified the PR: https://github.com/GoogleCloudPlatform/gcsfuse/pull/1859/files#diff-57454978dcd397e797f4c48e4b1f80b4b249655d5fc975ad55c54d560ba01de1

test/e2e/testsuites/metrics.go
```diff
@@ -0,0 +1,207 @@
...
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+var expectedMetricNames = map[string]int{
+	"fs_ops_count":                15,
+	"fs_ops_error_count":          2,
+	"fs_ops_latency":              15,
+	"gcs_download_bytes_count":    1,
+	"gcs_read_count":              1,
+	"gcs_read_bytes_count":        0, // TODO: cannot find this metric in e2e tests.
```
songjiaxun: Hi @sethiay, @kislaykishore , I modified the [GCSFuse PR](https://github.com/GoogleCloudPlatform/gcsfuse/pull/1859/files#diff-57454978dcd397e797f4c48e4b1f80b4b249655d5fc975ad55c54d560ba01de1) to enable the gcs metrics, then I see some new metrics.

However, I still don't see the metric `gcs_read_bytes_count`. Could you help check what I am missing? Thank you!

test/e2e/testsuites/metrics.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/onsi/gomega"
+	dto "github.com/prometheus/client_model/go"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2evolume "k8s.io/kubernetes/test/e2e/framework/volume"
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+var expectedMetricNames = map[string]int{
```
songjiaxun, line 40: Hi @sethiay , @kislaykishore , here is a list of metrics count I got from my e2e test. You can find the operations the test does against the volume [here](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/310/files#diff-9809c261a6f56bae5ce6555f53a1bfe2a87ca1ae5c31fda3a30dd105fba5f0daR140-R146).

Could you help confirm if the count of different metrics look good to you? I will share the detailed metrics output internally. Thank you!

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,255 @@
...
+		klog.Errorf("failed to register metrics collector for pod %v/%v, bucket %q: %v", podNamespace, podName, bucketName, err)
+
+		return
+	}
+
+	podUID, volumeName, _ := util.ParsePodIDVolumeFromTargetpath(targetPath)
+	promFilePath := emptyDirBasePath + metricsFileName
+	c := NewTextFileCollector(promFilePath, podUID, volumeName, map[string]string{
+		"pod_name":      podName,
+		"pod_namespace": podNamespace,
```
hime: we need to update this from `pod_namespace` to `namespace_name`

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,255 @@
...
+		"volume_name":   volumeName,
+		"bucket":        bucketName,
```
hime: we need to update this from `bucket` to `bucket_name`

test/e2e/testsuites/metrics.go
```diff
@@ -0,0 +1,207 @@
...
+
+		ginkgo.By("Deploying the pod")
+		tPod.Create(ctx)
+		defer tPod.Cleanup(ctx)
+
+		ginkgo.By("Checking that the pod is running")
+		tPod.WaitForRunning(ctx)
+
+		ginkgo.By("Checking that the pod command exits with no error")
+		tPod.VerifyExecInPodSucceed(f, specs.TesterContainerName, fmt.Sprintf("mount | grep %v | grep rw,", mountPath))
```
songjiaxun, line 139: @hime please follow the suggestion from the GCSFuse team to add more file system operations in this test. The goal is we want to trigger all the fs operations defined in this file: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/fs/wrappers/monitoring.go

test/e2e/testsuites/metrics.go
songjiaxun, line 40: @hime After you update the file system operations in the test case, also update this map.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,255 @@
...
+
+		for n, v := range c.constLabels {
+			LabelNames = append(LabelNames, n)
+			LabelValues = append(LabelValues, v)
+		}
+
+		emitNewConstMetric := func() {
+			ch <- prometheus.MustNewConstMetric(
+				prometheus.NewDesc(
+					metricFamily.GetName(),
```
songjiaxun, line 238: @hime Here is the place you can find the metric name, and you can modify the metric label `fs_error` in the metric `fs_ops_error_count`.

pkg/csi_driver/node.go
```diff
@@ -220,6 +227,11 @@ func (s *nodeServer) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpubli
 	}
 	defer s.volumeLocks.Release(targetPath)
 
+	// Unregister metrics collecter.
+	if s.driver.config.MetricsManager != nil {
+		s.driver.config.MetricsManager.UnregisterMetricsCollector(targetPath)
```
saikat-royc, line 233: is this idempotent as well? (example scenario: a node unpublish retry due to failures)

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -204,3 +212,102 @@ func logVolumeTotalSize(dirPath string) {
 		klog.Infof("total volume size of %v: %v bytes", dirPath, totalSize)
 	}
 }
+
+// collectMetrics collects metrics from the gcsfuse instance every 10 seconds.
+func collectMetrics(ctx context.Context, port, dirPath string) {
+	genNumber := 0
```
saikat-royc: please put a comment how to generation number is used (to co-ordinate the producer sidecar and consumer csi node driver)

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -204,3 +212,102 @@ func logVolumeTotalSize(dirPath string) {
...
+	ticker := time.NewTicker(10 * time.Second)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			newCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			scrapeMetrics(newCtx, metricEndpoint, dirPath, genNumber)
+			deleteOutdatedFile(dirPath, genNumber-2)
```
saikat-royc: can you please elaborate why we are only deleting genNum-2? also best to put a comment in code

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -204,3 +212,102 @@ func logVolumeTotalSize(dirPath string) {
...
+
+		return
+	}
+
+	// Ensure closure of output file.
+	out.Close()
+
+	// Update generation number.
+	generationFilePath := dirPath + metrics.GetGenerationFileName()
+	genFile, err := os.Create(generationFilePath)
```
saikat-royc: if the generation file path already exists what is the behavior here?

pkg/sidecar_mounter/sidecar_mounter.go
hime: Done.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -204,3 +212,102 @@ func logVolumeTotalSize(dirPath string) {
 		klog.Infof("total volume size of %v: %v bytes", dirPath, totalSize)
 	}
 }
+
+// collectMetrics collects metrics from the gcsfuse instance every 10 seconds.
+func collectMetrics(ctx context.Context, port, dirPath string) {
+	genNumber := 1
+	scrapeInterval := 10 * time.Second
```
saikat-royc: let's make this a configurable flag

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,312 @@
...
+
+// Describe emits the description of metrics.
+// Prometheus Registry relies on this func to identify collectors.
+func (c *textFileCollector) Describe(ch chan<- *prometheus.Desc) {
+	// Collector id is a hash of the values of the ConstLabels and fqName.
+	ch <- prometheus.NewDesc("gke_gcsfuse_csi_metric", "GKE GCSFuse CSI metric.", nil, map[string]string{"pod_uid": c.podUID, "volume_name": c.volumeName})
+}
+
+// Collect emits metrics.
+func (c *textFileCollector) Collect(ch chan<- prometheus.Metric) {
```
saikat-royc: thinking a version skew case. Say the csi driver pod upgraded to the version which supports metrics, and the user pod is still using the old sidecar which does not export metrics, this code would fail in ProcessMetricsFile() as it will fail to find gen file. The code would keep reporting error logs. Which may be fine (or perhaps change to warning), but something to call out in playbooks, on how to filter out if its a genuine issue or a false positive

hime: Will downgrade the error to warning and open an item to document this false positive scenario.

pkg/metrics/metrics.go
hime: Closing as `fs_error` label was completely dropped from gcsfuse.

pkg/sidecar_mounter/sidecar_mounter.go
hime: Done.

pkg/metrics/metrics.go
hime: Done.

cmd/csi_driver/main.go
```diff
@@ -44,6 +45,7 @@ var (
 	identityProvider          = flag.String("identity-provider", "", "The Identity Provider to authenticate with GCS API.")
 	enableProfiling           = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
 	informerResyncDurationSec = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
+	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
```
kislaykishore:
```suggestion
	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,324 @@
...
+
+	podUID, volumeName, _ := util.ParsePodIDVolumeFromTargetpath(targetPath)
+	c := NewTextFileCollector(emptyDirBasePath, podUID, volumeName, map[string]string{
+		"pod_name":       podName,
+		"namespace_name": podNamespace,
+		"volume_name":    volumeName,
+		"bucket_name":    bucketName,
+		"pod_uid":        podUID,
+	})
+	if err := mm.registry.Register(c); err != nil && !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
```
kislaykishore, line 114: Can we use errors.Is(err, prometheus.AlreadyRegisteredError) instead?

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -144,13 +150,15 @@ func NewMountConfig(sp string) *MountConfig {
...
-		"gid":         "0",
+		"app-name":        GCSFuseAppName,
+		"temp-dir":        mc.BufferDir + TempDir,
+		"config-file":     mc.ConfigFile,
+		"foreground":      "",
+		"uid":             "0",
+		"gid":             "0",
+		"prometheus-port": strconv.Itoa(prometheusPort),
 	}
+	prometheusPort++
```
kislaykishore, line 162: Can you help understand or add a comment explaining why we need to increment the prometheusPort?

test/e2e/specs/specs.go
```diff
@@ -357,6 +361,23 @@ func (t *TestPod) GetNode() string {
 	return t.pod.Spec.NodeName
 }
 
+func (t *TestPod) GetCISDriverNodePodIP(ctx context.Context) string {
```
kislaykishore:
```suggestion
func (t *TestPod) GetCSIDriverNodePodIP(ctx context.Context) string {
```

test/e2e/testsuites/metrics.go
```diff
@@ -0,0 +1,253 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Copyright 2022 Google LLC
```
kislaykishore, line 3:
```suggestion
Copyright 2024 Google LLC
```

cmd/sidecar_mounter/main.go
```diff
@@ -34,9 +34,10 @@ import (
 )
 
 var (
-	gcsfusePath    = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
-	volumeBasePath = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
-	_              = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
+	gcsfusePath           = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
+	volumeBasePath        = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
+	metricsScrapeInterval = flag.Int("metrics-scrape-interval", 10, "Scrape interval for gcsfuse metrics endpoint.")
```
songjiaxun: change the description to `Scrape interval in seconds for gcsfuse metrics endpoint.`

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -213,25 +214,34 @@ func logVolumeTotalSize(dirPath string) {
...
 		select {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
 			newCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
-			scrapeMetrics(newCtx, metricEndpoint, outputPath)
+			scrapeMetrics(newCtx, metricEndpoint, dirPath, genNumber)
+			// If x is the current generation of the metrics snapshot/file, we are leaving some buffer where the outdated (x-1)
+			// file exists so that the reader has time to finish reading the complete data.
+			// The -2 means we are deleting file generations up to x - 2.
```
hime: Michelle suggested that if the writer deletes a file while the reader holds a file descriptor, the kernel should not delete the file until the file descriptor is released by all processes. 

We can test if things work fine by deleting all outdated files, and only keeping the most recent.

hime: Will keep the current approach as the following scenario fails if removed:

Scenario:
1. Reader: reads directory and finds latest file is metrics_x.prom
2. Writer : creates metrics_y.prom & deletes metrics_x.prom
3. Reader: attempts to open metrics_x.prom and fails

cmd/sidecar_mounter/main.go
hime: Done.

test/e2e/specs/specs.go
hime: Done.

pkg/csi_driver/node.go
hime: Yes, this call is idempotent.

pkg/sidecar_mounter/sidecar_mounter_config.go
hime: Added comment.

hime: Added comment. We want to have a unique port for each gcsfuse process. We have one instance running per volume mounted.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -213,25 +214,34 @@ func logVolumeTotalSize(dirPath string) {
...
+			deleteOutdatedFile(dirPath, genNumber-2)
```
songjiaxun: This only deletes one outdated file "genNumber-2", but if this deletion keep failing, we will have prom file leak on the disk.

Can we try to delete all the outdated files each time? For example, delete all the files with genNumber <= latest genNumber-2

pkg/csi_driver/node.go
songjiaxun: Added some comments.

pkg/metrics/metrics.go
```diff
@@ -89,14 +94,14 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
-	c := NewTextFileCollector(promFilePath, podUID, volumeName, map[string]string{
+	c := NewTextFileCollector(emptyDirBasePath, podUID, volumeName, map[string]string{
 		"pod_name":       podName,
 		"namespace_name": podNamespace,
 		"volume_name":    volumeName,
 		"bucket_name":    bucketName,
+		"pod_uid":        podUID,
 	})
-	if err := mm.registry.Register(c); err != nil && !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
+	if err := mm.registry.Register(c); err != nil && !errors.Is(err, prometheus.AlreadyRegisteredError{}) {
```
songjiaxun: We don't have test coverage for this line. Please manually check log to make sure we don't see this error message in re-publish calls.

pkg/metrics/metrics.go
```diff
@@ -154,22 +166,78 @@ func (c *textFileCollector) Collect(ch chan<- prometheus.Metric) {
...
+		return "", fmt.Errorf("could not get generation number from %s: %w", directoryPath, err)
+	}
+
+	// Find latest metrics file name and open.
+	metricsFilePath := filepath.Join(directoryPath, GetMetricsFileName(genNumber))
+
+	return metricsFilePath, nil
+}
+
+// GetMetricsFilePath creates the expected name for the latest metrics file.
```
songjiaxun: should be // GetMetricsFileName

pkg/sidecar_mounter/sidecar_mounter.go
hime: Since we are listing the directory contents already, we can delete all the outdated metrics files that we find there.

Edit: nevermind, we list contents from reader side. we can do this here aswell.

pkg/metrics/metrics.go
hime: Done.

pkg/metrics/metrics.go
songjiaxun: From the CSI driver log, I see the error message `duplicate metrics collector registration attempted`. It means the change cannot skip the `AlreadyRegisteredError`.

hime: seeing the same thing. we can revert this change.

songjiaxun: @kislaykishore I afraid we need to reverse this change and use `strings.Contains`. Seems like the error is wrapped so we cannot directly use `errors.Is` to check if the error is the same.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -204,3 +212,124 @@ func logVolumeTotalSize(dirPath string) {
 		klog.Infof("total volume size of %v: %v bytes", dirPath, totalSize)
 	}
 }
+
+// collectMetrics collects metrics from the gcsfuse instance every 10 seconds.
```
kislaykishore:
```suggestion
// collectMetrics collects metrics from the gcsfuse instance every `scrapeInterval` seconds.
```

kislaykishore (approved):

PR #313, opened 2024-07-16
Add injection compatibility with istio 1.18

hime:
**What type of PR is this?**
>  
> /kind api-change  

/kind bug  

> /kind cleanup  
> /kind design  
> /kind documentation  
> /kind failing-test  
> /kind feature  
> /kind flake  

This PR has the goal of updating the OSS webhook to validate sidecar injection when present at non-zero index. Our driver by default injects at the 0th index and to validate that the sidecar is already present, we only perform the validation in the first index. This behavior does not work out well with istio webhook, which also prefers that the sidecar is injected at the beginning of the container list.

We previously added support to handle scenarios where the user is using Anthos Service Mesh (ASM) and GCSFuse CSI Driver enabled. However, this support was aimed for istio v1.19. This version of istio is not currently available in the `regular channel` which surfaced incompatibilities between our driver and ASM. 

Fixes #

```release-note
Update OSS webhook to validate sidecar injection when present at non-zero index
```


pkg/webhook/parsers.go
```diff
@@ -27,27 +27,34 @@ import (
...
-			if _, _, _, err := parsers.ParseImageName(image); err != nil {
-				return "", fmt.Errorf("could not parse input image: %q, error: %w", image, err)
-			}
+
+	// Find container named "gke-gcsfuse-sidecar" (SidecarContainerName), extract its image, and remove from list.
+	if index, present := containerPresent(pod.Spec.Containers, SidecarContainerName); present {
+		image = pod.Spec.Containers[index].Image
+
+		if _, _, _, err := parsers.ParseImageName(image); err != nil {
+			return "", fmt.Errorf("could not parse input image: %q, error: %w", image, err)
```
hime, line 45: Although we are returning an error, we should not block pod creation because we could run into the situation where the sidecar . Maybe if we encounter this we can use the default sidecar container image contained in the ConfigMap.

songjiaxun: What is the situation?

songjiaxun:
LGTM, thanks for the change!

pkg/webhook/parsers.go
hime: Resolving as we can keep as is.

This is related to the case where we provide a private sidecar image sidecar where the image is not valid. I think its okay to return the error.

PR #327, opened 2024-08-14
Handle read while writing race condition.

hime:
Prevent race condition between reading and writing at the same time.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -18,9 +18,11 @@ limitations under the License.
 package metrics
 
 import (
+	"bufio"
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -33,8 +35,9 @@ import (
 )
 
 const (
-	metricsPath     = "/metrics"
-	metricsFileName = "/metrics.prom"
+	metricsPath             = "/metrics"
+	metricsFileNameTemplate = `/metrics_%d.prom`
+	generationFileName      = "/generation.txt"
 )
 
 type Manager interface {
@@ -89,8 +92,7 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
 	}
 
 	podUID, volumeName, _ := util.ParsePodIDVolumeFromTargetpath(targetPath)
-	promFilePath := emptyDirBasePath + metricsFileName
-	c := NewTextFileCollector(promFilePath, podUID, volumeName, map[string]string{
+	c := NewTextFileCollector(emptyDirBasePath, podUID, volumeName, map[string]string{
 		"pod_name":       podName,
 		"namespace_name": podNamespace,
 		"volume_name":    volumeName,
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -154,22 +156,65 @@ func (c *textFileCollector) Collect(ch chan<- prometheus.Metric) {
 
 // ProcessMetricsFile processes a metrics file that follows Prometheus text format: https://prometheus.io/docs/instrumenting/exposition_formats/,
 // returning its MetricFamily.
-func ProcessMetricsFile(path string) (map[string]*dto.MetricFamily, error) {
-	f, err := os.Open(path)
+func ProcessMetricsFile(directoryPath string) (map[string]*dto.MetricFamily, error) {
+	// Find latest file generation.
+	generationFilePath := directoryPath + generationFileName
```
msau42, line 161: I would probably use the `filepath` library to construct name so that it's OS-independent

pkg/metrics/metrics.go
```diff
@@ -154,22 +156,65 @@ func (c *textFileCollector) Collect(ch chan<- prometheus.Metric) {
...
+	return fmt.Sprintf(metricsFileNameTemplate, genNumber)
+}
+
+func GetGenerationFileName() string {
+	return generationFileName
+}
+
+// getGenerationNumber opens the generation.txt file and parses the payload into
+// an integer. This integer represents the current generation of the metrics file.
+func getGenerationNumber(filePath string) (int, error) {
```
msau42, line 195: Is reading this file also prone to racing?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -224,14 +225,18 @@ func collectMetrics(ctx context.Context, port, dirPath string) {
 			return
 		case <-ticker.C:
 			newCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
-			scrapeMetrics(newCtx, metricEndpoint, outputPath)
+			scrapeMetrics(newCtx, metricEndpoint, dirPath, genNumber)
+			deleteOutdatedFile(dirPath, genNumber-2)
```
msau42, line 229: What is the reasoning behind "-2"?

hime: If x is the current generation, we are leaving some buffer where the outdated (x-1) file exists so that if this file is being read, it has time to for the reader to finish. The -2 means we are deleting file generations up to x - 2.

pkg/metrics/metrics.go
hime: The main concern with the original approach is that we were reading and writing large number of contents to our files which could cause the writer to write in chunks, and reader to see incomplete data. Here we are only writing a piece of information and closing. In this case, i believe its difficult to read corrupted data in this file.

pkg/sidecar_mounter/sidecar_mounter.go
msau42: Can you add a comment with the reasoning?

pkg/metrics/metrics.go
msau42: Possibly? It may still be worth using flock just to avoid the possibility.

hime: In that case would we still want to keep this implementation? 

The reason to steer away from file locks was that to properly apply them, we would have to lock on both ends (when reading and writing). Performing many reads on the endpoint could make us run into issues where we cant write because of lock. I will check if there's timing for how often we read this endpoint from metrics sidecar.

pkg/sidecar_mounter/sidecar_mounter.go
hime: Sure! Pending on comment above regarding keeping this approach.

pkg/metrics/metrics.go
hime: Since this is already merged, will follow up on next PR and put PR # here.

hime: #336 contains this change.

pkg/metrics/metrics.go
hime: See #310 for latest change.

pkg/sidecar_mounter/sidecar_mounter.go
hime: Done.

PR #331, opened 2024-08-19
add ValidatingAdmissionPolicy in 1.30 clusters

songjiaxun:
Add a `ValidatingAdmissionPolicy` in 1.30+ clusters.

This `ValidatingAdmissionPolicy` does the following validations:

1. It filters out Pods with an annotation `gke-gcsfuse/volumes: true`, and has a native sidecar container with name `gke-gcsfuse-sidecar`. The validation skips any other unmatched Pods.
2. It checks if the native sidecar container `restartPolicy` is `Always`.
3. It checks if the native sidecar container has env var `NATIVE_SIDECAR: TRUE`.

If there is any errors in the validator, it does not block the Pod creation request because of `failurePolicy: Ignore`.

This change gives users clear error message when the native sidecar container is mutated by other conflicting webhooks.


saikat-royc:
We should also update the https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/docs/installation.md and any other relevant docs?

saikat-royc:
lgtm

songjiaxun:
> We should also update the https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/docs/installation.md and any other relevant docs?

Yes I will update the docs in a follow-up PR.

PR #350, opened 2024-09-16
Use protobuf encoding for core K8s APIs in gcs-fuse-csi-driver

tosi3k:
For core K8s API objects like Pods, Nodes, etc., we can use protobuf encoding which, compared to the default JSON encoding, reduces CPU consumption related to (de)serialization, reduces overall latency of the API calls, reduces memory footprint and the work performed by the GC and results in quicker propagation of objects to event handlers of shared informers.

Core system components of K8s default their serialization method to protobuf for 8 years already: https://github.com/kubernetes/kubernetes/pull/25738.

Some benchmarks comparing JSON vs. protobuf showcasing how the latter data format (de)serializes faster and uses less memory:

- https://medium.com/@akresling/go-benchmark-json-v-protobuf-4ec3c62ec8d4
- https://www.codingexplorations.com/blog/performance-comparison-protobuf-marshaling-vs-json-marshaling-in-go
- https://medium.com/@david_turner/protocol-buffers-just-how-fast-are-they-9ec19ea580db

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


cmd/webhook/main.go
```diff
@@ -77,7 +78,9 @@ func main() {
 	kubeConfig := config.GetConfigOrDie()
 
 	// Setup client
-	client, err := kubernetes.NewForConfig(kubeConfig)
+	coreKubeConfig := rest.CopyConfig(kubeConfig)
```
songjiaxun, line 81: I am curious why do we need to make a copy here?

tosi3k: Because, later on in `main` function, we pass the `kubeConfig` variable to `manager.New` function call documentation of which specifies that if `kubeConfig` doesn't specify the `ContentType`, it will set it to protobuf for core K8s APIs and to JSON for custom resources: https://github.com/kubernetes-sigs/controller-runtime/blob/2eb879f25c4829825e6f2511cda035de41ec7030/pkg/manager/manager.go#L317-L321.

This is good for us, because:
- CRD doesn't support protobuf encoding (so passing `kubeConfig` with the `ContentType` field set to protobuf would prospectively break it if we start talking to K8s API about CRs),
- it sets protobuf encoding for the core K8s APIs automatically anyway - which was the intention of this PR.

songjiaxun: Sounds good, thank you for the PR!

PR #360, opened 2024-10-14
Add Cloud Storage FUSE Metadata/Stat cache prefetch from kubernetes.

hime:
**What type of PR is this?**
>  
> /kind api-change  
> /kind bug  
> /kind cleanup  
> /kind design  
> /kind documentation  
> /kind failing-test  

/kind feature  

> /kind flake  

Prefill stat/metadata cache as pod starts to reduce number of list calls during workload execution. 

Fixes #

Yes, this PR adds a new boolean API field within `volumeAttributes` called `gcsfuseMetadataPrefetchOnMount`.

```release-note
Add Cloud Storage FUSE Metadata/Stat cache prefetch from Kubernetes.
```

### DIFF
--- a/cmd/metadata_prefetch/Dockerfile
+++ b/cmd/metadata_prefetch/Dockerfile
@@ -1,0 +1,54 @@
+# Copyright 2018 The Kubernetes Authors.
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Build metadata-prefetch go binary
+FROM --platform=$BUILDPLATFORM golang:1.22.7 AS metadata-prefetch-builder
+
+ARG STAGINGVERSION
+
+WORKDIR /gcs-fuse-csi-driver
+ADD . .
+RUN make metadata-prefetch BINDIR=/bin
+
+# go/gke-releasing-policies#base-images
+FROM gke.gcr.io/debian-base:bookworm-v1.0.4-gke.2 AS debian
+
+# go/gke-releasing-policies#base-images
+FROM  gcr.io/distroless/base-debian12 AS output-image
+
+# Copy existing binaries.
+COPY --from=debian /bin/ls /bin/ls
+
+# Copy dependencies.
+COPY --from=debian /lib/x86_64-linux-gnu/libselinux.so.1 /lib/x86_64-linux-gnu/libselinux.so.1
+COPY --from=debian /lib/x86_64-linux-gnu/libc.so.6 /lib/x86_64-linux-gnu/libc.so.6
+COPY --from=debian /lib/x86_64-linux-gnu/libpcre2-8.so.0 /lib/x86_64-linux-gnu/libpcre2-8.so.0
+COPY --from=debian /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2 
+
+# Validate dependencies
+FROM output-image AS validator-image
+COPY --from=debian /bin/bash /bin/bash
+COPY --from=debian /usr/bin/ldd /usr/bin/ldd
+COPY --from=debian /bin/grep /bin/grep
+SHELL ["/bin/bash", "-c"]
+RUN if ldd /bin/ls | grep "not found"; then echo "!!! Missing deps for ls command !!!" && exit 1; fi
+
+# Final image
+FROM output-image
+
+# Copy the built binaries
+COPY --from=metadata-prefetch-builder /bin/gcs-fuse-csi-driver-metadata-prefetch /gcs-fuse-csi-driver-metadata-prefetch
+
+ENTRYPOINT ["/gcs-fuse-csi-driver-metadata-prefetch"]
--- a/cmd/metadata_prefetch/main.go
+++ b/cmd/metadata_prefetch/main.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Copyright 2024 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"flag"
+	"os"
+	"os/exec"
+	"os/signal"
+	"strings"
+	"syscall"
+
+	"k8s.io/klog/v2"
+)
+
+const (
+	mountPathsLocation = "/volumes/"
+)
+
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	// Create cancellable context to pass into exec.
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// Handle SIGTERM signal.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+
+	go func() {
+		<-sigs
+		klog.Info("Caught SIGTERM signal: Terminating...")
+		cancel()
+
+		os.Exit(0) // Exit gracefully
+	}()
+
+	// Start the "ls" command in the background.
+	// All our volumes are mounted under the /volumes/ directory.
+	cmd := exec.CommandContext(ctx, "ls", "-R", mountPathsLocation)
+	cmd.Stdout = nil // Connects file descriptor to the null device (os.DevNull).
+
+	// TODO(hime): We should research stratergies to parallelize ls execution and speed up cache population.
+	err := cmd.Start()
+	if err == nil {
+		mountPaths, err := getDirectoryNames(mountPathsLocation)
+		if err == nil {
+			klog.Infof("Running ls on mountPath(s): %s", strings.Join(mountPaths, ", "))
+		} else {
+			klog.Warningf("failed to get mountPaths: %v", err)
+		}
+
+		err = cmd.Wait()
+		if err != nil {
+			klog.Errorf("Error while executing ls command: %v", err)
+		} else {
+			klog.Info("Metadata prefetch complete")
+		}
+	} else {
+		klog.Errorf("Error starting ls command: %v.", err)
+	}
+
+	klog.Info("Going to sleep...")
+
+	// Keep the process running.
+	select {}
+}
+
+// getDirectoryNames returns a list of strings representing the names of
+// the directories within the provided path.
+func getDirectoryNames(dirPath string) ([]string, error) {
+	directories := []string{}
+	items, err := os.ReadDir(dirPath)
+	if err != nil {
+		return directories, err
+	}
+
+	for _, item := range items {
+		if item.IsDir() {
+			directories = append(directories, item.Name())
+		}
+	}
+
+	return directories, nil
+}
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -52,6 +52,7 @@ var (
 	ephemeralStorageRequest = flag.String("sidecar-ephemeral-storage-request", "5Gi", "The default ephemeral storage request for gcsfuse sidecar container.")
 	ephemeralStorageLimit   = flag.String("sidecar-ephemeral-storage-limit", "5Gi", "The default ephemeral storage limit for gcsfuse sidecar container.")
 	sidecarImage            = flag.String("sidecar-image", "", "The gcsfuse sidecar container image.")
+	metadataSidecarImage    = flag.String("metadata-sidecar-image", "", "The metadata prefetch sidecar container image.")
 
 	// These are set at compile time.
 	webhookVersion = "unknown"
@@ -72,7 +73,7 @@ func main() {
 	klog.Infof("Running Google Cloud Storage FUSE CSI driver admission webhook version %v, sidecar container image %v", webhookVersion, *sidecarImage)
 
 	// Load webhook config
-	c := wh.LoadConfig(*sidecarImage, *imagePullPolicy, *cpuRequest, *cpuLimit, *memoryRequest, *memoryLimit, *ephemeralStorageRequest, *ephemeralStorageLimit)
+	c := wh.LoadConfig(*sidecarImage, *metadataSidecarImage, *imagePullPolicy, *cpuRequest, *cpuLimit, *memoryRequest, *memoryLimit, *ephemeralStorageRequest, *ephemeralStorageLimit)
 
 	// Load config for manager, informers, listers
 	kubeConfig := config.GetConfigOrDie()
@@ -103,6 +104,8 @@ func main() {
 	// Setup Informer
 	informerFactory := informers.NewSharedInformerFactory(client, resyncDuration)
 	nodeLister := informerFactory.Core().V1().Nodes().Lister()
+	pvcLister := informerFactory.Core().V1().PersistentVolumeClaims().Lister()
+	pvLister := informerFactory.Core().V1().PersistentVolumes().Lister()
 
 	informerFactory.Start(context.Done())
 	informerFactory.WaitForCacheSync(context.Done())
@@ -140,6 +143,8 @@ func main() {
 			Config:        c,
 			Decoder:       admission.NewDecoder(runtime.NewScheme()),
 			NodeLister:    nodeLister,
+			PvLister:      pvLister,
+			PvcLister:     pvcLister,
 			ServerVersion: serverVersion,
 		},
 	})
--- a/pkg/webhook/client.go
+++ b/pkg/webhook/client.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Copyright 2024 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"errors"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// IsPreprovisionCSIVolume checks whether the volume is a pre-provisioned volume for the desired csiDriver.
+func (si *SidecarInjector) IsPreprovisionCSIVolume(csiDriver string, pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	_, ok, err := si.GetPreprovisionCSIVolume(csiDriver, pvc)
+
+	return ok, err
+}
+
+// GetPreprovisionCSIVolume gets the pre-provisioned persistentVolume when backed by the desired csiDriver.
+func (si *SidecarInjector) GetPreprovisionCSIVolume(csiDriver string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolume, bool, error) {
+	if csiDriver == "" {
+		return nil, false, errors.New("csiDriver is empty, cannot verify storage type")
+	}
+
+	if pvc == nil {
+		return nil, false, errors.New("pvc is nil, cannot get pv")
+	}
+
+	// We return a nil error because this volume is still valid.
+	// A pvc can have this field missing when requesting a dynamically provisioned volume and said PV it is not yet bound.
+	if pvc.Spec.VolumeName == "" {
+		return nil, false, nil
+	}
+
+	// GetPV returns an error if pvc.Spec.VolumeName is not empty and the associated PV object is not found in the API server.
+	pv, err := si.GetPV(pvc.Spec.VolumeName)
+	if err != nil {
+		return nil, false, err // no additional context needed for error.
+	}
+
+	if pv == nil {
+		return nil, false, errors.New("pv is nil, cannot get storage type")
+	}
+
+	// Returns false when PV - PVC pair was created for a different csi driver or different storage type.
+	if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == csiDriver {
+		return pv, true, nil
+	}
+
+	return pv, false, nil
+}
+
+func (si *SidecarInjector) GetPV(name string) (*corev1.PersistentVolume, error) {
+	pv, err := si.PvLister.Get(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return pv, nil
+}
+
+func (si *SidecarInjector) GetPVC(namespace, name string) (*corev1.PersistentVolumeClaim, error) {
+	pvc, err := si.PvcLister.PersistentVolumeClaims(namespace).Get(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return pvc, nil
+}
--- a/pkg/webhook/injection.go
+++ b/pkg/webhook/injection.go
@@ -28,6 +28,29 @@ import (
 
 const IstioSidecarName = "istio-proxy"
 
+func (si *SidecarInjector) injectAsNativeSidecar(pod *corev1.Pod) (bool, error) {
+	supportsNativeSidecar, err := si.supportsNativeSidecar()
+	if err != nil {
+		return false, fmt.Errorf("failed to determine native sidecar injection: %w", err)
+	}
+
+	nativeSidecarEnabled := true
+	if enable, ok := pod.Annotations[GcsFuseNativeSidecarEnableAnnotation]; ok {
+		parsedAnnotation, err := ParseBool(enable)
+		if err != nil {
+			klog.Errorf("failed to parse enableNativeSidecar annotation: %v", err)
+		} else {
+			nativeSidecarEnabled = parsedAnnotation
+			// Warn the user if they used annotation incorrectly.
+			if nativeSidecarEnabled && !supportsNativeSidecar {
+				klog.Errorf("attempting to enable native sidecar on a cluster that does not support it, this is not allowed")
+			}
+		}
+	}
+
+	return nativeSidecarEnabled && supportsNativeSidecar, nil
+}
+
 func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
 	if si.ServerVersion != nil && !si.ServerVersion.AtLeast(minimumSupportedVersion) {
 		return false, nil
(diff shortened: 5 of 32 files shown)


cmd/metadata_prefetch/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"syscall"
+
+	"k8s.io/klog/v2"
+)
+
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	// Create cancellable context to pass into exec.
```
hime, line 40: We should manually verify that container exits successfully in two scenarios:

1. Sigterm is received while `ls` is running.
2. Sigterm is received after `ls` has finished.

(or also run some unit or e2e tests somehow)

cmd/metadata_prefetch/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+		os.Exit(0) // Exit gracefully
+	}()
+
+	// Start the "ls" command in the background.
+	// All our volumes are mounted under the /volumes/ directory.
+	cmd := exec.CommandContext(ctx, "ls", "-R", "/volumes/")
+	cmd.Stdout = nil // Connects file descriptor to the null device (os.DevNull).
+
+	// TODO(hime): We should research stratergies to parallelize ls execution and speed up cache population.
```
hime, line 60: Is performing a single ls to cover all volumes enough for our v1 release?

pkg/webhook/client.go
```diff
@@ -0,0 +1,79 @@
...
+
+package webhook
+
+import (
+	"errors"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func (si *SidecarInjector) IsPreprovisionCSIVolume(csiDriver string, pvc *corev1.PersistentVolumeClaim) (bool, *corev1.PersistentVolume, error) {
```
hime: lets add unit tests for all methods in this file

pkg/webhook/injection.go
```diff
@@ -71,6 +71,37 @@ func injectSidecarContainer(pod *corev1.Pod, config *Config, supportsNativeSidec
 	}
 }
 
+func (si *SidecarInjector) injectMetadataPrefetchSidecarContainer(pod *corev1.Pod, config *Config, supportsNativeSidecar bool) {
```
hime: we should skip injection if the container is already injected. There's no need to do complex logic to reinject. if its already injected, skip.

pkg/webhook/injection.go
```diff
@@ -71,6 +71,37 @@ func injectSidecarContainer(pod *corev1.Pod, config *Config, supportsNativeSidec
...
+	var containerSpec corev1.Container
+	var index int
+
+	if supportsNativeSidecar {
```
hime: we might be using criteria when defining if the workload should be using supportsNativeSidecar. We need to verify the criteria is the same throughout.

pkg/webhook/injection.go
hime: either add unit tests directly to this method, or use the broader integration test that checks the whole flow.

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +121,70 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+			SeccompProfile: &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+			RunAsNonRoot:   ptr.To(true),
+			RunAsUser:      ptr.To(int64(NobodyUID)),
+			RunAsGroup:     ptr.To(int64(NobodyGID)),
+		},
+		Args: []string{
+			"--v=5",
+		},
+		Resources: corev1.ResourceRequirements{
+			// We should change these resources.
```
hime: we need to use specific resources here. im not sure if its worth setting up a pipeline to pass in custom resources here. for v1, id say no. Id probably just hardcode them.

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +121,70 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: ptr.To(false),
+			ReadOnlyRootFilesystem:   ptr.To(true),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{
+					corev1.Capability("ALL"),
+				},
+			},
+			SeccompProfile: &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+			RunAsNonRoot:   ptr.To(true),
```
hime: i think these mean that the pod should run with very minimal permissions, we should think about keeping these or not.

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +121,70 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+func (si *SidecarInjector) GetNativeMetadataPrefetchSidecarContainerSpec(pod *corev1.Pod, c *Config) corev1.Container {
+	container := si.GetMetadataPrefetchSidecarContainerSpec(pod, c)
+	container.Env = append(container.Env, corev1.EnvVar{Name: "NATIVE_SIDECAR", Value: "TRUE"})
+	container.RestartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
+
+	return container
+}
+
+func (si *SidecarInjector) GetMetadataPrefetchSidecarContainerSpec(pod *corev1.Pod, c *Config) corev1.Container {
+	limits, requests := prepareResourceList(c)
```
hime, line 172: these should be set independently, probably hardcoded.

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +121,70 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+		},
+		Resources: corev1.ResourceRequirements{
+			// We should change these resources.
+			Limits:   limits,
+			Requests: requests,
+		},
+		VolumeMounts: []corev1.VolumeMount{},
+	}
+
+	for _, v := range pod.Spec.Volumes {
```
hime, line 188: due to this logic, we should have unit tests for this func

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +121,70 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+		if b, volumeAttributes, _ := si.isGcsFuseCSIVolume(v, pod.Namespace); b {
+			enableMetaPrefetchRaw, ok := volumeAttributes["gcsfuseMetadataPrefetchOnMount"]
```
hime: move gcsfuseMetadataPrefetchOnMount to const var

pkg/webhook/injection.go
hime: this criteria is the same for now, but will change in #351

pkg/webhook/injection.go
hime: Done.

cmd/metadata_prefetch/main.go
hime: We should make ls optimizations after v1 release.

pkg/webhook/sidecar_spec.go
hime: we need to test how many resources ls is using

pkg/webhook/injection.go
hime: let's add unit tests directly, as the integration test that covers the entire webhook does not explicitly check the new pod spec.

pkg/webhook/sidecar_spec.go
hime: we are combining the test with the integration test for injectMetadataPrefetchContainer.

pkg/webhook/injection.go
```diff
@@ -63,23 +82,57 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
-	if supportsNativeSidecar && nativeSidecarEnabled {
+func injectSidecarContainer(pod *corev1.Pod, config *Config, injectAsNativeSidecar bool) {
+	if injectAsNativeSidecar {
 		pod.Spec.InitContainers = insert(pod.Spec.InitContainers, GetNativeSidecarContainerSpec(config), getInjectIndex(pod.Spec.InitContainers))
 	} else {
 		pod.Spec.Containers = insert(pod.Spec.Containers, GetSidecarContainerSpec(config), getInjectIndex(pod.Spec.Containers))
 	}
 }
 
+func (si *SidecarInjector) injectMetadataPrefetchSidecarContainer(pod *corev1.Pod, config *Config, injectAsNativeSidecar bool) {
```
hime, line 97: i dont think we should respect the injectAsNativeSidecar parameter. Rather, we should make this decision based on the location of the `gke-gcsfuse-sidecar`

hime: since we are injecting in the same flow, injectAsNativeSidecar inherently tells us the location of our fuse sidecar.

pkg/webhook/sidecar_spec.go
hime: see injection_test.go for the integration test.

cmd/metadata_prefetch/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG TARGETPLATFORM
+
+# Copy existing binaries.
+COPY --from=debian /bin/ls /bin/ls
+
+# Copy dependencies.
+COPY --from=debian /lib/x86_64-linux-gnu/libselinux.so.1 /lib/x86_64-linux-gnu/libselinux.so.1
+COPY --from=debian /lib/x86_64-linux-gnu/libc.so.6 /lib/x86_64-linux-gnu/libc.so.6
+COPY --from=debian /lib/x86_64-linux-gnu/libpcre2-8.so.0 /lib/x86_64-linux-gnu/libpcre2-8.so.0
+COPY --from=debian /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2
```
saikat-royc, line 38: Can we run an ldd check to ensure all the dependencies are captured. e.g https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/cmd/csi_driver/Dockerfile#L68.

cmd/metadata_prefetch/main.go
saikat-royc: our e2e tests should be sufficient to ensure pod teardown. beyond any light weight unit tests, we should create a backlog task of a e2e test to pod termination while ls is running (as a fast followup after launch). Have we validate manually that pod termination works while ls is in progress?

cmd/metadata_prefetch/main.go
```diff
@@ -0,0 +1,73 @@
...
+	err := cmd.Start()
+	if err == nil {
+		klog.Info("Running ls on bucket(s)")
```
saikat-royc: mention the volume path in the log

cmd/metadata_prefetch/main.go
```diff
@@ -0,0 +1,73 @@
...
+		err = cmd.Wait()
+		if err != nil {
+			klog.Errorf("Error while executing ls command: %v", err)
+		} else {
+			klog.Info("Metadata prefetch complete.")
```
saikat-royc: "." can be removed . same for other such logs

cmd/metadata_prefetch/main.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	} else {
+		klog.Errorf("Error starting ls command: %v.", err)
+	}
+
+	klog.Info("Going to sleep...")
+
+	// Keep the process running.
+	select {}
```
saikat-royc, line 83: please manually verify the cpu usage is minimal once the container invokes select{} (if not already done)

cmd/webhook/main.go
```diff
@@ -103,6 +104,9 @@ func main() {
 	// Setup Informer
 	informerFactory := informers.NewSharedInformerFactory(client, resyncDuration)
 	nodeLister := informerFactory.Core().V1().Nodes().Lister()
+	// need tests that use these listers.
```
saikat-royc: do you plan to add unit tests for it in this pr?

cmd/metadata_prefetch/Dockerfile
```diff
@@ -0,0 +1,44 @@
+# Copyright 2018 The Kubernetes Authors.
```
msau42, line 1: I think we can remove k8s header from new files.

cmd/metadata_prefetch/main.go
hime: tested manually. also added functional tests for termination in aut_termination testsuite.

cmd/metadata_prefetch/main.go
```diff
@@ -0,0 +1,73 @@
...
+		<-sigs
+		klog.Info("Caught SIGTERM signal: Terminating...")
+		cancel()
+
+		os.Exit(0) // Exit gracefully
+	}()
+
+	// Start the "ls" command in the background.
+	// All our volumes are mounted under the /volumes/ directory.
+	cmd := exec.CommandContext(ctx, "ls", "-R", "/volumes/")
```
msau42: Do we actually need to exec into a shell or can we use https://pkg.go.dev/io/fs? Then we could also use distroless without a shell.

pkg/webhook/client.go
```diff
@@ -0,0 +1,79 @@
...
+	// IsPreprovisionCSIVolume checks whether the volume is a pre-provisioned volume for the desired csiDriver.
```
saikat-royc: the comment can go before line 26

cmd/webhook/main.go
msau42: Is this a TODO?

cmd/metadata_prefetch/main.go
hime: tracking bug internally.

pkg/webhook/sidecar_spec.go
hime: we do not have a flow setup on OSS to use configmaps for setting sidecar spec. Hardcoding these in non-managed driver.

hime: Sidecar resources should be enough here based on testing and minimum requirements. Resolving...

pkg/webhook/sidecar_spec.go
hime: we want to keep these, need to add them again to PR. Non-blocking...

cmd/webhook/main.go
hime: Yes, need to delete stale comment. We have testing below

pkg/webhook/sidecar_spec.go
```diff
@@ -18,20 +18,27 @@ limitations under the License.
...
+
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 )
 
 const (
 	SidecarContainerName                  = "gke-gcsfuse-sidecar"
+	SidecarMetadataPrefetchName           = "gke-gcsfuse-metadata-prefetch"
```
saikat-royc: SidecarContainerMetadataPrefetchName

pkg/webhook/parsers.go
```diff
@@ -45,27 +45,29 @@ func ParseBool(str string) (bool, error) {
...
-func parseSidecarContainerImage(pod *corev1.Pod) (string, error) {
+//
+// Note: This methos MUST delete all containers using containerName in the Pod Spec.
+func ParseSidecarContainerImage(podSpec *corev1.PodSpec, containerName string) (string, error) {
 	var image string
 
-	// Find container named "gke-gcsfuse-sidecar" (SidecarContainerName), extract its image, and remove from list.
-	if index, present := containerPresent(pod.Spec.Containers, SidecarContainerName); present {
-		image = pod.Spec.Containers[index].Image
+	// Find container named "gke-gcsfuse-sidecar" (sidecarContainerName), extract its image, and remove from list.
```
saikat-royc: "gke-gcsfuse-sidecar" is a stale comment as we pass the containerName now

pkg/webhook/parsers.go
```diff
@@ -45,27 +45,29 @@ func ParseBool(str string) (bool, error) {
...
 
-		if image != "" {
-			copy(pod.Spec.Containers[index:], pod.Spec.Containers[index+1:])
-			pod.Spec.Containers = pod.Spec.Containers[:len(pod.Spec.Containers)-1]
+		if _, _, _, err := parsers.ParseImageName(image); err != nil {
+			return "", fmt.Errorf("could not parse input image: %q, error: %w", image, err)
 		}
 	}
 
 	// Remove any mention of gke-gcsfuse-sidecar from init container list.
```
saikat-royc: same here about gke-gcsfuse-sidecar

pkg/webhook/parsers.go
```diff
@@ -45,27 +45,29 @@ func ParseBool(str string) (bool, error) {
 //   - removes the container definition from the container list.
 //   - remove any mentions of "gke-gcsfuse-sidecar" from initContainer list.
 //   - return image
-func parseSidecarContainerImage(pod *corev1.Pod) (string, error) {
+//
+// Note: This methos MUST delete all containers using containerName in the Pod Spec.
+func ParseSidecarContainerImage(podSpec *corev1.PodSpec, containerName string) (string, error) {
```
saikat-royc: ParseSidecarContainerImage seems like a misnomer to me, since the function is also removing the container. 

also remove any mention of `gke-gcsfuse-sidecar` e.g L42, L43, L53 ...

pkg/webhook/parsers.go
saikat-royc: please explain in details in the comment why we are removing the container from the list

cmd/metadata_prefetch/main.go
hime: I actually ran some PoCs using this as one of the alternatives, the performance was really bad 🙃 (See design doc section regarding selection of base image)

pkg/webhook/mutatingwebhook.go
```diff
@@ -92,7 +94,7 @@ func (si *SidecarInjector) Handle(_ context.Context, req admission.Request) admi
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	if image, err := parseSidecarContainerImage(pod); err == nil {
+	if image, err := ParseSidecarContainerImage(&pod.Spec, SidecarContainerName); err == nil {
```
saikat-royc: gkeGcsFuseSidecarImage

pkg/webhook/injection.go
```diff
@@ -63,23 +82,57 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+	var containerSpec corev1.Container
+	var index int
+
+	// We allow sidecar to be present on regular container list to support
```
saikat-royc: let's not make an assumption that user would only inject the image (private or public) in regular container. we should check both the regular container and init container for existing user provided image and container spec

pkg/webhook/injection.go
```diff
@@ -63,23 +82,57 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+	// Privately Hosted Sidecar Image feature for clusters running with limited internet access.
+	privateMetadataPrefetchSidecarImage, err := ParseSidecarContainerImage(&pod.Spec, SidecarMetadataPrefetchName)
```
saikat-royc: `userProvidedContainerImage` is a more suitable name for this variable. Put additional context about private image and if needed pointer to documentation link  in comments

pkg/webhook/injection.go
```diff
@@ -63,23 +82,57 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+		klog.Errorf("failed to get privately hosted metadata prefetch image... skipping injection.")
+	}
+
+	if privateMetadataPrefetchSidecarImage != "" {
+		config.MetadataContainerImage = privateMetadataPrefetchSidecarImage
+	}
+
+	if injectAsNativeSidecar {
+		containerSpec = si.GetNativeMetadataPrefetchSidecarContainerSpec(pod, config.MetadataContainerImage)
+		index = getInjectIndexAfterContainer(pod.Spec.InitContainers, SidecarContainerName)
```
saikat-royc: GkeGcsFuseSidecarContainerName

pkg/webhook/injection.go
```diff
@@ -63,23 +82,57 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
 
+func (si *SidecarInjector) injectMetadataPrefetchSidecarContainer(pod *corev1.Pod, config *Config, injectAsNativeSidecar bool) {
+	var containerSpec corev1.Container
+	var index int
+
+	// We allow sidecar to be present on regular container list to support
+	// Privately Hosted Sidecar Image feature for clusters running with limited internet access.
+	privateMetadataPrefetchSidecarImage, err := ParseSidecarContainerImage(&pod.Spec, SidecarMetadataPrefetchName)
+	if err != nil {
+		klog.Errorf("failed to get privately hosted metadata prefetch image... skipping injection.")
```
saikat-royc: this can be a info or warning since we donot always manually inject containers

cmd/metadata_prefetch/main.go
hime: Done.

pkg/webhook/injection.go
hime: Updated comments here and in the function.

pkg/webhook/injection.go
hime: This should be an error. We only emit this error when we're sure the user provided a custom sidecar image that was invalid.

pkg/webhook/parsers.go
hime: Renamed method to `ExtractContainerImageAndDeleteContainer`. Removed any stale comments.

pkg/webhook/sidecar_spec.go
hime: renamed to `MetadataPrefetchSidecarName`

cmd/metadata_prefetch/Dockerfile
hime: We have a check that requires this header to be present, created https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/367 to remove this requirement.

cmd/metadata_prefetch/Dockerfile
hime: Good idea! Done.

cmd/webhook/main.go
hime: client_test.go has unit tests for these listers.

pkg/webhook/sidecar_spec.go
hime: Done.

pkg/webhook/volumes.go
```diff
@@ -0,0 +1,64 @@
...
+	if pvc == nil {
+		return false, nil, nil
+	}
+	pvcName := pvc.ClaimName
+	pvcObj, err := si.GetPVC(namespace, pvcName)
+	if err != nil {
+		return false, nil, err
+	}
+
+	// Check if the PVC is a preprovisioned parallelstore volume.
```
saikat-royc: gcsfuse volume

pkg/webhook/client.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, err // no additional context needed for error.
+	}
+
+	if pv.Spec.CSI == nil {
+		return nil, fmt.Errorf("perisistentVolume is not backed by %s", csiDriver)
+	}
+
+	// PV is using dynamic mounting, which does not count as a preprovisioned volume.
+	// See details: https://cloud.google.com/storage/docs/gcsfuse-mount#dynamic-mount
+	if csiDriver == gcsFuseCsiDriverName && pv.Spec.CSI.VolumeHandle == "_" {
```
saikat-royc: I don't understand this logic, why are we returning error if volumeHandle is "_". Can you please elaborate? 
Are we saying that for dynamic mounting, PV/PVC way of expressing that intent is not supported?

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +113,95 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+	}
+
+	for _, v := range pod.Spec.Volumes {
+		isGcsFuseCSIVolume, volumeAttributes, err := si.isGcsFuseCSIVolume(v, pod.Namespace)
+		if err != nil {
+			klog.Errorf("failed to determine if %s is a GcsFuseCSI backed volume: %v", v.Name, err)
+		}
+
+		if isGcsFuseCSIVolume {
+			enableMetaPrefetchRaw, ok := volumeAttributes[gcsFuseMetadataPrefetchOnMountVolumeAttribute]
```
saikat-royc, line 201: When  metadata preftch flag is setup for volumes with "_" handle, we should flag a warning, that metadata prefetch will ignore this volume.

pkg/webhook/client.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, errors.New("persistentVolume is setup using dynamic mounting")
+	}
+
+	// Returns false when PV - PVC pair was created for a different csi driver or different storage type.
```
saikat-royc, line 59: this check should go before the dynamic mounting volume handle

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +113,95 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+			Limits:   limits,
+			Requests: requests,
+		},
+		VolumeMounts: []corev1.VolumeMount{},
+	}
+
+	for _, v := range pod.Spec.Volumes {
+		isGcsFuseCSIVolume, volumeAttributes, err := si.isGcsFuseCSIVolume(v, pod.Namespace)
+		if err != nil {
+			klog.Errorf("failed to determine if %s is a GcsFuseCSI backed volume: %v", v.Name, err)
```
saikat-royc, line 191: seem to me like we will write error logs for volumes which are not of type gcsfusecsi (https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/360/files#diff-b6aadbfe5e9536eb570a91f11fea0ba3127177b43e0b865f2389ba9540bebe33R68). This error log can be misleading during debugging.


Any volume for which csidriver name is not the same can simply be skipped in the  GetPreprovisionCSIVolume (similar to how you are skipping it for csi ephemeral volume which does not match the gcsfuse csidriver name https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/360/files#diff-838b88d01a3358247a8414ffe005610bd045beca211c41cb74ad9996a984a832R36)

pkg/webhook/volumes.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	gcsFuseCsiDriverName = "gcsfuse.csi.storage.gke.io"
+)
+
+func (si *SidecarInjector) isGcsFuseCSIVolume(volume corev1.Volume, namespace string) (bool, map[string]string, error) {
+	// Check if it is ephemeral volume.
+	if volume.CSI != nil {
+		if volume.CSI.Driver == gcsFuseCsiDriverName {
+			return true, volume.CSI.VolumeAttributes, nil
```
saikat-royc: what about csi ephemeral volume which sets the "_" volume handle. that should be skipped as well for metadata prefetch

saikat-royc:
Please squash commits before merge

pkg/webhook/sidecar_spec.go
hime: Done.

pkg/webhook/sidecar_spec.go
hime: Seems that [here](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/metadata-prefetch/pkg/webhook/volumes.go#L61) is the only place we are logging failures. This log is only run when we return an error from `GetPreprovisionCSIVolume`. I modified `GetPreprovisionCSIVolume` to only return errors when it is obvious that there's a misconfiguration in the Volume. Previously it would return an error when the volume type was not gcsfusecsi.

pkg/webhook/volumes.go
hime: Dynamic mount is now checked inside `isGcsFuseVolume` for both ephemeral and PV.

pkg/webhook/client.go
hime: This function should only check if its a persistent volume. Moved this check to `isGcsFuseVolume`

cmd/metadata_prefetch/Dockerfile
hime: Will remove when #367 is merged.

pkg/webhook/client.go
```diff
@@ -0,0 +1,78 @@
...
+	if csiDriver == "" {
+		return pv, false, errors.New("csiDriver is empty, cannot verify storage type")
+	}
+
+	if pvc == nil {
+		return pv, false, errors.New("pvc is nil, cannot get pv")
+	}
+
+	if pvc.Spec.VolumeName == "" {
+		return pv, false, errors.New("pvc.spec.volumeName is not set, cannot find pv")
```
hime: Not sure if we should return an error here... a PVC is still valid without a volumeName, because it can point to a dynamically provisioned volume

hime: No longer returning any errors. Resolving.

pkg/webhook/client.go
```diff
@@ -0,0 +1,78 @@
...
+	if pvc == nil {
+		return pv, false, errors.New("pvc is nil, cannot get pv")
+	}
+
+	if pvc.Spec.VolumeName == "" {
+		return pv, false, nil
+	}
+
+	// GetPV returns an error if pvc.Spec.VolumeName is not empty and the associated PV object is not found in the API server.
+	fetchedPv, err := si.GetPV(pvc.Spec.VolumeName)
```
saikat-royc: we can reuse the same pv variable that we defined in L35

pkg/webhook/volumes.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	gcsFuseCsiDriverName = "gcsfuse.csi.storage.gke.io"
+)
+
+func (si *SidecarInjector) isGcsFuseCSIVolume(volume corev1.Volume, namespace string) (bool, bool, map[string]string, error) {
```
saikat-royc, line 36: for readability maybe we should explain the return fields, what each bool means

pkg/webhook/client.go
```diff
@@ -0,0 +1,78 @@
...
+	var pv corev1.PersistentVolume
+	if csiDriver == "" {
+		return pv, false, errors.New("csiDriver is empty, cannot verify storage type")
+	}
+
+	if pvc == nil {
+		return pv, false, errors.New("pvc is nil, cannot get pv")
+	}
+
+	if pvc.Spec.VolumeName == "" {
```
saikat-royc, line 45: why is this returning nil?

pkg/webhook/injection.go
```diff
@@ -63,23 +86,65 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+
+	injected, _ := validatePodHasSidecarContainerInjected(MetadataPrefetchSidecarName, pod, []corev1.Volume{}, []corev1.VolumeMount{})
+	if injected {
+		klog.Infof("%s is already injected, skipping injection...", MetadataPrefetchSidecarName)
+
+		return
+	}
+
+	// Extract user provided metadata prefetch sidecar image.
+	userProvidedMetadataPrefetchSidecarImage, err := ExtractImageAndDeleteContainer(&pod.Spec, MetadataPrefetchSidecarName)
```
saikat-royc, line 109: please create a bug to update https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/cloud-storage-fuse-csi-driver#private-image with metadata sidecar (starting gke version ...)

pkg/webhook/injection.go
```diff
@@ -63,23 +86,65 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+	if err != nil {
+		klog.Errorf("failed to get user provided metadata prefetch image... skipping injection.")
```
saikat-royc: "..." not needed. save unnecessary chars and reduce log storage. same for "." in the end :)

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +113,101 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+	}
+
+	for _, v := range pod.Spec.Volumes {
+		isGcsFuseCSIVolume, isDynamicMount, volumeAttributes, err := si.isGcsFuseCSIVolume(v, pod.Namespace)
+		if err != nil {
+			klog.Errorf("failed to determine if %s is a GcsFuseCSI backed volume: %v", v.Name, err)
+		}
+
+		if isDynamicMount {
+			klog.Warningf("dynamic mount set for %s, this is not supported for metadata prefetch. skipping volume...", v.Name)
```
saikat-royc: remove "..."

pkg/webhook/sidecar_spec.go
```diff
@@ -118,6 +113,101 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
...
+	container := si.GetMetadataPrefetchSidecarContainerSpec(pod, image)
+	container.Env = append(container.Env, corev1.EnvVar{Name: "NATIVE_SIDECAR", Value: "TRUE"})
+	container.RestartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
+
+	return container
+}
+
+func getMetadataPrefetchConfig(image string) *Config {
+	return &Config{
+		CPURequest:              resource.MustParse("10m"),
```
saikat-royc: please create constants for these e.g metadataSidecarCPURequest

saikat-royc:
For the new tests added for this feature in generateTestSkip() 1. we need to skip managed driver tests 2. eventually add a min version (based on the expected gke component version)

pkg/webhook/client.go
hime: @saikat-royc we are returning a nil error because the volume is valid. A pvc can have this field missing when its requesting a dynamically provisioned volume and it is not yet bound.

pkg/webhook/injection.go
hime: Done.

pkg/webhook/injection.go
hime: Done.

pkg/webhook/sidecar_spec.go
hime: Done

pkg/webhook/volumes.go
hime: Done.

pkg/webhook/client.go
hime: @saikat-royc will leave up to you to resolve this comment.

saikat-royc: Ack. Please put a comment in the code for this and resolve this conversation

PR #368, opened 2024-11-07
Add platform flag to Golang binary to honor build platform

uriel-guzman:
Adding this flag fixes a specific issue we observed where attempting to build a linux/ARM64 driver would either time out for several hours or produce the error: `.buildkit_qemu_emulator: /bin/sh: Invalid ELF image for this architecture`.

Adding this line ensures that our build platform is honored in our Dockerfile.

More information about this flag: https://docs.docker.com/build/building/multi-platform/#cross-compilation

### DIFF
--- a/cmd/csi_driver/Dockerfile
+++ b/cmd/csi_driver/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Build driver go binary
-FROM golang:1.22.7 AS driver-builder
+FROM --platform=$BUILDPLATFORM golang:1.22.7 AS driver-builder
 
 ARG STAGINGVERSION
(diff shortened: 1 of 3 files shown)


hime:
Where is $BUILDPLATFORM defined? Do we need to update the Makefile too?

uriel-guzman:
> Where is $BUILDPLATFORM defined? Do we need to update the Makefile too?

When you build a Docker image, Docker automatically detects the platform (operating system and architecture) of your build environment. This information is stored in the BUILDPLATFORM argument.

image

Source: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

PR #405, opened 2024-12-10
Run gRPC e2e integration tests.

hime:
Run gRPC protocol for e2e tests.

Subsequent PR contents:
-  Run grpc on any additional functional testsuites
- Run grpc on performance testsuites.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -41,6 +41,7 @@ var (
 	err            error
 	c              clientset.Interface
 	m              metadata.Service
+	clientProtocol = flag.String("client-protocol", "http", "the test bucket location")
 	bucketLocation = flag.String("test-bucket-location", "us-central1", "the test bucket location")
 	skipGcpSaTest  = flag.Bool("skip-gcp-sa-test", true, "skip GCP SA test")
 	apiEnv         = flag.String("api-env", "prod", "cluster API env")
@@ -114,7 +115,7 @@ var _ = ginkgo.Describe("E2E Test Suite", func() {
 		testsuites.InitGcsFuseCSIMetadataPrefetchTestSuite,
 	}
 
-	testDriver := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, false)
+	testDriver := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, false, *clientProtocol)
 
 	ginkgo.Context(fmt.Sprintf("[Driver: %s]", testDriver.GetDriverInfo().Name), func() {
 		storageframework.DefineTestSuites(testDriver, GCSFuseCSITestSuites)
@@ -126,7 +127,7 @@ var _ = ginkgo.Describe("E2E Test Suite", func() {
 		testsuites.InitGcsFuseCSIGCSFuseIntegrationFileCacheParallelDownloadsTestSuite,
 	}
 
-	testDriverHNS := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, true)
+	testDriverHNS := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, true, *clientProtocol)
 
 	ginkgo.Context(fmt.Sprintf("[Driver: %s HNS]", testDriverHNS.GetDriverInfo().Name), func() {
 		storageframework.DefineTestSuites(testDriverHNS, GCSFuseCSITestSuitesHNS)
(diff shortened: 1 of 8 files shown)


test/e2e/e2e_test.go
```diff
@@ -126,9 +126,27 @@ var _ = ginkgo.Describe("E2E Test Suite", func() {
...
 		storageframework.DefineTestSuites(testDriverHNS, GCSFuseCSITestSuitesHNS)
 	})
+
+	testDriverHNSAndGrpc := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, true, true)
+
+	ginkgo.Context(fmt.Sprintf("[Driver: %s HNS + GRPC]", testDriverHNSAndGrpc.GetDriverInfo().Name), func() {
+		storageframework.DefineTestSuites(testDriverHNSAndGrpc, GCSFuseCSITestSuitesHNS)
+	})
+
+	GCSFuseCSITestSuitesGRPC := []func() storageframework.TestSuite{
```
hime: @saikat-royc any other testsuites we should consider running with grpc? Some of my considerations could be istio and metric tests.

test/e2e/main.go
```diff
@@ -56,7 +56,7 @@ var (
 	ginkgoFocus         = flag.String("ginkgo-focus", "", "pass to ginkgo run --focus flag")
 	ginkgoSkip          = flag.String("ginkgo-skip", "", "pass to ginkgo run --skip flag")
 	ginkgoProcs         = flag.String("ginkgo-procs", "10", "pass to ginkgo run --procs flag")
-	ginkgoTimeout       = flag.String("ginkgo-timeout", "4h", "pass to ginkgo run --timeout flag")
+	ginkgoTimeout       = flag.String("ginkgo-timeout", "8h", "pass to ginkgo run --timeout flag")
```
saikat-royc: please make a note to check the timeout, and the frequency of test runs for the internal prow test config

test/e2e/specs/specs.go
```diff
@@ -89,7 +89,7 @@ const (
 	pollInterval     = 1 * time.Second
 	pollTimeout      = 1 * time.Minute
 	pollIntervalSlow = 10 * time.Second
-	pollTimeoutSlow  = 20 * time.Minute
+	pollTimeoutSlow  = 30 * time.Minute
```
saikat-royc: what is the reason for this increase?

test/e2e/testsuites/gcsfuse_integration.go
```diff
@@ -133,6 +140,11 @@ func (t *gcsFuseCSIGCSFuseIntegrationTestSuite) DefineTests(driver storageframew
 			}
 		}
 
+		// GRPC is supported after v2.5.1
+		if !v.AtLeast(version.MustParseSemantic("v2.5.1-gke.0")) && grpcEnabled(driver) {
```
saikat-royc: we are using the gke builds because these tests are intended to run only with gke managed csi driver?  and the OSS driver head uses gcsfuse master head? Is this correct?

test/e2e/e2e_test.go
saikat-royc: As discussed let's create a new prow job spec with gRPC enabled driver, and run the existing test suites. So we can keep the timeout of the current prow job to its existing value of 4h, and get high coverage of the existing testsuite test cases with gRPC

test/e2e/e2e_test.go
```diff
@@ -114,7 +114,7 @@ var _ = ginkgo.Describe("E2E Test Suite", func() {
 		testsuites.InitGcsFuseCSIMetadataPrefetchTestSuite,
 	}
 
-	testDriver := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, false)
+	testDriver := specs.InitGCSFuseCSITestDriver(c, m, *bucketLocation, *skipGcpSaTest, false, false)
```
saikat-royc: so if we want to run a separate prow job  with gRPC enabled, this bool would need to become a flag, and that would be setup in the run-e2e*.sh is my understanding

test/e2e/specs/specs.go
hime: reverted.

test/e2e/testsuites/gcsfuse_integration.go
hime: I also had this question. When we build for the non-managed driver, we build gcsfuse from master. When we build from master, we assign that build a tag version of 999. This automatically qualifies the non-managed driver to run all the tests. This tag is only really applicable when we run the testsuite against the managed driver.

So yes, the tests run against all non-managed driver versions, and for selected gke managed driver versions.

saikat-royc: Maybe add a comment so we dont forget

hime: Done.

PR #410, opened 2024-12-12
Remove setting client-protocol for performance tests.

hime:
cc. @saikat-royc 

http1 is the default protocol used, so it is safe to remove as it will still be used by default.


saikat-royc:
Please add details in the PR why we had to remove this, was there a test failure?

PR #413, opened 2024-12-17
Filtering informerFactory to only return relevant PodSpec information

halimsam:
Filtering informerFactory to only return relevant PodSpec information to improve GCSFUSE CSI driver memory usage for pods and deployments with larger Pod Specs.

### DIFF
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -78,11 +78,48 @@ func New(kubeconfigPath string, informerResyncDurationSec int) (Interface, error
 }
 
 func (c *Clientset) ConfigurePodLister(nodeName string) {
-	trimManagedFields := func(obj interface{}) (interface{}, error) {
+	trim := func(obj interface{}) (interface{}, error) {
 		if accessor, err := meta.Accessor(obj); err == nil {
 			if accessor.GetManagedFields() != nil {
 				accessor.SetManagedFields(nil)
 			}
+		}
+
+		// We are filtering only for relevant PodSpec info to optimize memory usage.
+		// Relevant info is for NodePublishVolume calls:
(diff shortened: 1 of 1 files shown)


mborsz:
LGTM

Just to confirm, you are using 1.27+ informer library, right? Only in that version you can change the object in place, see https://github.com/kubernetes/client-go/blob/67da6d1a4174c5f881389516fd53f5e13d3eaf85/tools/cache/delta_fifo.go#L146-L147

halimsam:
> LGTM
> 
> Just to confirm, you are using 1.27+ informer library, right? Only in that version you can change the object in place, see https://github.com/kubernetes/client-go/blob/67da6d1a4174c5f881389516fd53f5e13d3eaf85/tools/cache/delta_fifo.go#L146-L147

Considering this is [1.27/v0.27.0](https://github.com/kubernetes/client-go/releases/tag/v0.27.0), we are using [v0.30.3](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/909db593a1323ac0b175e036955bbe047ccf1597/go.mod#L29). Thanks for the LGTM!

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -78,13 +78,47 @@ func New(kubeconfigPath string, informerResyncDurationSec int) (Interface, error
...
 			if accessor.GetManagedFields() != nil {
 				accessor.SetManagedFields(nil)
 			}
 		}
 
+		podObj, ok := obj.(*corev1.Pod)
+		if !ok {
+			return obj, nil
+		}
+
```
saikat-royc, line 95: put comments around why we are only filtering relevant parts of the pod spec, link to the node.go csi node publish

msau42:
I think we need to grab status as well in order to handle the non-sidecar container Job handling.

PR #417, opened 2024-12-27
Track restartPolicy in node pod informer.

hime:
In Kubernetes versions that do not support native sidecar, we put an exit file for the sidecar to read once all the containers in the pod have terminated. This behavior partly depends on the RestartPolicy of the Pod. Tracking of this policy was recently removed in #413. Adding it back to Pod.Spec.RestartPolicy in node server pod informer to fix regression in 1.28.

#### Tested
Ran non-managed driver on cluster running GKE v1.28.15-gke.1020000 and verified auto-termination testsuite succeeds.

cc. @saikat-royc @halimsam

### DIFF
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -115,11 +115,13 @@ func (c *Clientset) ConfigurePodLister(nodeName string) {
 
 		nodeName := podObj.Spec.NodeName
 		volumes := podObj.Spec.Volumes
+		restartPolicy := podObj.Spec.RestartPolicy
 		podObj.Spec = corev1.PodSpec{
 			NodeName:       nodeName,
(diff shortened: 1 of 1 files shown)


pkg/cloud_provider/clientset/clientset.go
```diff
@@ -115,11 +115,13 @@ func (c *Clientset) ConfigurePodLister(nodeName string) {
 
 		nodeName := podObj.Spec.NodeName
 		volumes := podObj.Spec.Volumes
+		restartPolicy := podObj.Spec.RestartPolicy
```
msau42, line 118: Do you also need to specify status fields?

hime: We are only filtering out fields inside Pod.Spec. Pod.Status is unaffected.

PR #420, opened 2025-01-03
Removing KLC delete dir test

raj-prince:
The rationale is to run minimum and important test case in GKE world, GCSFuse will run all the tests.


saikat-royc:
if stale please close this PR @raj-prince

PR #428, opened 2025-01-14
Use projected token volume for hostNetwork pods.

siyanshen:
## Why we need this change
Before this change, GCSFuse CSI driver does not support pods with hostNetwork:true. This is because the gcsfuse process runs as part of a sidecar container that is injected into the user pod. GCSFuse uses the [ADC](https://cloud.google.com/docs/authentication/application-default-credentials#attached-sa) workflow to fetch the necessary token to access a GCS bucket. However, with hostNetwork enabled, GKE metadata server cannot intercept the token requests for `GET /computeMetadata/v1/instance/service-accounts/default/token` endpoint.

## What is in this change
1.  For `HostNetwork=true` user pods, GCSFuse CSI webhook injects a projected SA token volume to the user pod.
2. `gke-gcsfuse-sidecar` container prepares a unix domain socket and starts a handler to serve requests on this token; invoke gcsfuse with a config option to point to the socket `gcs-auth:token-url:<path to the token>`
3. Token request handler in the `gke-gcsfuse-sidecar` handles token request from GCSFuse.

## Local set up and testing

**Test cases covered**
1. pod with hostNetwork=true: can I/O GCS bucket.
2. pod with hostNetwork=false: is not affected and can I/O GCS bucket.
3. pod with hostNetwork=true, with GCS bucket A mounted as ephemeral storage and Gcs bucket B mounted as PV: can access bucket A & B without issue.

**Setup**
1. Create a cluster with workload identity pool enabled & GCSFuse disabled.
2. Grant GCS permissions to an SA
```
kubectl create namespace <ns>
kubectl create serviceaccount <test-ksa-ns> \
    --namespace <ns>

gcloud storage buckets add-iam-policy-binding gs://<bucket> \
    --member "principal://iam.googleapis.com/projects/<project-number>/locations/global/workloadIdentityPools/<project-id>.svc.id.goog/subject/ns/<ns>/sa/<test-ksa-ns>" \
```
4. Build and install your GCSFuse image
```
make build-image-and-push-multi-arch REGISTRY=<your-registry> STAGINGVERSION=<your-dev-tag>
make install REGISTRY=<your-registry> PROJECT=<project-id> STAGINGVERSION=<your-dev-tag>
```
5. Create a user pod with `hostNetwork=true`, GCS bucket <bucket> mounted as a volume, and service account set as the one you created in step 2. Check I/O to your bucket from a container in your pod.


deploy/base/webhook/deployment.yaml
```diff
@@ -45,7 +45,7 @@ spec:
               drop:
                 - ALL
           image: gke.gcr.io/gcs-fuse-csi-driver-webhook
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: Always
```
mattcary: Did you switch the pull policy for dev? In practice I'd think we'd want IfNotPresent

cmd/sidecar_mounter/main.go
```diff
@@ -117,3 +136,130 @@ func main() {
...
+		return
+	}
+	klog.Infof("created a listener using the socket path %s", tokenURLSocketPath)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", func(w http.ResponseWriter, _ *http.Request) {
+		klog.Infof("got request on socket path %s", tokenURLSocketPath)
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		k8stoken, err := getK8sToken("/gcsfuse-sa-token/token")
```
mattcary: Why not use the constant defined in webhook/sidecar_spec.go?

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -276,6 +278,11 @@ func (mc *MountConfig) prepareConfigFile() error {
 			}
 		}
 	}
+	if mc.HostNetwork {
+		configMap["gcs-auth"] = map[string]interface{}{
+			"token-url": "unix:///gcsfuse-tmp/.volumes/gcs-fuse-csi-ephemeral/token.sock",
```
mattcary: Why not use a constant for this? Probably define it here, then it can be used in the main.go.

cmd/sidecar_mounter/main.go
```diff
@@ -117,3 +136,130 @@ func main() {
 
 	klog.Info("exiting sidecar mounter...")
 }
+
```
mattcary: It's not great having all this logic in a main.go, it makes it untestable.

That said I don't know possible it is to test this stuff anyway. But as a general code smell thing, I get nervous when there's anything nontrivial in main.go.

deploy/base/webhook/deployment.yaml
siyanshen: Changed it back

cmd/sidecar_mounter/main.go
siyanshen: Good point- refactored it

cmd/sidecar_mounter/main.go
siyanshen: Moved it to pkg/sidecar_mounter/sidecar_mounter.go

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -275,3 +288,130 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
 
 	return nil
 }
+
+func getK8sToken(tokenPath string) (string, error) {
+	token, err := ioutil.ReadFile(tokenPath)
```
kislaykishore: ioutil is deprecated. Please use os.ReadFile instead

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -275,3 +288,130 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
...
+		fmt.Fprint(w, string(jsonToken))
+	})
+
+	server := http.Server{
+		Handler:      mux,
+		ReadTimeout:  10 * time.Second,
+		WriteTimeout: 10 * time.Second,
+	}
+
+	if err := server.Serve(socket); err != nil {
```
kislaykishore: Is there a way to stop the server gracefully during shutdown?

cmd/webhook/main.go
```diff
@@ -44,7 +44,7 @@ var (
 	certDir                 = flag.String("cert-dir", "/etc/tls-certs", "The directory that contains the server key and certificate.")
 	certName                = flag.String("cert-name", "cert.pem", "The server certificate name.")
 	keyName                 = flag.String("key-name", "key.pem", "The server key name.")
-	imagePullPolicy         = flag.String("sidecar-image-pull-policy", "IfNotPresent", "The default image pull policy for gcsfuse sidecar container.")
+	imagePullPolicy         = flag.String("sidecar-image-pull-policy", "Always", "The default image pull policy for gcsfuse sidecar container.")
```
hime: What's the context behind this change?

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -85,6 +85,7 @@ func (ts *GCPTokenSource) fetchK8sSAToken(ctx context.Context) (*oauth2.Token, e
 		return nil, fmt.Errorf("could not find token for the identity pool %q", ts.meta.GetIdentityPool())
 	}
 
+	klog.Infof("Calling Create SA token for SA %s/%s", ts.k8sSANamespace, ts.k8sSAName)
```
hime: Can we remove or mark under debug logs flag to prevent spam logging?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -275,3 +288,130 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
 
 	return nil
 }
+
+func getK8sToken(tokenPath string) (string, error) {
```
hime: can we rename to getK8sTokenFromFile or similar

pkg/sidecar_mounter/sidecar_mounter.go
kislaykishore: Serve always returns a non-nil error - https://pkg.go.dev/net/http#Serve. This means that the following log statement is incorrect.

pkg/webhook/mutatingwebhook.go
```diff
@@ -110,6 +113,15 @@ func (si *SidecarInjector) Handle(_ context.Context, req admission.Request) admi
 
 	// Inject container.
 	injectSidecarContainer(pod, config, injectAsNativeSidecar)
+
+	if pod.Spec.HostNetwork {
```
hime, line 117: Are there any cases in which the cx would want to keep using the GCE SA? (even though its not very safe)

pkg/webhook/mutatingwebhook.go
```diff
@@ -110,6 +113,15 @@ func (si *SidecarInjector) Handle(_ context.Context, req admission.Request) admi
...
+		projectID, err := metadata.ProjectIDWithContext(ctx)
+		if err != nil {
+			return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to get project id: %w", err))
+		}
+		pod.Spec.Volumes = append(pod.Spec.Volumes, GetSATokenVolume(projectID))
```
hime, line 122: do you think we should have the logic that injects volumes in one place?

pkg/sidecar_mounter/token_manager.go
```diff
@@ -0,0 +1,38 @@
...
+limitations under the License.
+*/
+
+package sidecarmounter
+
+import (
+	"golang.org/x/oauth2"
+)
+
+type TokenManager interface {
```
kislaykishore: Exported types should be documented - https://google.github.io/styleguide/go/decisions.html#doc-comments

pkg/webhook/sidecar_spec.go
```diff
@@ -27,21 +28,25 @@ import (
...
-	SidecarContainerCacheVolumeMountPath  = "/gcsfuse-cache"
+	GcsFuseSidecarName                     = "gke-gcsfuse-sidecar"
+	MetadataPrefetchSidecarName            = "gke-gcsfuse-metadata-prefetch"
+	SidecarContainerTmpVolumeName          = "gke-gcsfuse-tmp"
+	SidecarContainerTmpVolumeMountPath     = "/gcsfuse-tmp"
+	SidecarContainerBufferVolumeName       = "gke-gcsfuse-buffer"
+	SidecarContainerBufferVolumeMountPath  = "/gcsfuse-buffer"
+	SidecarContainerCacheVolumeName        = "gke-gcsfuse-cache"
+	SidecarContainerCacheVolumeMountPath   = "/gcsfuse-cache"
+	SidecarContainerSATokenVolumeName      = "gcsfuse-sa-token"  // #nosec G101
```
hime, line 39: Why do we have `// #nosec G101`? I'm guessing golint is complaining because we have the word token in the variableName right?

pkg/webhook/sidecar_spec.go
```diff
@@ -103,6 +113,13 @@ func GetNativeSidecarContainerSpec(c *Config) corev1.Container {
 func GetSidecarContainerSpec(c *Config) corev1.Container {
 	limits, requests := prepareResourceList(c)
 
+	var volumeMounts []corev1.VolumeMount
```
hime: you can simplify this a bit like so:

```
volumeMounts = ...

if hostnetwork {
  volumeMounts.append(...)
}
```

pkg/webhook/sidecar_spec.go
```diff
@@ -27,21 +28,25 @@ import (
...
 
 	// Webhook relevant volume attributes.
 	gcsFuseMetadataPrefetchOnMountVolumeAttribute = "gcsfuseMetadataPrefetchOnMount"
 
 	// See the nonroot user discussion: https://github.com/GoogleContainerTools/distroless/issues/443
-	NobodyUID = 65534
-	NobodyGID = 65534
+	NobodyUID           = 65534
+	NobodyGID           = 65534
+	tokenExpiryDuration = 600
```
hime: Curious, why 600? Is `tokenExpiryDuration` something that can be adjusted? if so, should the cx have the ability to adjust?

pkg/webhook/sidecar_spec.go
```diff
@@ -218,6 +236,27 @@ func (si *SidecarInjector) GetMetadataPrefetchSidecarContainerSpec(pod *corev1.P
 	return container
 }
 
+func GetSATokenVolume(projectID string) corev1.Volume {
+	saTokenVolume := corev1.Volume{
+		Name: SidecarContainerSATokenVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			Projected: &corev1.ProjectedVolumeSource{
```
hime, line 241: Is a `ProjectedVolume` a new feature? Sometimes when a new field is added on an API, it can get dropped after by older APIs that don't yet support it.

pkg/sidecar_mounter/token_manager.go
```diff
@@ -0,0 +1,38 @@
...
+	GetTokenSource(token *oauth2.Token) oauth2.TokenSource
+}
+
+type tokenManager struct{}
+
+func NewTokenManager() TokenManager {
```
kislaykishore: Where is this method being used?

cmd/webhook/main.go
kislaykishore: Why are we changing this?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -275,3 +288,130 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
...
+		"identitynamespace:%s.svc.id.goog:%s",
+		projectID,
+		onePlatformClusterResourceURL,
+	)
+	klog.Infof("audience: %s", audience)
+
+	return audience, nil
+}
+
+func StartTokenServer(ctx context.Context) {
```
kislaykishore: What's the behavior when the server failed to start? Will the GCS calls use node's auth scopes?

pkg/sidecar_mounter/token_source.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"golang.org/x/oauth2"
+)
+
+type TokenSource struct {
+	token *oauth2.Token
+}
+
+func (ts *TokenSource) Token() (*oauth2.Token, error) {
```
kislaykishore: How is this method invoked?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -20,23 +20,36 @@ package sidecarmounter
...
+	"golang.org/x/oauth2"
+	"google.golang.org/api/option"
+	"google.golang.org/api/sts/v1"
 	"k8s.io/klog/v2"
 )
 
-const metricEndpointFmt = "http://localhost:%v/metrics"
+const (
+	metricEndpointFmt  = "http://localhost:%v/metrics"
+	TokenURLSocketPath = "/gcsfuse-tmp/.volumes/gcs-fuse-csi-ephemeral/token.sock" // #nosec G101
```
hime: I'm a bit confused about `gcs-fuse-csi-ephemeral`. Is this supposed to be the volume name? If so, it should match the volume name. This caught my eye because a volume can also be PV.

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -234,7 +237,7 @@ func (mc *MountConfig) prepareMountArgs() {
 		klog.Warningf("got invalid arguments for volume %q: %v. Will discard invalid args and continue to mount.",
 			invalidArgs, mc.VolumeName)
 	}
-
+	klog.Infof("configFileFlagMap: %+v", mc.ConfigFileFlagMap)
```
hime: place behind debug flag or remove

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -275,3 +288,130 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
...
+	// Create a unix domain socket and listen for incoming connections.
+	socket, err := net.Listen("unix", TokenURLSocketPath)
+	if err != nil {
+		klog.Errorf("failed to create socket %q: %v", TokenURLSocketPath, err)
+
+		return
+	}
+	klog.Infof("created a listener using the socket path %s", TokenURLSocketPath)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", func(w http.ResponseWriter, _ *http.Request) {
```
hime, line 379: can we specify a path other than `"/"`. I think for metrics we use `/metrics`, we could do something similar here.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -275,3 +288,130 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
...
+		klog.Infof("got request on socket path %s", TokenURLSocketPath)
```
hime: each time we call this endpoint this log gets called. if we expect this endpoint to be called a lot, we should hide these logs behind a flag.

pkg/sidecar_mounter/sidecar_mounter_config_test.go
```diff
@@ -311,6 +311,44 @@ func TestPrepareConfigFile(t *testing.T) {
...
+				"file-cache": map[string]interface{}{
+					"max-size-mb":               10000,
+					"cache-file-for-range-read": true,
+				},
+				"metadata-cache": map[string]interface{}{
+					"stat-cache-max-size-mb": 1000,
+					"type-cache-max-size-mb": -1,
+				},
+				"cache-dir": "/gcsfuse-cache/.volumes/volume-name",
+				"gcs-auth":  map[string]interface{}{"token-url": "unix:///gcsfuse-tmp/.volumes/gcs-fuse-csi-ephemeral/token.sock"},
```
hime: can we add more tests using different volume names.

pkg/sidecar_mounter/sidecar_mounter_config.go
siyanshen: Done

pkg/webhook/sidecar_spec.go
siyanshen: Yeah that's correct

pkg/webhook/sidecar_spec.go
siyanshen: Looks like ProjectedVolumeSource has been introduced since 8 years ago. Is "Projected" field being dropped your concern? Or some other fields? https://github.com/kubernetes/kubernetes/commit/8fb1b71c6670b13669c365058546c9671c81b0c9#diff-41f03de739aac2ef36071b5a8ceefd4d8400b34f8eec6420941ab99e8ef48185

cmd/webhook/main.go
siyanshen: reverted

pkg/cloud_provider/auth/token_sources.go
siyanshen: Removed

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: done

pkg/webhook/sidecar_spec.go
siyanshen: Prolonged it to 3600-should be a more reasonable token expiry duration. See more about what this value does: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#launch-a-pod-using-service-account-token-projection

It's not very useful to make tokenExpiryDuration configurable.

pkg/sidecar_mounter/token_manager.go
siyanshen: Removed this file. Accidentally included it into the PR.

pkg/sidecar_mounter/token_manager.go
siyanshen: Removed this file

pkg/sidecar_mounter/token_source.go
siyanshen: Removed this file

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: Thanks for spotting this! Updated the implementation to support this feature for pods with different kinds of volumes

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: `mux.HandleFunc("/", ...)`: This registers the handler function to respond to requests made to the root path ("/").  So, a request to the socket path without any additional path elements would trigger this handler. 

We should keep this as is and serve the requests going to the socket path.

pkg/sidecar_mounter/sidecar_mounter_config.go
siyanshen: Done

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: fixed

pkg/webhook/sidecar_spec.go
hime: Thanks, I think we should be okay. When an API field is very new, some k8s libraries don't support them and at times remove the "invalid" field from the spec. This happened when we introduced native sidecars to GCSFuse soon after they went GA, other components were using an older k8s api library which wasn't compatible.

pkg/webhook/sidecar_spec.go
hime: When this expires, we have the ability to renew because of the `StartTokenServer` call. Do we plan to add e2e tests in a followup PR to test the functionality of the token server? If we have a `TokenServer`, do we favor a shorter or longer `tokenExpiryDuration`?

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: The pod will not have access to GCS bucket in this case

pkg/webhook/mutatingwebhook.go
siyanshen: Is that a common use case? This implementation will not be able to accommodate that

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: And mounting will fail.

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: This implementation follows the standard pattern across the repo. e.g.: metrics server: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e5d8871c76dcbed3a213eab89a2365568899d097/pkg/sidecar_mounter/sidecar_mounter.go#L249

Can you give me an example of gracefully shutting down a server? Is there a significant benefit for doing that?

pkg/webhook/sidecar_spec.go
siyanshen: I plan to add an e2e test to test GCS bucket access with hostnetwork enabled pods. That should cover the functionality of the token server.

Re`tokenExpiryDuration`- we will have to weigh in the security benefits of shorter token lifetimes against the performance benefits of longer lifetimes. Kubernetes' approach provides a good starting point- it will refresh when it hits 80% of TTL or 24h (whichever is shorter). I think 3600s(1h) is a nice balance point between security and network traffic overhead.

pkg/webhook/mutatingwebhook.go
siyanshen: Could you elaborate?

pkg/webhook/mutatingwebhook.go
hime: That's the workaround that some customers are currently using. I don't think its common though.

pkg/webhook/mutatingwebhook.go
hime: Sorry, disregard.

pkg/webhook/sidecar_spec.go
hime: Sounds good. Thanks!

cmd/sidecar_mounter/main.go
```diff
@@ -60,7 +61,12 @@ func main() {
 		// 1. different gcsfuse logs mixed together.
 		// 2. memory usage peak.
 		time.Sleep(1500 * time.Millisecond)
-		mc := sidecarmounter.NewMountConfig(sp)
+		if *hostNetwork {
+			tp := filepath.Join(filepath.Dir(sp), sidecarmounter.TokenFileName)
+			klog.Infof("Pod has hostNetwork enabled. Starting Token Server on %s.", tp)
+			go sidecarmounter.StartTokenServer(ctx, tp)
```
hime: Maybe move inside mounter.Mount to match pattern.

cmd/sidecar_mounter/main.go
```diff
@@ -34,6 +34,7 @@ import (
 )
 
 var (
+	hostNetwork    = flag.Bool("host-network", false, "pod hostNetwork configuration")
```
hime, line 37: If this goes inside mounter.Mount, do we need this flag? I think we pass (or can pass) hostNetwork flag through mountConfig.

hime: This can simplify the sidecar injection logic too.

siyanshen: We still need this to pass to mount config

cmd/sidecar_mounter/main.go
siyanshen: done

pkg/webhook/mutatingwebhook.go
siyanshen: The problem we are trying to solve is involuntarily having the pod to use the node default SA to request access token. It will not affect customers when they do the workaround steps, but those steps will be unnecessary.

cmd/sidecar_mounter/main.go
hime: You are using the new HostNetwork flag to:
1. set mc.HostNetwork
2. setup gcs-auth:token-url:...

To avoid creating flags, we can determine if the Pod supports this feature on `gcs-fuse-csi-driver` container by checking if `hostNetwork` is enabled on the pod and the volume we need is injected. We have a pod informer that has that information.

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e5d8871c76dcbed3a213eab89a2365568899d097/pkg/csi_driver/node.go#L137

We can then pass the mountConfig from `gcs-fuse-csi-driver` -> `gke-gcsfuse-sidecar`

driver sending mc to sidecar:

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e5d8871c76dcbed3a213eab89a2365568899d097/pkg/csi_mounter/csi_mounter.go#L86

sidecar receiving mc from driver:

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e5d8871c76dcbed3a213eab89a2365568899d097/pkg/sidecar_mounter/sidecar_mounter_config.go#L129

After that, we can process mc to set the configmap with the uds path

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e5d8871c76dcbed3a213eab89a2365568899d097/pkg/sidecar_mounter/sidecar_mounter_config.go#L151

pkg/webhook/mutatingwebhook.go
hime: Gotcha. Thank you for explaining.

cmd/sidecar_mounter/main.go
siyanshen: Hey Jaime, thanks for the suggestion. Somehow the resulting pod of k8sclients.GetPod(...) does not persist the HostNetwork value. It is "false" despite my test pod setting. It might be worthwhile to derive the HostNetwork from `pod.Annotations["kubectl.kubernetes.io/last-applied-configuration"]`, but it might be very messy.

This is the last-applied annotation.
`{"apiVersion":"v1","kind":"Pod","metadata":{"annotations":{"gke-gcsfuse/volumes":"true"},"name":"test-pv-pod-hnw","namespace":"ns1"},"spec":{"containers":[{"command":["sleep","3600"],"image":"busybox","name":"busybox","volumeMounts":[{"mountPath":"/data","name":"gcp-cloud-storage-pvc"},{"mountPath":"/dataEph","name":"gcs-fuse-csi-ephemeral"}]}],"hostNetwork":true,"serviceAccountName":"test-ksa-ns1","volumes":[{"name":"gcp-cloud-storage-pvc","persistentVolumeClaim":{"claimName":"gcp-cloud-storage-csi-static-pvc"}},{"csi":{"driver":"gcsfuse.csi.storage.gke.io","volumeAttributes":{"bucketName":"test-wi-host-network-2","mountOptions":"implicit-dirs"}},"name":"gcs-fuse-csi-ephemeral"}]}}`


So far the cleanest and safest solution is passing the bool flag from webhook. What do you think? Let me know.

siyanshen: Synced with Jaime offline. PodSpec was filtered in this PR: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/413

This approach proves to be **unfeasible** even after adding back hostNetwork in podSpec. In csi driver, the msg is sent to the listner in the csi path, that looks like "/var/lib/kubelet/pods/c9a5236b-cf41-49c0-bcfe-f61988440b8f/volumes/kubernetes.io~csi/gcs-fuse-csi-ephemeral/mount". The msg received in sidecar mounter is received in the path of the volume socket: "connecting to socket "/gcsfuse-tmp/.volumes/gcs-fuse-csi-ephemeral/socket". They are not the same msg.

Will stick to the original design.

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: I have decided to stick with the common pattern found in the gcsfuse csi driver repo.

cmd/sidecar_mounter/main.go
hime: @siyanshen Do we know how all the other options are making their way from the node-server to the sidecar? Regarding the different paths being used, I believe we are using a symbolic link for socket communication. Could you clarify this so going forward we follow a different approach then?

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/eeeb147f62994c559bd423221f2cef72bac68658/pkg/csi_mounter/csi_mounter.go#L211-L215

hime:
lgtm

cmd/sidecar_mounter/main.go
siyanshen: Ok I got it. Sending a new PR now.

PR #430, opened 2025-01-16
Increase sidecar memory limit for gcsfuse integration tests on GKE.

hime:
/cc @saikat-royc


saikat-royc:
Please put some background context why we are making this change

PR #429, opened 2025-01-14
fuechr/Add automated go version extraction for e2e tests with fuse 

chrisThePattyEater:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>  
> /kind api-change  

/kind bug  

> /kind cleanup  
> /kind design  
> /kind documentation  
> /kind failing-test  
> /kind feature  
> /kind flake  

**What this PR does / why we need it**:  
Adds additional step during test to parse  '/gcsfuse/tools/cd_scripts/e2e_test.sh' for the version of go being used by fuse teams integration tests and uses that version to run 

This has been causing issues for the team while running fuse teams integrations tests because whenever fuse team updates their version of go and our versions are not in sync, integration tests fail causing us to have to manually upgrade the version in the command that runs those integration test as seen in  https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/commit/11e2f8e7e866257a555aad1b698b58482269d77c. This happens about once a month

**Which issue(s) this PR fixes**:  
Fixes # https://buganizer.corp.google.com/issues/379879910

**Special notes for your reviewer**:
n/a
**Does this PR introduce a user-facing change?**:  
n/a

PR #444, opened 2025-01-25
Corrected go version command to use cut command

chrisThePattyEater:
**What type of PR is this?**
Bug fix

**What this PR does / why we need it:**
This is a correction for https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/429 
This change corrects the command to grep go version it is needed because it is causing test failures

**Which issue(s) this PR fixes:**
Fixes # https://buganizer.corp.google.com/issues/379879910

**Special notes for your reviewer:**
n/a
**Does this PR introduce a user-facing change?:**
n/a


hime:
lgtm

PR #450, opened 2025-01-31
Add HostNetwork Enabled test case

siyanshen:
This test case covered the IO functionality of a hostnetwork enabled pod.

This PR needs to be checked in ONLY AFTER [the hostnetwork PR](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/428) is checked in.

command to test HTTP protocol:
```
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=true E2E_TEST_FOCUS=should.successfully.mount.for.hostnetwork.enabled.pods STAGINGVERSION=v999.999.222 REGISTRY=gcr.io/shensiyan-joonix
```

command to test GRPC protocol

```
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=true E2E_TEST_FOCUS=should.successfully.mount.for.hostnetwork.enabled.pods STAGINGVERSION=v999.999.222 REGISTRY=gcr.io/shensiyan-joonix GCSFUSE_CLIENT_PROTOCOL=grpc
```


test/e2e/testsuites/mount.go
```diff
@@ -111,10 +111,46 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
...
+	}
+
 	ginkgo.It("[read ahead config] should update read ahead config knobs", func() {
 		if pattern.VolType == storageframework.DynamicPV {
 			e2eskipper.Skipf("skip for volume type %v", storageframework.DynamicPV)
 		}
 		testCaseStoreAndRetainData(specs.EnableCustomReadAhead)
 	})
+
+	ginkgo.It("should successfully mount for hostnetwork enabled pods", func() {
```
hime, line 173: do we have a check to prevent running this testsuite against older GKE releases? this testsuite runs at head for both non-managed and managed CSI and runs for all GKE versions that are supported (not EoL)

hime:
lgtm (left nit comment on skipping based on versioning)

one thing to consider though, we have a lot of testsuites, it would be good to run some more variations of those tests with hostnetworking enabled if there's a possibility of regression there.

test/e2e/testsuites/mount.go
siyanshen: Added min version check

PR #455, opened 2025-02-05
add integration test for gcsfuse v2.9.0

tyuchn:
Add an integration test for gcsfuse `v2.9.0`, which includes a new flag, `enable-streaming-writes`. The flag is disabled by default in gcsfuse

### DIFF
--- a/test/e2e/testsuites/gcsfuse_integration.go
+++ b/test/e2e/testsuites/gcsfuse_integration.go
@@ -37,19 +37,20 @@ import (
 const (
 	gcsfuseIntegrationTestsBasePath = "gcsfuse/tools/integration_tests"
 
-	testNameOperations           = "operations"
-	testNameReadonly             = "readonly"
-	testNameRenameDirLimit       = "rename_dir_limit"
-	testNameImplicitDir          = "implicit_dir"
-	testNameExplicitDir          = "explicit_dir"
-	testNameReadLargeFiles       = "read_large_files"
-	testNameWriteLargeFiles      = "write_large_files"
-	testNameGzip                 = "gzip"
-	testNameLocalFile            = "local_file"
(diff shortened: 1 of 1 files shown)


test/e2e/testsuites/gcsfuse_integration.go
```diff
@@ -586,4 +587,19 @@ func (t *gcsFuseCSIGCSFuseIntegrationTestSuite) DefineTests(driver storageframew
...
 		gcsfuseIntegrationTest(testNameManagedFolders+":TestEnableEmptyManagedFoldersTrue", false, "implicit-dirs=true")
 	})
+
+	ginkgo.It(testNamePrefixSucceed+testNameEnableStreamingWrites+testNameSuffix(1), func() {
+		init()
+		defer cleanup()
+
+		v, err := version.ParseSemantic(gcsfuseVersionStr)
+		// If error != nil, this means we've autogenerated a tag (meaning we run from HEAD)
+		// Otherise, we have a valid tag, and we compare against the supported release.
```
hime: Because you are only calling gcsfuseIntegrationTest, you can more easily implement the skip by doing it on the `skipTestOrProceedWithBranch` function. This way, the test will show as skipped instead of pass. Checking the error outside this function is useful if test functionality has been updated/changed over time and we want different behavior for different versions.

Where to add skip: 
https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/test/e2e/testsuites/gcsfuse_integration.go#L129C2-L129C29

hime:
lgtm

test/e2e/testsuites/gcsfuse_integration.go
tyuchn: Done

PR #460, opened 2025-02-07
Use projected token volume for hostNetwork pods 

siyanshen:
## Why we need this change
Before this change, GCSFuse CSI driver does not support pods with hostNetwork:true. This is because the gcsfuse process runs as part of a sidecar container that is injected into the user pod. GCSFuse uses the [ADC](https://cloud.google.com/docs/authentication/application-default-credentials#attached-sa) workflow to fetch the necessary token to access a GCS bucket. However, with hostNetwork enabled, GKE metadata server cannot intercept the token requests for `GET /computeMetadata/v1/instance/service-accounts/default/token` endpoint.

## What is in this change
1.  For `HostNetwork=true` user pods, GCSFuse CSI webhook injects a projected SA token volume to the user pod.
2. `gke-gcsfuse-sidecar` container prepares a unix domain socket and starts a handler to serve requests on this token; invoke gcsfuse with a config option to point to the socket `gcs-auth:token-url:<path to the token>`
3. Token request handler in the `gke-gcsfuse-sidecar` handles token request from GCSFuse.
4. feature flag.

## Local set up and testing
**Feature flag**
Added the flag to gate recent feature: SA Volume injection for hostNetwork support. This feature will not be in effect unless a new GCW component is rolled out with "--should-inject-sa-vol" set as true.

test cases covered:
1. Leave the flag unset
a. hostnetwork=true pod- no access to gcs bucket
b. hostnetwork=false pod - normal access

2. Update `deploy/base/webhook.yaml ` to use arg `- --should-inject-sa-vol=true`
a. hostnetwork=true pod- successful access to bucket
b. hostnetwork=false pod - normal access to bucket

**Test cases covered**
1. pod with hostNetwork=true: can I/O GCS bucket.
2. pod with hostNetwork=false: is not affected and can I/O GCS bucket.
3. pod with hostNetwork=true, with GCS bucket A mounted as ephemeral storage and Gcs bucket B mounted as PV: can access bucket A & B without issue.

**Setup**
1. Create a cluster with workload identity pool enabled & GCSFuse disabled.
2. Grant GCS permissions to an SA
```
kubectl create namespace <ns>
kubectl create serviceaccount <test-ksa-ns> \
    --namespace <ns>

gcloud storage buckets add-iam-policy-binding gs://<bucket> \
    --member "principal://iam.googleapis.com/projects/<project-number>/locations/global/workloadIdentityPools/<project-id>.svc.id.goog/subject/ns/<ns>/sa/<test-ksa-ns>" \
```
5. Build and install your GCSFuse image
```
make build-image-and-push-multi-arch REGISTRY=<your-registry> STAGINGVERSION=<your-dev-tag>
make install REGISTRY=<your-registry> PROJECT=<project-id> STAGINGVERSION=<your-dev-tag>
```
6. Create a user pod with `hostNetwork=true`, GCS bucket <bucket> mounted as a volume, and service account set as the one you created in step 2. Check I/O to your bucket from a container in your pod.


pkg/csi_driver/node.go
```diff
@@ -139,6 +141,10 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		return nil, status.Errorf(codes.NotFound, "failed to get pod: %v", err)
 	}
 
+	if pod.Spec.HostNetwork {
+		fuseMountOptions = joinMountOptions(fuseMountOptions, []string{"start-token-server=" + strconv.FormatBool(s.shouldStartTokenServer(pod))})
```
hime, line 145: You can replace "start-token-server" with `tokenServerFlag`

hime:
lgtm

pkg/csi_driver/node.go
siyanshen: I still prefer start-token-server, just because it's more intuitive than token-server-flag.

PR #472, opened 2025-02-12
Added capability to provide metadata prefetch memory request and limits in pod spec

chrisThePattyEater:
Added capability to provide metadata prefetch memory request and limits in pod spec to fuse csi annotations Additionally added unit tests verifying changes.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature

> /kind flake

This pr adds functionality to allow fuse driver users to provide custom metadata prefetch sidecar memory and limits. The changes add additional annotations and variables to extract prefetch limits and requests if provided and uses those values for the prefetch sidecar config if provided. Otherwise it uses default values

Fixes https://b.corp.google.com/issues/385198703

```release-note Annotations to provide metadataprefetch mem limit and requests are : gke-gcsfuse/metadata-prefetch/memory-request and gke-gcsfuse/metadata-prefetch/memory-limit

```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -39,21 +39,27 @@ import (
 )
 
 var (
-	port                    = flag.Int("port", 443, "The port that the webhook server serves at.")
-	healthProbeBindAddress  = flag.String("health-probe-bind-address", ":8080", "The TCP address that the controller should bind to for serving health probes.")
-	certDir                 = flag.String("cert-dir", "/etc/tls-certs", "The directory that contains the server key and certificate.")
-	certName                = flag.String("cert-name", "cert.pem", "The server certificate name.")
-	keyName                 = flag.String("key-name", "key.pem", "The server key name.")
-	imagePullPolicy         = flag.String("sidecar-image-pull-policy", "IfNotPresent", "The default image pull policy for gcsfuse sidecar container.")
-	cpuRequest              = flag.String("sidecar-cpu-request", "250m", "The default CPU request for gcsfuse sidecar container.")
-	cpuLimit                = flag.String("sidecar-cpu-limit", "250m", "The default CPU limit for gcsfuse sidecar container.")
-	memoryRequest           = flag.String("sidecar-memory-request", "256Mi", "The default memory request for gcsfuse sidecar container.")
-	memoryLimit             = flag.String("sidecar-memory-limit", "256Mi", "The default memory limit for gcsfuse sidecar container.")
-	ephemeralStorageRequest = flag.String("sidecar-ephemeral-storage-request", "5Gi", "The default ephemeral storage request for gcsfuse sidecar container.")
-	ephemeralStorageLimit   = flag.String("sidecar-ephemeral-storage-limit", "5Gi", "The default ephemeral storage limit for gcsfuse sidecar container.")
-	sidecarImage            = flag.String("sidecar-image", "", "The gcsfuse sidecar container image.")
-	metadataSidecarImage    = flag.String("metadata-sidecar-image", "", "The metadata prefetch sidecar container image.")
-	injectSAVol             = flag.Bool("should-inject-sa-vol", false, "Inject projected service account volume when true")
+	port                                    = flag.Int("port", 443, "The port that the webhook server serves at.")
+	healthProbeBindAddress                  = flag.String("health-probe-bind-address", ":8080", "The TCP address that the controller should bind to for serving health probes.")
+	certDir                                 = flag.String("cert-dir", "/etc/tls-certs", "The directory that contains the server key and certificate.")
+	certName                                = flag.String("cert-name", "cert.pem", "The server certificate name.")
+	keyName                                 = flag.String("key-name", "key.pem", "The server key name.")
+	imagePullPolicy                         = flag.String("sidecar-image-pull-policy", "IfNotPresent", "The default image pull policy for gcsfuse sidecar container.")
+	cpuRequest                              = flag.String("sidecar-cpu-request", "250m", "The default CPU request for gcsfuse sidecar container.")
+	cpuLimit                                = flag.String("sidecar-cpu-limit", "250m", "The default CPU limit for gcsfuse sidecar container.")
+	memoryRequest                           = flag.String("sidecar-memory-request", "256Mi", "The default memory request for gcsfuse sidecar container.")
+	memoryLimit                             = flag.String("sidecar-memory-limit", "256Mi", "The default memory limit for gcsfuse sidecar container.")
+	ephemeralStorageRequest                 = flag.String("sidecar-ephemeral-storage-request", "5Gi", "The default ephemeral storage request for gcsfuse sidecar container.")
+	ephemeralStorageLimit                   = flag.String("sidecar-ephemeral-storage-limit", "5Gi", "The default ephemeral storage limit for gcsfuse sidecar container.")
+	sidecarImage                            = flag.String("sidecar-image", "", "The gcsfuse sidecar container image.")
+	metadataSidecarImage                    = flag.String("metadata-sidecar-image", "", "The metadata prefetch sidecar container image.")
+	injectSAVol                             = flag.Bool("should-inject-sa-vol", false, "Inject projected service account volume when true")
+	metadataMemoryRequest                   = flag.String("metadata-sidecar-memory-request", "10Mi", "Flag to use default value for gcsfuse memory prefetch sidecar container memory request.")
+	metadataMemoryLimit                     = flag.String("metadata-sidecar-memory-limit", "10Mi", "Flag to use default value for gcsfuse memory prefetch sidecar container memory limit.")
+	metadataPrefetchCPURequest              = flag.String("metadata-sidecar-cpu-request", "10m", "The default cpu request for gcsfuse memory prefetch sidecar container cpu request.")
+	metadataPrefetchCPULimit                = flag.String("metadata-sidecar-cpu-request", "50m", "Flag to use default value for gcsfuse memory prefetch sidecar container cpu limit.")
+	metadataPrefetchEphemeralStorageRequest = flag.String("metadata-sidecar-ephemeral-storage-request", "10Mi", "The default value for gcsfuse memory prefetch sidecar ephemeral storage request.")
+	metadataPrefetchEphemeralStorageLimit   = flag.String("metadata-sidecar-ephemeral-storage-request", "10Mi", "The default value for gcsfuse memory prefetch sidecar ephemeral storage limit.")
 	// These are set at compile time.
 	webhookVersion = "unknown"
 )
@@ -73,9 +79,11 @@ func main() {
 	klog.Infof("Running Google Cloud Storage FUSE CSI driver admission webhook version %v, sidecar container image %v", webhookVersion, *sidecarImage)
 
 	// Load webhook config
-	c := wh.LoadConfig(*sidecarImage, *metadataSidecarImage, *imagePullPolicy, *cpuRequest, *cpuLimit, *memoryRequest, *memoryLimit, *ephemeralStorageRequest, *ephemeralStorageLimit)
-	c.ShouldInjectSAVolume = *injectSAVol
-	klog.Infof("Webhook should inject SA volume: %t", c.ShouldInjectSAVolume)
+	fuseSideCarConfig := wh.LoadConfig(*sidecarImage, *imagePullPolicy, *cpuRequest, *cpuLimit, *memoryRequest, *memoryLimit, *ephemeralStorageRequest, *ephemeralStorageLimit)
+	fuseSideCarConfig.ShouldInjectSAVolume = *injectSAVol
+	klog.Infof("Webhook should inject SA volume: %t", fuseSideCarConfig.ShouldInjectSAVolume)
+
+	metadataPrefetchSideCarConfig := wh.LoadConfig(*metadataSidecarImage, *imagePullPolicy, *metadataPrefetchCPURequest, *metadataPrefetchCPULimit, *metadataMemoryRequest, *metadataMemoryLimit, *metadataPrefetchEphemeralStorageRequest, *metadataPrefetchEphemeralStorageLimit)
 
 	// Load config for manager, informers, listers
 	kubeConfig := config.GetConfigOrDie()
@@ -141,13 +149,14 @@ func main() {
 	klog.Info("Registering webhooks to the webhook server.")
 	hookServer.Register("/inject", &webhook.Admission{
 		Handler: &wh.SidecarInjector{
-			Client:        mgr.GetClient(),
-			Config:        c,
-			Decoder:       admission.NewDecoder(runtime.NewScheme()),
-			NodeLister:    nodeLister,
-			PvLister:      pvLister,
-			PvcLister:     pvcLister,
-			ServerVersion: serverVersion,
+			Client:                 mgr.GetClient(),
+			Config:                 fuseSideCarConfig,
+			MetadataPrefetchConfig: metadataPrefetchSideCarConfig,
+			Decoder:                admission.NewDecoder(runtime.NewScheme()),
+			NodeLister:             nodeLister,
+			PvLister:               pvLister,
+			PvcLister:              pvcLister,
+			ServerVersion:          serverVersion,
 		},
 	})
 
--- a/pkg/webhook/config.go
+++ b/pkg/webhook/config.go
@@ -20,6 +20,7 @@ package webhook
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -27,29 +28,27 @@ import (
 )
 
 type Config struct {
-	ShouldInjectSAVolume   bool   `json:"-"`
-	PodHostNetworkSetting  bool   `json:"-"`
-	ContainerImage         string `json:"-"`
-	MetadataContainerImage string `json:"-"`
-	ImagePullPolicy        string `json:"-"`
+	ShouldInjectSAVolume  bool   `json:"-"`
+	PodHostNetworkSetting bool   `json:"-"`
+	ContainerImage        string `json:"-"`
+	ImagePullPolicy       string `json:"-"`
 	//nolint:tagliatelle
-	CPURequest resource.Quantity `json:"gke-gcsfuse/cpu-request,omitempty"`
+	CPURequest resource.Quantity `json:"cpu-request,omitempty"`
 	//nolint:tagliatelle
-	CPULimit resource.Quantity `json:"gke-gcsfuse/cpu-limit,omitempty"`
+	CPULimit resource.Quantity `json:"cpu-limit,omitempty"`
 	//nolint:tagliatelle
-	MemoryRequest resource.Quantity `json:"gke-gcsfuse/memory-request,omitempty"`
+	MemoryRequest resource.Quantity `json:"memory-request,omitempty"`
 	//nolint:tagliatelle
-	MemoryLimit resource.Quantity `json:"gke-gcsfuse/memory-limit,omitempty"`
+	MemoryLimit resource.Quantity `json:"memory-limit,omitempty"`
 	//nolint:tagliatelle
-	EphemeralStorageRequest resource.Quantity `json:"gke-gcsfuse/ephemeral-storage-request,omitempty"`
+	EphemeralStorageRequest resource.Quantity `json:"ephemeral-storage-request,omitempty"`
 	//nolint:tagliatelle
-	EphemeralStorageLimit resource.Quantity `json:"gke-gcsfuse/ephemeral-storage-limit,omitempty"`
+	EphemeralStorageLimit resource.Quantity `json:"ephemeral-storage-limit,omitempty"`
 }
 
-func LoadConfig(containerImage, metadataContainerImage, imagePullPolicy, cpuRequest, cpuLimit, memoryRequest, memoryLimit, ephemeralStorageRequest, ephemeralStorageLimit string) *Config {
+func LoadConfig(containerImage, imagePullPolicy, cpuRequest, cpuLimit, memoryRequest, memoryLimit, ephemeralStorageRequest, ephemeralStorageLimit string) *Config {
 	return &Config{
 		ContainerImage:          containerImage,
-		MetadataContainerImage:  metadataContainerImage,
 		ImagePullPolicy:         imagePullPolicy,
 		CPURequest:              resource.MustParse(cpuRequest),
 		CPULimit:                resource.MustParse(cpuLimit),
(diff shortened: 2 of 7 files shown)


pkg/webhook/sidecar_spec.go
```diff
@@ -175,14 +175,25 @@ func getMetadataPrefetchConfig(image string) *Config {
 	}
 }
 
-func (si *SidecarInjector) GetMetadataPrefetchSidecarContainerSpec(pod *corev1.Pod, image string) corev1.Container {
+func getMetadataPrefetchConfig(c *Config) *Config {
+	config := getDefaultMetadataPrefetchConfig(c.MetadataContainerImage)
+	if !c.MetadataPrefetchMemoryLimit.IsZero() {
```
hime: @chrisThePattyEater what happens here if we want to set memory limit to zero? We want to support zero when setting [burstable pods](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#burstable) with no limit.

cmd/webhook/main.go
```diff
@@ -54,6 +54,8 @@ var (
 	sidecarImage            = flag.String("sidecar-image", "", "The gcsfuse sidecar container image.")
 	metadataSidecarImage    = flag.String("metadata-sidecar-image", "", "The metadata prefetch sidecar container image.")
 	injectSAVol             = flag.Bool("should-inject-sa-vol", false, "Inject projected service account volume when true")
+	metadataMemoryRequest   = flag.String("metadata-sidecar-memory-request", "-1", "Flag to use default value for gcsfuse memory prefetch sidecar container request.")
```
hime: Can we set the default values here? This is where we define the default values for the other resources.

hime: Right now its set to a default of `-1`.

pkg/webhook/config.go
```diff
@@ -44,27 +44,33 @@ type Config struct {
...
+		MetadataPrefetchMemoryLimit:   resource.MustParse(metadataPrefetchMemoryLimit),
 	}
 }
 
 func FakeConfig() *Config {
 	fakeImage1 := "fake-repo/fake-sidecar-image:v999.999.999-gke.0@sha256:c9cd4cde857ab8052f416609184e2900c0004838231ebf1c3817baa37f21d847"
 	fakeImage2 := "fake-repo/fake-sidecar-image:v888.888.888-gke.0@sha256:c9cd4cde857ab8052f416609184e2900c0004838231ebf1c3817baa37f21d847"
 
-	return LoadConfig(fakeImage1, fakeImage2, "Always", "250m", "250m", "256Mi", "256Mi", "5Gi", "5Gi")
+	return LoadConfig(fakeImage1, fakeImage2, "Always", "250m", "250m", "256Mi", "256Mi", "5Gi", "5Gi", "-1", "-1")
```
hime: Try to use relevant values when testing your changes. If the resource is memory, then use memory units.

pkg/webhook/mutatingwebhook.go
```diff
@@ -34,14 +34,16 @@ import (
...
-	ephemeralStorageRequestAnnotation    = "gke-gcsfuse/ephemeral-storage-request"
+	GcsFuseVolumeEnableAnnotation           = "gke-gcsfuse/volumes"
+	GcsFuseNativeSidecarEnableAnnotation    = "gke-gcsfuse/enable-native-sidecar"
+	cpuLimitAnnotation                      = "gke-gcsfuse/cpu-limit"
+	cpuRequestAnnotation                    = "gke-gcsfuse/cpu-request"
+	memoryLimitAnnotation                   = "gke-gcsfuse/memory-limit"
+	memoryRequestAnnotation                 = "gke-gcsfuse/memory-request"
+	ephemeralStorageLimitAnnotation         = "gke-gcsfuse/ephemeral-storage-limit"
+	ephemeralStorageRequestAnnotation       = "gke-gcsfuse/ephemeral-storage-request"
+	metadataPrefetchMemoryLimitAnnotation   = "gke-gcsfuse/metadata-prefetch/memory-limit"
```
hime, line 45: fyi we might need confirmation on best name to give here from others.

pkg/webhook/sidecar_spec.go
```diff
@@ -155,15 +155,15 @@ func GetSecurityContext() *corev1.SecurityContext {
...
 	container.RestartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
 
 	return container
 }
 
-func getMetadataPrefetchConfig(image string) *Config {
+func getDefaultMetadataPrefetchConfig(image string) *Config {
 	return &Config{
 		CPURequest:              metadataPrefetchCPURequest,
 		CPULimit:                metadataPrefetchCPULimit,
```
hime: I think here you want to set the memory of the metadata prefetch fields, not override the sidecar fields. I actually think that it makes more sense right now, where we use one config struct per sidecar. I think we want to simplify the code as much as possible to increase readability instead of adding more fields.

hime: I think we should add more broad unit tests that cover end-to-end behavior where we explicitly set the annotations, and we verify these annotations are applied and don't overlap.

pkg/webhook/mutatingwebhook_test.go
```diff
@@ -226,6 +226,87 @@ func TestPrepareConfig(t *testing.T) {
 			},
 			expectErr: false,
 		},
+		{
```
hime, line 230: can we add a test that doesn't provide limits and verify we use defaults?

chrisThePattyEater: isn't that every other test that doesn't provide limits

hime: You're right. In that case the current tests should be failing because the `wantConfig` should contain the default values.

pkg/webhook/injection.go
```diff
@@ -24,8 +24,18 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 )
 
+var sidecarPrefixMap = map[string]string{
+	GcsFuseSidecarName:          "gke-gcsfuse",
+	MetadataPrefetchSidecarName: "gke-gcsfuse-metadata-prefetch",
+}
+var containerIndexOrderMap = map[string]string{
```
hime, line 36: Interesting approach! Could you add a comment that adds more context for the map? Thanks!

pkg/webhook/injection.go
```diff
@@ -24,8 +24,18 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 )
 
+var sidecarPrefixMap = map[string]string{
+	GcsFuseSidecarName:          "gke-gcsfuse",
+	MetadataPrefetchSidecarName: "gke-gcsfuse-metadata-prefetch",
```
hime: Can we keep the `gke-gcsfuse/` prefix for both? I think customers are most familiar with this one. We could make the prefix for metadata prefetch look like so: `gke-gcsfuse/metadata-prefetch-memory-limit`

cmd/webhook/main.go
```diff
@@ -73,9 +79,11 @@ func main() {
...
 
 	// Load webhook config
-	c := wh.LoadConfig(*sidecarImage, *metadataSidecarImage, *imagePullPolicy, *cpuRequest, *cpuLimit, *memoryRequest, *memoryLimit, *ephemeralStorageRequest, *ephemeralStorageLimit)
-	c.ShouldInjectSAVolume = *injectSAVol
-	klog.Infof("Webhook should inject SA volume: %t", c.ShouldInjectSAVolume)
+	fuseSideCarConfig := wh.LoadConfig(*sidecarImage, *imagePullPolicy, *cpuRequest, *cpuLimit, *memoryRequest, *memoryLimit, *ephemeralStorageRequest, *ephemeralStorageLimit)
+	fuseSideCarConfig.ShouldInjectSAVolume = *injectSAVol
+	klog.Infof("Webhook should inject SA volume: %t", fuseSideCarConfig.ShouldInjectSAVolume)
+
+	metaDataPrefetchSideCarConfig := wh.LoadConfig(*metadataSidecarImage, *imagePullPolicy, *metadataPrefetchCPURequest, *metadataPrefetchCPULimit, *metadataMemoryRequest, *metadataMemoryLimit, *metadataPrefetchEphemeralStorageRequest, *metadataPrefetchEphemeralStorageLimit)
```
hime: Just for consistency with the rest of the codebase, you can rename `metaDataPrefetchSideCarConfig` to `metadataPrefetchSidecarConfig`.

pkg/webhook/config.go
```diff
@@ -59,12 +57,45 @@ func LoadConfig(containerImage, metadataContainerImage, imagePullPolicy, cpuRequ
 		EphemeralStorageLimit:   resource.MustParse(ephemeralStorageLimit),
 	}
 }
+func generateAnnotationsFromConfig(config *Config, prefix string) map[string]string {
+	var annotations = make(map[string]string)
+	if config.ImagePullPolicy != "" {
+		annotations[prefix+"/image-pull-policy"] = config.ImagePullPolicy
```
hime: to match the json config, you can say `annotations[prefix+"image-pull-policy"]` where prefix is `gke-gcsfuse/`. this comment also applies for all the references of annotations below.

pkg/webhook/config.go
```diff
@@ -59,12 +57,45 @@ func LoadConfig(containerImage, metadataContainerImage, imagePullPolicy, cpuRequ
...
 func FakeConfig() *Config {
 	fakeImage1 := "fake-repo/fake-sidecar-image:v999.999.999-gke.0@sha256:c9cd4cde857ab8052f416609184e2900c0004838231ebf1c3817baa37f21d847"
-	fakeImage2 := "fake-repo/fake-sidecar-image:v888.888.888-gke.0@sha256:c9cd4cde857ab8052f416609184e2900c0004838231ebf1c3817baa37f21d847"
-
-	return LoadConfig(fakeImage1, fakeImage2, "Always", "250m", "250m", "256Mi", "256Mi", "5Gi", "5Gi")
+	
+	return LoadConfig(fakeImage1, "Always", "250m", "250m", "256Mi", "256Mi", "5Gi", "5Gi")
+}
+func FakePrefetchConfig() *Config {
+	fakeImage2 := "fake-image"
```
hime: you can also hardcode the string into the parameter since its such a short name (instead of having local variable)

pkg/webhook/config.go
```diff
@@ -115,28 +146,44 @@ func populateResource(requestQuantity, limitQuantity *resource.Quantity, default
 	}
 }
 
-// prepareConfig overwrittes config values set by user input from pod annotations,
+// prepareConfig overwrites config values set by user input from pod annotations,
 // remaining values that are not specified by user are kept as the default config values.
-func (si *SidecarInjector) prepareConfig(annotations map[string]string) (*Config, error) {
+func (si *SidecarInjector) prepareConfig(annotations map[string]string, prefix string) (*Config, error) {
+	var defaultConfig *Config
```
hime: This is a bit hard to follow, would it be easier to make a new field in the Config struct called `annotationPrefix`? Then we can pass the defaultConfig as a parameter.

pkg/webhook/config.go
```diff
@@ -115,28 +146,44 @@ func populateResource(requestQuantity, limitQuantity *resource.Quantity, default
...
+		ShouldInjectSAVolume: defaultConfig.ShouldInjectSAVolume,
+		ContainerImage:       defaultConfig.ContainerImage,
+		ImagePullPolicy:      defaultConfig.ImagePullPolicy,
 	}
 
-	jsonData, err := json.Marshal(annotations)
+	extractedData := make(map[string]string)
+	for key, value := range annotations {
+		// Check if the key starts with the given prefix
+		if strings.HasPrefix(key, prefix+"/") {
```
hime: You can add the `/` to the prefix string to simplify code.

pkg/webhook/config.go
```diff
@@ -115,28 +146,44 @@ func populateResource(requestQuantity, limitQuantity *resource.Quantity, default
...
+			// Remove the prefix and keep only the case name
+			newKey := strings.TrimPrefix(key, prefix+"/")
+			extractedData[newKey] = value
+		}
+	}
+	extractedJSON, err := json.Marshal(extractedData)
 	if err != nil {
-		return nil, fmt.Errorf("failed to marshal pod annotations: %w", err)
+		return nil, fmt.Errorf("failed to parse sidecar container resource allocation from pod annotations: %w", err)
```
hime: since we now have 2 sidecar that we do this for, it would be good to mention in the log what annotation or sidecar caused this failure.

pkg/webhook/injection.go
hime: In here, I think we want to also explain why does this map exist? Why does the ordering exist.

pkg/webhook/injection.go
```diff
@@ -24,13 +24,26 @@ import (
...
+	GcsFuseSidecarName:          IstioSidecarName,
+	MetadataPrefetchSidecarName: GcsFuseSidecarName,
+}
+
 const IstioSidecarName = "istio-proxy"
 
 func (si *SidecarInjector) injectAsNativeSidecar(pod *corev1.Pod) (bool, error) {
 	supportsNativeSidecar, err := si.supportsNativeSidecar()
 	if err != nil {
+
```
hime: random new line, maybe go through the code and remove these new lines unless they are needed.

pkg/webhook/injection.go
```diff
@@ -86,63 +101,90 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
-func injectSidecarContainer(pod *corev1.Pod, config *Config, injectAsNativeSidecar bool) {
-	if injectAsNativeSidecar {
-		pod.Spec.InitContainers = insert(pod.Spec.InitContainers, GetNativeSidecarContainerSpec(config), getInjectIndex(pod.Spec.InitContainers))
-	} else {
-		pod.Spec.Containers = insert(pod.Spec.Containers, GetSidecarContainerSpec(config), getInjectIndex(pod.Spec.Containers))
-	}
-}
-
-func (si *SidecarInjector) injectMetadataPrefetchSidecarContainer(pod *corev1.Pod, config *Config, injectAsNativeSidecar bool) {
+func (si *SidecarInjector) injectSidecarContainer(containerName string, pod *corev1.Pod, injectAsNativeSidecar bool) (bool, error) {
```
hime: This function no longer just performs `injectSidecarContainer`. It now also performs validation steps , prepares configs, etc. can we keep this function true to its name? Sometimes we can oversimplify the code, which makes it harder to read. 

I'd say lets keep this function as is (unchanged from original) for this PR. If you want to refactor this part of the code, you can create a new PR that focuses on that. We want to keep our PRs focused on our goal (which in this case is adding metadata custom resource support). Its good to keep your changes self contained.

Might be good to refactor the code, but I'd create a separate PR & issue to track that.

hime: This comment applies for the whole `injection.go` file.

pkg/webhook/mutatingwebhook.go
hime: gave suggestions on flag variable names.

pkg/webhook/mutatingwebhook.go
```diff
@@ -34,88 +34,55 @@ import (
...
-	if err != nil {
-		return admission.Errored(http.StatusBadRequest, err)
-	}
-
-	config.PodHostNetworkSetting = pod.Spec.HostNetwork
-
-	if userProvidedGcsFuseSidecarImage, err := ExtractImageAndDeleteContainer(&pod.Spec, GcsFuseSidecarName); err == nil {
-		if userProvidedGcsFuseSidecarImage != "" {
-			config.ContainerImage = userProvidedGcsFuseSidecarImage
+	// Validate injection request
```
hime: Same suggestion that I left on `injection.go`. Lets keep this PR self contained. Refactoring can go on another PR.

pkg/webhook/config.go
chrisThePattyEater: How do you do this

pkg/webhook/config.go
chrisThePattyEater: but this is not constant its either gke-gcsfuse or its metadata

pkg/webhook/config.go
chrisThePattyEater: i don't think this makes sense as its not part of the configuration of a pod its and interal path lever, do you really want this change?

pkg/webhook/config.go
```diff
@@ -59,12 +57,43 @@ func LoadConfig(containerImage, metadataContainerImage, imagePullPolicy, cpuRequ
 		EphemeralStorageLimit:   resource.MustParse(ephemeralStorageLimit),
 	}
 }
+func generateAnnotationsFromConfig(config *Config, prefix string) map[string]string {
```
hime: is this only used for testing? if so lets move to config_test.go

pkg/webhook/config.go
hime: Just an idea, we can think of more ideas here. I think we should find a way to simplify a bit.

pkg/webhook/injection.go
```diff
@@ -53,11 +65,13 @@ func (si *SidecarInjector) injectAsNativeSidecar(pod *corev1.Pod) (bool, error)
 
 func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
 	if si.ServerVersion != nil && !si.ServerVersion.AtLeast(minimumSupportedVersion) {
+
```
hime: Can you remove blank lines throughout PR. Thanks!

pkg/webhook/injection.go
hime: Lets discuss a good description here in 1:1

pkg/webhook/injection.go
```diff
@@ -86,63 +100,72 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+	containerSpec.Env = append(containerSpec.Env, corev1.EnvVar{Name: "NATIVE_SIDECAR", Value: "TRUE"})
+	containerSpec.RestartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
+
+	return containerSpec
+}
+func (si *SidecarInjector) getContainerSpec(containerName string, pod *corev1.Pod, config *Config) corev1.Container {
+	if containerName == MetadataPrefetchSidecarName {
+		return si.GetMetadataPrefetchSidecarContainerSpec(pod, config)
+	}
+	return GetSidecarContainerSpec(config)
```
hime, line 167: Could you fix verify each time you request PR review. Sometimes these checks catch bugs.

pkg/webhook/config.go
```diff
@@ -115,28 +117,44 @@ func populateResource(requestQuantity, limitQuantity *resource.Quantity, default
...
+	}
 	config := &Config{
-		ShouldInjectSAVolume:   si.Config.ShouldInjectSAVolume,
-		ContainerImage:         si.Config.ContainerImage,
-		MetadataContainerImage: si.Config.MetadataContainerImage,
-		ImagePullPolicy:        si.Config.ImagePullPolicy,
+		ShouldInjectSAVolume: defaultConfig.ShouldInjectSAVolume,
+		ContainerImage:       defaultConfig.ContainerImage,
+		ImagePullPolicy:      defaultConfig.ImagePullPolicy,
 	}
```
hime, line 137: extractedData can be its own function

pkg/webhook/config.go
```diff
@@ -115,28 +117,44 @@ func populateResource(requestQuantity, limitQuantity *resource.Quantity, default
...
 
-	jsonData, err := json.Marshal(annotations)
+	extractedData := make(map[string]string)
```
hime, line 146: can we have extractedData be its own function

pkg/webhook/sidecar_spec.go
```diff
@@ -155,41 +155,40 @@ func GetSecurityContext() *corev1.SecurityContext {
 	}
 }
 
-func (si *SidecarInjector) GetNativeMetadataPrefetchSidecarContainerSpec(pod *corev1.Pod, image string) corev1.Container {
-	container := si.GetMetadataPrefetchSidecarContainerSpec(pod, image)
+func (si *SidecarInjector) GetNativeMetadataPrefetchSidecarContainerSpec(pod *corev1.Pod, config *Config) corev1.Container {
```
hime: Where do we use this function?

pkg/webhook/sidecar_spec.go
```diff
@@ -155,41 +155,40 @@ func GetSecurityContext() *corev1.SecurityContext {
...
+	container := si.GetMetadataPrefetchSidecarContainerSpec(pod, config)
 	container.Env = append(container.Env, corev1.EnvVar{Name: "NATIVE_SIDECAR", Value: "TRUE"})
 	container.RestartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
 
 	return container
 }
 
-func getMetadataPrefetchConfig(image string) *Config {
+func getDefaultMetadataPrefetchConfig(image string) *Config {
```
hime: are we reading the flags or the hardcoded variables? I remember we had hardcoded variables names `metadataPrefetchCPURequest` etc.

pkg/webhook/sidecar_spec.go
chrisThePattyEater: removed this, you are right we are not using anymore, its been replaced with a common one func (si *SidecarInjector) getNativeContainerSpec(containerName string, pod *corev1.Pod, config *Config) corev1.Container {

pkg/webhook/sidecar_spec.go
chrisThePattyEater: this is only for tests
removed

pkg/webhook/config.go
chrisThePattyEater: yes

pkg/webhook/injection.go
chrisThePattyEater: yes sir

pkg/webhook/config.go
chrisThePattyEater: ok

pkg/webhook/config.go
hime: okay, lets do that and resolve this thread.

pkg/webhook/injection.go
hime: feel free to resolve once meeting ideas for desc are applied here.

pkg/webhook/config.go
chrisThePattyEater: moved everything into separate methods

pkg/webhook/injection.go
```diff
@@ -86,63 +98,72 @@ func (si *SidecarInjector) supportsNativeSidecar() (bool, error) {
...
+		index = getInjectIndexAfterContainer(pod.Spec.Containers, containerIndexOrderMap[containerName])
 	}
 
-	if len(containerSpec.VolumeMounts) == 0 {
+	// Skip metadata prefetch sidecar injection if no volumes are requesting metadata prefetch.
+	if containerName == MetadataPrefetchSidecarName && len(containerSpec.VolumeMounts) == 0 {
 		klog.Info("no volumes are requesting metadata prefetch, skipping metadata prefetch sidecar injection")
 
-		return
+		return true, nil
```
hime: why are we returning true here if we didn't inject? can you add a description to the function that explains what it does?

e.g.
// injectSidecarContainer performs xyz
// returns true/false when sidecar is injected ...

pkg/webhook/injection_test.go
```diff
@@ -25,14 +25,38 @@ import (
...
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/utils/ptr"
 )
 
-var UnsupportedVersion = version.MustParseGeneric("1.28.0")
+var (
+	UnsupportedVersion = version.MustParseGeneric("1.28.0")
+
+	// Metadata prefetch container resources.
+	metadataPrefetchCPURequest              = resource.MustParse("10m")
```
hime: these variables are named the same as the flag names. can you change the name to avoid confusion? You can also just define the values directly in the `getDefaultMetadataPrefetchConfig` function and get rid of the variables.

pkg/webhook/mutatingwebhook_test.go
```diff
@@ -48,18 +48,19 @@ func TestPrepareConfig(t *testing.T) {
...
 	testCases := []struct {
 		name        string
+		prefix      string
 		annotations map[string]string
 		wantConfig  *Config
 		expectErr   bool
 	}{
 		{
-			name: "use default values if no annotation is found",
+			name:   "use default values if no annotation is found",
```
hime, line 57: can you confirm if we have test coverage in cases where customer wants to customize resource request and limits for both sidecars? would be good to have if we don't

hime (approved):
lgtm

PR #499, opened 2025-02-28
Add map lock on Publish/Unpublish volume

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Add lock on volumeStateStore map during Publish/Unpublish volume. This allows for concurrent Publish/Unpublish operations.

Fixes #394212208

```release-note
NONE
```


pkg/csi_driver/node.go
```diff
@@ -309,3 +309,30 @@ func (s *nodeServer) shouldStartTokenServer(pod *corev1.Pod) bool {
...
+	timeout := time.After(5 * time.Second) // Timeout after 5 seconds
+	for {
+		select {
+		case <-timeout:
+			// If timeout occurs, log and return
+			klog.Infof("Timeout while trying to acquire lock for resource: %s", resource)
+
+			return
+		default:
+			if acquired := s.volumeLocks.TryAcquire(resource); acquired {
```
pwschuurman: Please create a mutex specifically for what you're tracking to lock/unlock. `volumeLocks` is intended to prevent concurrent operations per PVC (`volumeID`). The resource you're protecting (s.volumeStateStore) needs global locking based on the scope of the nodeServer object.

With the way you're calling `volumeLocks`, the mutex lock is only acquired during the `TryAcquire` or `Release` calls. You could still have two concurrent threads writing or reading from `s.volumeStateStore` with different volumeIDs as the same time. This would violate golang map's concurrency guarantees: https://go.dev/doc/go1.6#runtime

```
The runtime has added lightweight, best-effort detection of concurrent misuse of maps. As always, if one goroutine is writing to a map, no other goroutine should be reading or writing the map concurrently. If the runtime detects this condition, it prints a diagnosis and crashes the program.
```

Also, I don't think you need a deadlock avoidance timeout. As long as the operation logic you're adding (and it should be explicit, not passing in a generic function, as it becomes much harder to spot deadlocks) deterministically completes, you can block on lock acquisition.

IMO you should encapsulate volumeStateStore map in its own struct, with thread safe get/set methods. As an alternative, you can use an existing thread safe map datastructure, that already has these functions built-in.

pwschuurman (approved):

PR #503, opened 2025-03-05
Increase buffer size in Recvmsg to allow for longer mountOptions

siyanshen:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

We want to support longer mount options up to 5000 bytes

Fixes #

```release-note
Increase mountOptions size allowance
```


pkg/util/fdchannel.go
```diff
@@ -59,7 +59,7 @@ func RecvMsg(via net.Conn) (int, []byte, error) {
 
 	klog.V(4).Info("calling recvmsg...")
 	buf := make([]byte, syscall.CmsgSpace(4))
-	b := make([]byte, 500)
+	b := make([]byte, 5000)
```
amacaskill, line 62: How do we know 5000 will handle most of the mountOption cases? Is there a way to make this a resizable buffer like Vikas suggested in our eng sync on March 4?

siyanshen: Added a helper function for that.

pkg/util/fdchannel.go
```diff
@@ -81,3 +80,32 @@ func RecvMsg(via net.Conn) (int, []byte, error) {
...
+		var nn int
+		nn, _, _, _, err := syscall.Recvmsg(fd, buffer[totalReceived:], oob, flags)
+		if err != nil {
+			return 0, nil, err
+		}
+
+		totalReceived += nn
+		klog.Infof("total bytes of received msg: %d", totalReceived)
+
+		if nn < len(buffer)-totalReceived {
```
amacaskill: buffer := make([]byte, 500)
len(buffer) will always be 500 because you used make, even with no data inside. https://go.dev/tour/moretypes/13.

pkg/util/fdchannel.go
```diff
@@ -59,9 +59,8 @@ func RecvMsg(via net.Conn) (int, []byte, error) {
 
 	klog.V(4).Info("calling recvmsg...")
 	buf := make([]byte, syscall.CmsgSpace(4))
-	b := make([]byte, 500)
-	//nolint:dogsled
-	n, _, _, _, err := syscall.Recvmsg(socket, b, buf, 0)
+
+	n, b, err := ReceiveCompleteMsg(socket, buf, 0, 5000)
```
amacaskill: Thinking more about this, if the max is still the same 5000, I think it might be better to just set this out write, rather than increasing by x2 each time. Its not that much memory, and I don't think it is a big deal. I think this suggestion by Vikas was good if we weren't going to set a max, but since we set a max, I think lets just do that for simplicity sake. 

 Starting at 500 then increasing *2, and creating new buffers  and copying data over, takes more memory (500+1000+2000+4000) vs just creating 1 5k buffer, assuming it fills up all the way. Short term, maybe this would save a little memory, but I don't think its enough to really make a difference.

pkg/util/fdchannel.go
```diff
@@ -81,3 +80,32 @@ func RecvMsg(via net.Conn) (int, []byte, error) {
...
 	return fds[0], b[:n], err
 }
+
+func ReceiveCompleteMsg(fd int, oob []byte, flags int, bufferLimit int) (int, []byte, error) {
+	buffer := make([]byte, 500)
+	totalReceived := 0
+
+	for {
+		var nn int
+		nn, _, _, _, err := syscall.Recvmsg(fd, buffer[totalReceived:], oob, flags)
```
amacaskill: nn is redeclared here.

pkg/util/fdchannel.go
siyanshen: SG. Changed to the first draft approach.

pkg/util/fdchannel.go
siyanshen: Resolving this as we are going back to the original approach

PR #504, opened 2025-03-06
Update go.mod to pick up k8s 1.32 changes, fixing several CVEs.

mattcary:
Our auto-update tools can't deal with the k8s.io/kubernetes includes. To fix this, I cleared out go.mod (leaving only the package and go version), and deleted vendor/, and then ran

github.com/kubernetes-csi/csi-release-tools/go-get-kubernetes.sh -p 1.32.2 go mod tidy
run the go get commands suggested from the above command
re-run it
go mod vendor

/kind cleanup

```release-note
Update k8s.io/kubernetes to 1.32.2 to fix several CVEs.
```


tyuchn:
Are we intending to merge the PR into `release-1.5` branch only, rather than master?

mattcary:
Yes. Cherry-picking will have the same number of CLs, with the disadvantage of merge conflicts to work through.

PR #509, opened 2025-03-07
Put test/ into submodule

mattcary:
Putting test in a submodule means k8s.io/kubernetes is not used in main code, which is not supported and breaks our auto-vc.

The one use of k8s.io/kubernetes in the main code was resolved by copy/paste importing the single used function.

This also updates imported packages to latest versions.

 /kind cleanup

```release-note
Package update and removal of k8s.io/kubernetes dependency.
```


mattcary:
I have not been able to run the e2e tests, the webhook is getting a TLS handshake error.

But at least I've gotten them to run, so I think I've updated correctly for k/k changes (see the minor code changes in test/e2e).

I think we can submit and see what the CI build does with the e2e tests?

hime (approved):

PR #520, opened 2025-03-13
Only pass identity provider mountOption for managed sidecar version >= v1.12.2-gke.0

siyanshen:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR makes sure latest version of driver is compatible with custom or outdated sidecar images.

Fixes #

Tested:
1. e2e test shows version filter works as expected.
```
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=true E2E_TEST_FOCUS=should.gcsfuse.process.succeed.without.missing.flag.error.for.hostnetwork.pods.using.custom.sidecar.container.image STAGINGVERSION=support2 REGISTRY=gcr.io/shensiyan-joonix
```
checked cloud logs 
a. When min version is set as v1.12.2-gke.0, log shows the version was not supported. 
b. When min version is set as v1.6.x-..., log shows the version was supported. 

2. Confirmed official sidecar image can pass the filter to enjoy the feature: Locally updated webhook to set `--shouldInjectSAVol=true`, ran e2e test with a sidecar image of versioin v1.12.2-gke.0. Confirmed the test Pod and IO test bucket via our hostnetwork support feature.

3. Confirmed private sidecar can use this feature when user adds required mountOption in their pod. 
a. Locally updated webhook to set `--shouldInjectSAVol=true`. 
b. Installed onto my test cluster a custom driver and sidecar. 
c. Created a test pod from file
```
apiVersion: v1
kind: Pod
metadata:
  name: test-pv-pod-hnw
  namespace: ns1
  annotations:
    gke-gcsfuse/volumes: "true"
spec:
  serviceAccountName: test-ksa-ns1
  hostNetwork: true
  containers:
  - image: busybox
    name: busybox
    command:
      - sleep
      - "3600"
    volumeMounts:
    - name: gcs-fuse-csi-ephemeral
      mountPath: /dataEph
  volumes:
  - name: gcs-fuse-csi-ephemeral
    csi:
      volumeAttributes:
        bucketName: test-wi-host-network-2
        mountOptions: "implicit-dirs,token-server-identity-provider=https://container.googleapis.com/v1/projects/<my-project>/locations/us-central1/clusters/ssy-e2e"
```
d. Verified that we do not pass the mountOption from driver, but the user pod can successfully IO thanks to the mountOption specified in podSpec. https://screenshot.googleplex.com/9R6sFqrypUhbZ5W

```release-note
Only pass identity provider mountOptions for managed sidecar version >= v1.12.2-gke.0
```

### DIFF
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -105,6 +106,11 @@ func (c *Clientset) ConfigurePodLister(nodeName string) {
 
 		var newInitContainers []corev1.Container
 		for _, cont := range podObj.Spec.InitContainers {
+			if cont.Name == webhook.GcsFuseSidecarName {
+				newInitContainers = append(newInitContainers, cont)
+
+				continue
+			}
 			container := corev1.Container{
 				Name:            cont.Name,
 				SecurityContext: cont.SecurityContext,
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -29,6 +30,7 @@ import (
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 	pbSanitizer "github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
+	"golang.org/x/mod/semver"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -63,6 +65,7 @@ const (
 	VolumeContextKeyPodNamespace        = "csi.storage.k8s.io/pod.namespace"
 	VolumeContextKeyEphemeral           = "csi.storage.k8s.io/ephemeral"
 	VolumeContextKeyBucketName          = "bucketName"
+	tokenServerSidecarMinVersion        = "v1.12.2-gke.0" // #nosec G101
 )
 
 func NewVolumeCapabilityAccessMode(mode csi.VolumeCapability_AccessMode_Mode) *csi.VolumeCapability_AccessMode {
@@ -459,4 +462,25 @@ func getSidecarContainerStatus(isInitContainer bool, pod *corev1.Pod) (*corev1.C
 	}
 
 	return nil, errors.New("the sidecar container was not found")
+}
+
+func isSidecarVersionSupportedForTokenServer(imageName string) bool {
+	managedSidecarPattern := `.*/gke-release(-staging)?/gcs-fuse-csi-driver-sidecar-mounter:v\d+.\d+.\d+-gke\.\d+.*`
+	re := regexp.MustCompile(managedSidecarPattern)
+	isManagedSidecar := re.MatchString(imageName)
+
+	if !isManagedSidecar {
+		klog.Infof("mountOptions should not be passed because this is a private sidecar image %q", imageName)
+
+		return false
+	}
+	imageVersion := strings.Split(strings.Split(imageName, ":")[1], "@")[0]
+	klog.Infof("sidecar image version: %v", imageVersion)
+	if semver.Compare(imageVersion, tokenServerSidecarMinVersion) >= 0 {
+		klog.Infof("sidecar version is supported for token server")
+
+		return true
+	}
+
+	return false
 }
--- a/test/e2e/testsuites/volumes.go
+++ b/test/e2e/testsuites/volumes.go
@@ -321,6 +321,11 @@ func (t *gcsFuseCSIVolumesTestSuite) DefineTests(driver storageframework.TestDri
 		tPod.SetCustomSidecarContainerImage()
 		tPod.SetupVolume(l.volumeResource, volumeName, mountPath, false)
 
+		if configPrefix == specs.EnableHostNetworkPrefix {
+			ginkgo.By("Turn on hostnetwork setting")
+			tPod.EnableHostNetwork()
+		}
+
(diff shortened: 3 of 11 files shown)


test/e2e/testsuites/volumes.go
```diff
@@ -330,15 +335,20 @@ func (t *gcsFuseCSIVolumesTestSuite) DefineTests(driver storageframework.TestDri
...
+			ginkgo.By("Checking that the pod command exits with no error")
+			tPod.VerifyExecInPodSucceed(f, specs.TesterContainerName, fmt.Sprintf("mount | grep %v | grep rw,", mountPath))
+			tPod.VerifyExecInPodSucceed(f, specs.TesterContainerName, fmt.Sprintf("echo 'hello world' > %v/data && grep 'hello world' %v/data", mountPath, mountPath))
+		}
 	}
 
 	ginkgo.It("should store data using custom sidecar container image", func() {
 		testCaseStoreDataCustomContainerImage("")
 	})
+	ginkgo.It("should mount normally for hostnetwork using custom sidecar container image", func() {
```
amacaskill: Not sure what should mount normally means.  Can you please make it more descriptive like "token is not injected for hostnetwork using custom sidecar container image" or whatever you feel captures what the test does

amacaskill:
Is there a way to test this works before we submit?

test/e2e/testsuites/volumes.go
siyanshen: Updated the test name

siyanshen:
> Is there a way to test this works before we submit?

Yeah, updated the PR description field to document my testing method

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -105,6 +106,11 @@ func (c *Clientset) ConfigurePodLister(nodeName string) {
 
 		var newInitContainers []corev1.Container
 		for _, cont := range podObj.Spec.InitContainers {
+			if cont.Name == webhook.GcsFuseSidecarName {
+				newInitContainers = append(newInitContainers, cont)
+
+				continue
+			}
```
amacaskill, line 113: why do we need this? Is this because we need container.Image?

pkg/csi_driver/utils.go
```diff
@@ -460,3 +463,24 @@ func getSidecarContainerStatus(isInitContainer bool, pod *corev1.Pod) (*corev1.C
...
 	return nil, errors.New("the sidecar container was not found")
 }
+
+func isSidecarVersionSupportedForTokenServer(imageName string) bool {
+	managedSidecarPattern := `.*/gke-release/gcs-fuse-csi-driver-sidecar-mounter:v\d+.\d+.\d+-gke\.\d+.*`
+	re := regexp.MustCompile(managedSidecarPattern)
+	isManagedSidecar := re.MatchString(imageName)
+
+	if !isManagedSidecar {
+		klog.Infof("mountOptions should not be passed because this is a private sidecar %v", imageName)
```
amacaskill: ("mountOptions should not be passed because this is a private sidecar image %q", imageName)

pkg/cloud_provider/clientset/clientset.go
amacaskill: Should we just add that, instead of the entire container object?

amacaskill:
Can you please add an e2e test for setting "mountOptions:token-server-identity-provider=https://container.googleapis.com/v1/projects/<my-project>/locations/us-central1/clusters/ssy-e2e" + hostnetwork + private sidecar image works as expected

Also, what happens in our logic if the customer sets "token-server-identity-provider=https://container.googleapis.com/v1/projects/<my-project>/locations/us-central1/clusters/ssy-e2e" in mountOptions even though its not a private registry  or without hostnetwork enabled. What happens in those cases?

pkg/cloud_provider/clientset/clientset.go
siyanshen: Yes. Otherwise this info is filtered out for memory optimization

siyanshen:
> Can you please add an e2e test for setting "mountOptions:token-server-identity-provider=https://container.googleapis.com/v1/projects//locations/us-central1/clusters/ssy-e2e" + hostnetwork + private sidecar image works as expected

I would rather not have this in e2e test. For one thing the functionality can only be verified during I/O, but the webhook feature flag is not enabled by default yet. For another, we only want to officially support the managed sidecar. Implementing tests for private images would add unnecessary complexity. We have to retrieve project id, cluster name, container endpoint information, etc. in the test environment. And the benefit is not that significant.

> Also, what happens in our logic if the customer sets "token-server-identity-provider=https://container.googleapis.com/v1/projects//locations/us-central1/clusters/ssy-e2e" in mountOptions even though its not a private registry or without hostnetwork enabled. What happens in those cases?
If the sidecar detects a tokenServerIdentity is provided, it will start the token server.

pkg/cloud_provider/clientset/clientset.go
siyanshen: I think I saw you added a comment somewhere asking whether we can just add container.Image instead. I did it this way because initContainers is an existing field of podSpec but container image is not

pkg/csi_driver/utils.go
siyanshen: done

pkg/csi_driver/utils.go
```diff
@@ -460,3 +463,24 @@ func getSidecarContainerStatus(isInitContainer bool, pod *corev1.Pod) (*corev1.C
 
 	return nil, errors.New("the sidecar container was not found")
 }
+
+func isSidecarVersionSupportedForTokenServer(imageName string) bool {
+	managedSidecarPattern := `.*/gke-release/gcs-fuse-csi-driver-sidecar-mounter:v\d+.\d+.\d+-gke\.\d+.*`
```
hime: @siyanshen If we were to try this out on staging environment, would this work? Subpath of staging environment is slightly different. E.g. would a test covering this scenario work on the managed driver testgrid running in staging env?

Another potential approach would be to set some type of pod label (e.g environment variable) within the sidecar spec at the time of webhook injection. At this time, the webhook would know if its injecting the managed sidecar or the private sidecar, so we could set an env var `managed_sidecar=true`. We can then read the env var through the pod spec that we save in the pod informer and decide if we should support hostnetworking in that scenario.

siyanshen: edited the regex to match staging gcr too.

hime:
@siyanshen Would be good to have e2e test that @amacaskill mentioned if/when its feasible. We support GCSFuse on private nodes by using the custom sidecar feature. Maybe its also good to test everything on sandbox before cutting release.

Other than that, lgtm!

PR #528, opened 2025-03-17
skip mountOptions test for versions where it is not supported

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

Skip mount options test for versions where mountOption length hasn't been extended. We are only going to add https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/503 into the 1.32 release, so we should skip for versions less than 1.32.

Fixes #

```release-note
Skip mount options test for versions where mountOption length hasn't been extended
```

### DIFF
--- a/test/e2e/utils/cluster.go
+++ b/test/e2e/utils/cluster.go
@@ -34,6 +34,7 @@ var (
 	skipBucketCheckMinimumVersion     = version.MustParseGeneric("1.29.0")
 	kernelReadAheadMinimumVersion     = version.MustParseGeneric("1.32.0")
 	metadataPrefetchMinimumVersion    = version.MustParseGeneric("1.32.0")
+	longMountOptionsMinimumVersion    = version.MustParseGeneric("1.32.0")
 )
 
 func clusterDownGKE(testParams *TestParameters) error {
--- a/test/e2e/utils/handler.go
+++ b/test/e2e/utils/handler.go
@@ -239,6 +239,11 @@ func generateTestSkip(testParams *TestParameters) string {
 		skipTests = append(skipTests, "metadata.prefetch")
 	}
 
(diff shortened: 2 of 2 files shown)


test/e2e/utils/cluster.go
```diff
@@ -34,6 +34,7 @@ var (
 	skipBucketCheckMinimumVersion     = version.MustParseGeneric("1.29.0")
 	kernelReadAheadMinimumVersion     = version.MustParseGeneric("1.32.0")
 	metadataPrefetchMinimumVersion    = version.MustParseGeneric("1.32.0")
+	longMountOptionsMinimumVersion    = version.MustParseGeneric("1.32.0")
```
siyanshen, line 37: Looks like this is failing for 1.32 clusters too: [failed test](https://prow-gob.gcpnode.com/view/gcs/gob-prow/logs/periodic-gke-gcs-fuse-csi-managed-driver-k8s-integration-standard-1.32-regional-prod/1900556579200045056). Can we increase the min version to some future version, like 1.32.3?

hime:
lgtm. note that this test will fail on managed driver until mountOptions fix is out on rapid channel. We can either wait rollout to complete or submit now and track fix.

test/e2e/utils/cluster.go
amacaskill: it will likely end up being 1.32.2-gke.X or 1.32.3, but we don't know that version yet. I'll update it once we know the version. Our tests run on the lastest version for each minor, so if we set it to 1.32.0, it will auto pass once the version with your fix is in rapid. Then I can go back later and add the exact version (in case we ever want to run the tests on more than just latest version in rapid)

PR #533, opened 2025-03-27
Fail NodePublishVolume with actionable error in case Workload Identity is not yet enabled on node level

hungnguyen243:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Add a check for GKE metadata server label on node to fail MountVolume (NodePublishVolume) with an actionable error. The target scenario is when an user enabled Workload Identity on an existing cluster, but didn't update their node pool to use GKE metadata server. Currently, this error takes more effort to debug since NodePublishVolume still succeeds, but we would see R/W errors from the sidecar due to PermissionDenied as the default SA lacks the appropriate permission for the GCS bucket.

Testing documentation with custom driver image: https://paste.googleplex.com/5532151901323264

Fixes #

```release-note
Fail MountVolume with actionable error in case Workload Identity is not yet enabled on node level
```

### DIFF
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -39,9 +39,11 @@ import (
 
 type Interface interface {
 	ConfigurePodLister(nodeName string)
+	ConfigureNodeLister(nodeName string)
 	GetPod(namespace, name string) (*corev1.Pod, error)
 	CreateServiceAccountToken(ctx context.Context, namespace, name string, tokenRequest *authenticationv1.TokenRequest) (*authenticationv1.TokenRequest, error)
 	GetGCPServiceAccountName(ctx context.Context, namespace, name string) (string, error)
+	GetNode(name string) (*corev1.Node, error)
 }
 
 type PodInfo struct {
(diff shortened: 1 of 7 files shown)


pkg/csi_driver/node.go
```diff
@@ -141,6 +141,18 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+	node, err := s.k8sClients.GetNode(s.driver.config.NodeID)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "failed to get node: %v", err)
+	}
+
+	val, ok := node.Labels[clientset.GkeMetaDataServerKey]
+	// If Workload Identity is not enabled, the key should be missing; the check for "val == false" is just for extra caution
+	isWorkloadIdentityDisabled := val == "false" || !ok
+	if isWorkloadIdentityDisabled && !pod.Spec.HostNetwork {
+		return nil, status.Errorf(codes.FailedPrecondition, "Workload Identity Federation is not enabled on node. Please make sure this is enabled on both cluster and node pool level.")
```
amacaskill: nit: no period needed at the end of this sentence

From go style decisions: 
>Error strings should not be capitalized (unless beginning with an exported name, a proper noun or an acronym) and should not end with punctuation. This is because error strings usually appear within other context before being printed to the user.

amacaskill: We could also provide a linkto https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity  in this error message

pkg/csi_driver/node.go
```diff
@@ -141,6 +141,18 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
 	}
 
+	node, err := s.k8sClients.GetNode(s.driver.config.NodeID)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "failed to get node: %v", err)
+	}
+
+	val, ok := node.Labels[clientset.GkeMetaDataServerKey]
+	// If Workload Identity is not enabled, the key should be missing; the check for "val == false" is just for extra caution
+	isWorkloadIdentityDisabled := val == "false" || !ok
```
amacaskill: "false" is not a valid value according to GCW which validates this field: 
```
              allowedKeysWithValues:
              - key: "iam.gke.io/gke-metadata-server-enabled"
                pattern: "true"
```

If you want to keep this extra check, then I think it makes sense to change it to 
isWorkloadIdentityDisabled := val != "true"

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -52,9 +54,58 @@ type PodInfo struct {
...
+		newLabels := map[string]string{}
+		isGkeMetaDataServerEnabled, ok := nodeObj.ObjectMeta.Labels[GkeMetaDataServerKey]
+		if ok {
+			newLabels[GkeMetaDataServerKey] = isGkeMetaDataServerEnabled
+		}
+
+		nodeObj.Spec = corev1.NodeSpec{}
+		nodeObj.Status = corev1.NodeStatus{}
+		nodeObj.ObjectMeta.Annotations = nil
+		nodeObj.ObjectMeta.Labels = newLabels
```
amacaskill, line 88: A couple questions, why do we set nodeObj.ObjectMeta.Annotations = nil, and clear out the nodeObj.spec / status? What are we going to use to get the machineType of the node?

Also , is it intentional we replace the nodeObj.ObjectMeta.Labels to only be the GkeMetaDataServerKey? Is that the part where we "We are filtering only for relevant Node annotations to optimize memory usage." ? Or where does the filtering happen.  


Comment says we do filter for releavant node annotations (its labels), but if GkeMetaDataServerKey isn't there, I tink we

hungnguyen243: IIUC the point of the function is to fetch only the necessary information we want (i.e. constructing a new Node/Pod object with just the info/fields we need from the actual k8s object, rather than passing the whole object around). This is just my guess looking at the way ConfigurePodLister is currently working :)
When we want to get machineType from the labels, we can do similar processing and add the 'machineType' label ("node.kubernetes.io/instance-type") in newLabels.

amacaskill: Ok Sounds good. Thanks!

saikat-royc: this is an optimization trick we used to reduce informer mem usage

PR #547, opened 2025-04-07
Limit the maximum number of registered metric collectors.

hime:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

GCSFuse is designed around exporting metrics from the GCSFuse processes that runs on each sidecar, through node-server container, and going through GKE Metrics collector. Currently we do not have limits set on the number of metrics we can export. We can easily reach limits in the amount of metrics we can export and cause node-server to become unhealthy. Although the only symptom would be metrics not exporting, this would cause our SLOs to trigger on behavior that we do not support.

**Current PR approach**
This PR proposes setting a limit based on the number of combinations of volumes and the pods mounting the volume, that can export metrics. Each GCSFuse CSI volume is backed by a GCSFuse process that runs as part of the sidecar container of the pod mounting the volume, which is where the metrics originate from. If customers exceed this limit, we do not export metrics for any new volumes on that node. 

This approach does not handle node or daemonset restarts for clusters that have exceeded this limit. The ordering when re-registering the metrics collectors after these events would not be deterministic. This is fine because we do not provide a guarantee of metrics working in this scenario.

Fixes #

```release-note
Limit the maximum number of registered metric collectors.
```

### DIFF
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -47,6 +47,7 @@ var (
 	informerResyncDurationSec = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
 	fuseSocketDir             = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
 	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
+	maximumNumberOfCollectors = flag.Int("max-metric-collectors", -1, "Maximum number of prometheus metric collectors exporting metrics at a time, less than 0 (e.g -1) means no limit.")
 
 	// These are set at compile time.
 	version = "unknown"
@@ -109,7 +110,7 @@ func main() {
 		}
 
 		if *metricsEndpoint != "" {
-			mm = metrics.NewMetricsManager(*metricsEndpoint, *fuseSocketDir, clientset)
+			mm = metrics.NewMetricsManager(*metricsEndpoint, *fuseSocketDir, *maximumNumberOfCollectors, clientset)
 			mm.InitializeHTTPHandler()
 		}
 	}
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -26,7 +26,10 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
@@ -56,14 +59,21 @@ type manager struct {
 	metricsEndpoint string
 	fuseSocketDir   string
 	clientset       clientset.Interface
+
+	maximumNumberOfCollectors   int
+	volumePublishPathRegistered sets.Set[string]
+	mutex                       sync.Mutex
 }
 
-func NewMetricsManager(metricsEndpoint, fuseSocketDir string, clientset clientset.Interface) Manager {
+func NewMetricsManager(metricsEndpoint, fuseSocketDir string, maximumNumberOfCollectors int, clientset clientset.Interface) Manager {
 	mm := &manager{
-		registry:        prometheus.NewRegistry(),
-		metricsEndpoint: metricsEndpoint,
-		fuseSocketDir:   fuseSocketDir,
-		clientset:       clientset,
+		registry:                    prometheus.NewRegistry(),
+		metricsEndpoint:             metricsEndpoint,
+		fuseSocketDir:               fuseSocketDir,
+		clientset:                   clientset,
+		volumePublishPathRegistered: sets.Set[string]{},
+		maximumNumberOfCollectors:   maximumNumberOfCollectors,
+		mutex:                       sync.Mutex{},
 	}
 
 	return mm
@@ -115,8 +125,43 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
 		"bucket_name":    bucketName,
 		"pod_uid":        podUID,
 	}, mm.clientset)
-	if err := mm.registry.Register(c); err != nil && !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
-		klog.Errorf("failed to register metrics collector for pod  %v/%v, volume %q, bucket %q: %v", podNamespace, podName, volumeName, bucketName, err)
+
+	// Lock the number of registered collectors while we attempt to register a new collector.
+	mm.mutex.Lock()
+	defer mm.mutex.Unlock()
+
+	if mm.maximumNumberOfCollectors == 0 {
+		klog.Infof("could not register metrics collector: podUID: %s, volume: %s. metrics collector limit is set to zero.", podUID, bucketName)
+
+		return
+	}
+
+	// Check if we need to register collector. We register a collector when the following are met:
+	// 1. There is space on the metrics pipeline for the collector to be registered.
+	// 2. The metrics collector has not previously been registered.
+	if mm.maximumNumberOfCollectors > 0 {
+		// If volume is already registered, do not register again. This flow can get triggered
+		//  since CSI driver has republishVolume capability.
+		if mm.volumePublishPathRegistered.Has(targetPath) {
+			return
+		}
+		// If collector hasn't been registered and there's no space left, log a warning.
+		if mm.volumePublishPathRegistered.Len() >= mm.maximumNumberOfCollectors {
+			klog.V(6).Infof("could not register a metrics collector: podUID: %s, volume: %s. there's already %d collectors registered.", podUID, bucketName, mm.volumePublishPathRegistered.Len())
+
+			return
+		}
+	}
+
+	// Attempt to register new metrics collector and record success.
+	err = mm.registry.Register(c)
+	if err != nil {
+		if !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
+			klog.Errorf("failed to register metrics collector for pod  %v/%v, volume %q, bucket %q: %v", podNamespace, podName, volumeName, bucketName, err)
+		}
+	} else {
+		mm.volumePublishPathRegistered.Insert(targetPath)
+		klog.Infof("successfully registered a new metrics collector: podUID: %s, volume: %s. there's %d collectors registered.", podUID, bucketName, mm.volumePublishPathRegistered.Len())
 	}
 }
 
@@ -126,8 +171,16 @@ func (mm *manager) UnregisterMetricsCollector(targetPath string) {
 
 	// metricsCollector uses a hash of pod UID and volume name as an identifier.
 	c := NewMetricsCollector("", "", "", "", podUID, volumeName, nil, nil)
+
+	// Lock the number of registered collectors while we attempt to unregister a collector.
+	mm.mutex.Lock()
+	defer mm.mutex.Unlock()
+
 	if ok := mm.registry.Unregister(c); !ok {
 		klog.Infof("Unregister metrics collector for targetPath %q is not needed since the collector is not registered", targetPath)
+	} else {
+		mm.volumePublishPathRegistered.Delete(targetPath)
+		klog.Infof("successfully unregistered a metrics collector: podUID: %s, volume: %s. there's %d collectors registered.", podUID, volumeName, mm.volumePublishPathRegistered.Len())
 	}
 }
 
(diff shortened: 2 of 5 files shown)


saikat-royc:
please put detailed context in the PR description

saikat-royc:
minor nit on the PR description: "This PR proposes setting a limit based on the number of volumes that can export metrics. Each GCSFuse CSI volume is backed by a GCSFuse process, which is where the metrics originate from" 

change suggestion:
"This PR proposes setting a limit based on the number of combinations of volumes and the pods mounting the volume, that can export metrics. Each GCSFuse CSI volume is backed by a GCSFuse process that runs as part of the sidecar container of the pod mounting the volume, which is where the metrics originate from"

pkg/metrics/metrics.go
```diff
@@ -115,8 +122,24 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
 		"bucket_name":    bucketName,
 		"pod_uid":        podUID,
 	}, mm.clientset)
-	if err := mm.registry.Register(c); err != nil && !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
-		klog.Errorf("failed to register metrics collector for pod  %v/%v, volume %q, bucket %q: %v", podNamespace, podName, volumeName, bucketName, err)
+
+	// Lock the number of registered collectors while we attemtp to register a new collector.
+	mm.mutex.Lock()
```
saikat-royc, line 130: we already acquire lock in nodepublishvolume https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/547/files#diff-a88a1edc80798adc4ab0772fd137ad8c22e699c910f50d3dc8486f4beef474a1R100

so is this mutex lock required?

pkg/metrics/metrics.go
```diff
@@ -115,8 +122,24 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
+	defer mm.mutex.Unlock()
+
+	// If we already registered all metrics collectors, we stop registration of new collectors.
+	if mm.registeredCollectorsCount >= mm.maximumNumberOfCollectors {
+		return
```
saikat-royc, line 136: we can log a warning message  indicating which pod, volume, bucket was skipped.

pkg/metrics/metrics.go
```diff
@@ -126,8 +149,16 @@ func (mm *manager) UnregisterMetricsCollector(targetPath string) {
 
 	// metricsCollector uses a hash of pod UID and volume name as an identifier.
 	c := NewMetricsCollector("", "", "", "", podUID, volumeName, nil, nil)
-	if ok := mm.registry.Unregister(c); !ok {
+
+	// Lock the number of registered collectors while we attemtp to unregister a collector.
+	mm.mutex.Lock()
```
saikat-royc, line 176: same comment here about lock being acquired in node unpublish.

pkg/metrics/metrics.go
```diff
@@ -115,8 +122,24 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
+	}
+
+	err = mm.registry.Register(c)
+	if err != nil {
+		if !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
+			klog.Errorf("failed to register metrics collector for pod  %v/%v, volume %q, bucket %q: %v", podNamespace, podName, volumeName, bucketName, err)
+		}
+	} else {
+		mm.registeredCollectorsCount += 1
+		klog.V(6).Infof("successfully registered a new metrics collector, currently there's %d collectors registerd.", mm.registeredCollectorsCount)
```
saikat-royc: for debuggability I think we can change this verbosity, and also clearly log which pod, volume, bucket and registeredCollectorsCount was registered.

pkg/metrics/metrics.go
```diff
@@ -126,8 +149,16 @@ func (mm *manager) UnregisterMetricsCollector(targetPath string) {
...
+
+	// Lock the number of registered collectors while we attemtp to unregister a collector.
+	mm.mutex.Lock()
+	defer mm.mutex.Unlock()
+
+	if ok := mm.registry.Unregister(c); ok {
 		klog.Infof("Unregister metrics collector for targetPath %q is not needed since the collector is not registered", targetPath)
+	} else {
+		mm.registeredCollectorsCount -= 1
+		klog.V(6).Infof("successfully unregistered a metrics collector, now there's %d collectors registerd.", mm.registeredCollectorsCount)
```
saikat-royc: same comment here to reduce verbose level and clearly log what pod, volume, bucket and count

cmd/csi_driver/main.go
```diff
@@ -47,6 +47,7 @@ var (
 	informerResyncDurationSec = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
 	fuseSocketDir             = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
 	metricsEndpoint           = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
+	maximumNumberOfCollectors = flag.Int("max-metric-collectors", 10, "Maximum number of prometheus metric collectors exporting metrics at a time, -1 means no limit.")
```
saikat-royc: should we keep the default as -1?

pkg/metrics/metrics.go
hime: I've put the volume name and the Pod UID in log. Its not feasible to put the pod name, bucket name when doing both registration and unregistration of collectors.

pkg/metrics/metrics.go
hime: done.

cmd/csi_driver/main.go
hime: done. set so that any negative number unsets the limit.

pkg/metrics/metrics.go
hime: using thread above to address mutex usage.

pkg/metrics/metrics.go
hime: The mutex lock you have linked prevents overlapping NodePublish and NodeUnpublish volume calls **for the same volume**. This does not prevent two NodePublishVolume calls for different volumes from overlapping. This mutex makes sure that only one volume is able to register at a time for **all** NodePublish and NodeUnpublish volume calls.

pkg/metrics/metrics.go
hime: added set to track which volumes are not registered to then log the failures.

pkg/metrics/metrics.go
saikat-royc: Ack.

pkg/metrics/metrics.go
```diff
@@ -115,8 +125,37 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
-		klog.Errorf("failed to register metrics collector for pod  %v/%v, volume %q, bucket %q: %v", podNamespace, podName, volumeName, bucketName, err)
+
+	// Lock the number of registered collectors while we attempt to register a new collector.
+	mm.mutex.Lock()
+	defer mm.mutex.Unlock()
+
+	// Check if we need to register collector. We register a collector when the following are met:
+	// 1. There is space on the metrics pipeline for the collector to be registered.
+	// 2. The metrics collector has not previously been registered.
+	if mm.maximumNumberOfCollectors >= 0 {
```
saikat-royc: why do we need to register collector for  maximumNumberOfCollectors = 0? We should skip registration I think

pkg/metrics/metrics.go
```diff
@@ -56,14 +59,21 @@ type manager struct {
...
 	fuseSocketDir   string
 	clientset       clientset.Interface
+
+	maximumNumberOfCollectors int
+	volumesRegistered         sets.Set[string]
+	mutex                     sync.Mutex
 }
 
-func NewMetricsManager(metricsEndpoint, fuseSocketDir string, clientset clientset.Interface) Manager {
+func NewMetricsManager(metricsEndpoint, fuseSocketDir string, maximumNumberOfCollectors int, clientset clientset.Interface) Manager {
```
saikat-royc, line 68: Unit test for this change?

pkg/metrics/metrics.go
```diff
@@ -56,14 +59,21 @@ type manager struct {
 	metricsEndpoint string
 	fuseSocketDir   string
 	clientset       clientset.Interface
+
+	maximumNumberOfCollectors int
+	volumesRegistered         sets.Set[string]
```
saikat-royc: nit: volumePublishPathRegistered seems more accurate description for this variable?
Else it may confuse the user to think we are registering a collector per volume. we are registering a collector per pod publish path that is mounting a volume

pkg/metrics/metrics.go
```diff
@@ -126,8 +165,16 @@ func (mm *manager) UnregisterMetricsCollector(targetPath string) {
...
+
+	// Lock the number of registered collectors while we attempt to unregister a collector.
+	mm.mutex.Lock()
+	defer mm.mutex.Unlock()
+
 	if ok := mm.registry.Unregister(c); !ok {
 		klog.Infof("Unregister metrics collector for targetPath %q is not needed since the collector is not registered", targetPath)
+	} else {
+		mm.volumesRegistered.Delete(targetPath)
+		klog.Infof("successfully unregistered a metrics collector: podUID: %s, volume: %s. there's %d collectors registered.", podUID, volumeName, mm.volumesRegistered.Len())
```
saikat-royc: this log can cause confusion since we may or may not have added this path. Maybe we should check for existence of the key before delete and only log if the key actually existed before deletion.

pkg/metrics/metrics.go
hime: We still skip registration on line 143, but we can add a dedicated if statement for this scenario.

pkg/metrics/metrics.go
hime: this sounds more futureproof. done.

pkg/metrics/metrics.go
hime: We only add to the set when we successfully register the collector and we only remove from the set when we successfully unregister the collector. This is all protected by a mutex, do you think this could get out of sync somehow?

pkg/metrics/metrics.go
hime: added

hime: I've also shared a doc on the manual testing done. I think non-trivial to automate this test flow.

pkg/metrics/metrics.go
```diff
@@ -115,8 +125,41 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
 	}, mm.clientset)
-	if err := mm.registry.Register(c); err != nil && !strings.Contains(err.Error(), prometheus.AlreadyRegisteredError{}.Error()) {
-		klog.Errorf("failed to register metrics collector for pod  %v/%v, volume %q, bucket %q: %v", podNamespace, podName, volumeName, bucketName, err)
+
+	// Lock the number of registered collectors while we attempt to register a new collector.
+	mm.mutex.Lock()
+	defer mm.mutex.Unlock()
+
+	if mm.maximumNumberOfCollectors == 0 {
+		klog.Infof("could not register metrics collector: podUID: %s, volume: %s. metrics collector limit is set to zero.", podUID, bucketName)
```
saikat-royc, line 134: return missing

pkg/metrics/metrics.go
```diff
@@ -115,8 +125,41 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
+	// 2. The metrics collector has not previously been registered.
+	if mm.maximumNumberOfCollectors > 0 {
+		// If volume is already registered, do not register again. This flow can get triggered
+		//  since CSI driver has republishVolume capability.
+		if mm.volumePublishPathRegistered.Has(targetPath) {
+			return
+		}
+		// If collector hasn't been registered and there's no space left, log a warning.
+		if mm.volumePublishPathRegistered.Len() >= mm.maximumNumberOfCollectors {
+			klog.Infof("could not register a metrics collector: podUID: %s, volume: %s. there's already %d collectors registered.", podUID, bucketName, mm.volumePublishPathRegistered.Len())
```
saikat-royc: on second thought, this could inundate the logs on every node republish for targetPath that have not been registered. Log at L162 would give us info on what target path have been registered. So I think we can put this log of L148 under a verbosity level V=6 (> default verbosity V=5).

pkg/metrics/metrics.go
hime: done

pkg/metrics/metrics.go
hime: done

pkg/metrics/metrics.go
saikat-royc: ack

pkg/metrics/metrics.go
saikat-royc: ack

pkg/metrics/metrics.go
hime: done.

saikat-royc (approved):

PR #553, opened 2025-04-08
Pass --machine-type and --disable-autoconfig flags from driver to gcsfuse

hungnguyen243:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR bumped gcsfuse binary to v2.12.0-gke-0, where the the new two mountOptions start to be supported. Note that gcsfuse receives these flags but will not turn on the default values until v3.

**Summary of changes:**

As part of GCSFuse's intelligent defaults, new mountOption flags are supported: --machine-type and --disable-autoconfig.
We want to pass these two flags over from the driver to GCSFuse in the following mechanism:
1. --machine-type: The driver has existing mechanism to get Node objects from K8s server starting from [v1.14.0](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/releases/tag/v1.14.0). We're fetching the node object inside NodePublishVolume, get the instance-type label, 
2. --disable-autoconfig: This will be a new driver flag, defaulted to 'false', and added to driver's config, which is accessible from inside NodePublishVolume. This flag denotes whether gcsfuse will apply the machine-type based intelligent defaults.

Both flags are then written as a line-separated map to a temporary file under targetPath, which will be read by the sidecar container upon startup. The sidecar parses the file, reconstructs the map, and merges this map into the config file map. This approach is extensible, allowing us to add  any new flag that we want to pass from the GCSFuse Driver container to the sidecar container and ultimately to the GCSFuse process.

Note that the PR also wraps the flag-writing logic on the driver behind a sidecar version filter, which checks that the sidecar container is a **managed** sidecar and **on a compatible version** (linked to a gcsfuse binary where the new flags are supported). This is to ensure that the file won't be unnecessarily written by the driver if not ultimately consumable by the sidecar (in the case of old/private sidecars). This minimum version is currently using a placeholder of v1.99.0-gke.0 in this PR, and we'll update this with the exact value after we completed the first round of release and get the corresponding sidecar image's version.

**Testing**: 

Flag-passing functionality testing:

The PR has been tested both manually and with two new E2E tests by checking the GCSFuse config logged by the sidecar container:

1. Test that --machine-type and --disable-autoconfig=false (using driver's default value) are passed  correctly to GCSFuse
2. Test that --machine-type and --disable-autoconfig=true (overwritten by the user pod's mountOptions) are passed correctly to GCSFuse

I ran these E2E tests with driver code built without the sidecar version filter, and made sure they're WAI: https://paste.googleplex.com/4777435008335872

Sidecar version filter testing:

1. Added unit tests for sidecar version check for all possible scenarios (compatible managed sidecar, old managed sidecar, private sidecar)
2. Tested that [existing E2E tests](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6113e07468228144f8c149d523763838741f8bb8/test/e2e/testsuites/volumes.go#L346) for custom sidecars with an older gcsfuse tag are [passing](https://paste.googleplex.com/5805843058262016), and the flag file is not written by the driver (by checking driver's logs).

Fixes #

```release-note
Pass --machine-type and --disable-autoconfig flags to gcsfuse to enable intelligent defaults on GCSFuse for high-performance machine types.
```

### DIFF
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -28,6 +28,7 @@ import (
 	"syscall"
 	"time"
 
+	driver "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/csi_driver"
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 	"k8s.io/klog/v2"
@@ -55,12 +56,24 @@ func main() {
 	mounter := sidecarmounter.New(*gcsfusePath)
 	ctx, cancel := context.WithCancel(context.Background())
 
+	flagsFromDriver := map[string]string{}
+	defaultingFlagFilePath := *volumeBasePath + "/" + driver.FlagFileForDefaultingPath
+	klog.Infof("Checking if defaulting-flag file %q exists", defaultingFlagFilePath)
+	if _, err := os.Stat(defaultingFlagFilePath); err == nil {
+		machineTypeBytes, err := os.ReadFile(defaultingFlagFilePath)
+		if err != nil {
+			klog.Fatalf("failed to read defaulting-flag file: %v", err)
+		}
+		fileContent := string(machineTypeBytes)
+		flagsFromDriver = driver.ParseFlagMapFromFlagFile(fileContent)
+	}
+
 	for _, sp := range socketPaths {
 		// sleep 1.5 seconds before launch the next gcsfuse to avoid
 		// 1. different gcsfuse logs mixed together.
 		// 2. memory usage peak.
 		time.Sleep(1500 * time.Millisecond)
-		mc := sidecarmounter.NewMountConfig(sp)
+		mc := sidecarmounter.NewMountConfig(sp, flagsFromDriver)
 		if mc != nil {
 			if err := mounter.Mount(ctx, mc); err != nil {
 				mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to mount bucket %q for volume %q: %v\n", mc.BucketName, mc.VolumeName, err))
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -20,6 +20,7 @@ package driver
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -182,6 +183,20 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 	}
 
+	// Only pass mountOptions flags for defaulting if sidecar container is managed and satisifies min version requirement
+	if s.shouldPassDefaultingFlags(pod) {
+		shouldDisableAutoConfig := s.driver.config.DisableAutoconfig
+		machineType, ok := node.Labels[clientset.MachineTypeKey]
+		if ok {
+			flagMap := map[string]string{"machine-type": machineType, "disable-autoconfig": strconv.FormatBool(shouldDisableAutoConfig)}
+			if err := PutFlagsFromDriverToTargetPath(flagMap, targetPath, FlagFileForDefaultingPath); err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		} else {
+			klog.Warningf("Unable to fetch target node %v's machine type", node.Name)
+		}
+	}
+
 	// Check if there is any error from the gcsfuse
 	code, err := checkGcsFuseErr(isInitContainer, pod, targetPath)
 	if code != codes.OK {
@@ -329,4 +344,17 @@ func (s *nodeServer) shouldStartTokenServer(pod *corev1.Pod) bool {
 	}
 
 	return tokenVolumeInjected && sidecarVersionSupported
+}
+
+func (s *nodeServer) shouldPassDefaultingFlags(pod *corev1.Pod) bool {
+	var sidecarVersionSupported bool
+	for _, container := range pod.Spec.InitContainers {
+		if container.Name == webhook.GcsFuseSidecarName {
+			sidecarVersionSupported = isSidecarVersionSupportedForGivenFeature(container.Image, AutoconfigDefaultingSidecarMinVersion)
+
+			break
+		}
+	}
+
+	return sidecarVersionSupported
 }
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -66,6 +66,9 @@ const (
 	VolumeContextKeyEphemeral           = "csi.storage.k8s.io/ephemeral"
 	VolumeContextKeyBucketName          = "bucketName"
 	tokenServerSidecarMinVersion        = "v1.12.2-gke.0" // #nosec G101
+	// TODO: Update with actual minimum sidecar version after the first feature release
+	AutoconfigDefaultingSidecarMinVersion = "v1.99.0-gke.0"
+	FlagFileForDefaultingPath             = "flags-for-defaulting"
 )
 
 var volumeIDRegEx = regexp.MustCompile(`:.*$`)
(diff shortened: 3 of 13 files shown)


cmd/sidecar_mounter/main.go
```diff
@@ -55,12 +56,24 @@ func main() {
 	mounter := sidecarmounter.New(*gcsfusePath)
 	ctx, cancel := context.WithCancel(context.Background())
 
+	flagsFromDriver := map[string]string{}
+	defaultingFlagFilePath := *volumeBasePath + driver.FlagFileForDefaultingPath
```
amacaskill: This is hard to tell if we are going to have duplicate "/" or not in the path because I need to navigate to a different path to see  that 	FlagFileForDefaultingPath  = "/flags-for-defaulting", then I try to figure out if volumeBasePath ends in a slash or not.

cmd/sidecar_mounter/main.go
```diff
@@ -55,12 +56,24 @@ func main() {
...
+	klog.Infof("Checking if defaulting-flag file exists: %v", defaultingFlagFilePath)
```
amacaskill: %v -> %q

Also nit:  should this be "Checking if defaulting-flag file %q exists"?

pkg/csi_driver/utils.go
```diff
@@ -473,22 +476,73 @@ func getSidecarContainerStatus(isInitContainer bool, pod *corev1.Pod) (*corev1.C
...
 }
+
+func PutFlagsFromDriverToTargetPath(flagMap map[string]string, targetPath string, fileName string) error {
+	emptyDirBasePath, err := util.PrepareEmptyDir(targetPath, true)
+	if err != nil {
+		return fmt.Errorf("failed to get emptyDir path: %w", err)
+	}
+
+	absolutePath := filepath.Dir(emptyDirBasePath) + fileName
+	klog.V(4).Infof("Writing flags needed for gcsfuse defaulting logic to file %v: %v", absolutePath, flagMap)
```
amacaskill: %v for absolute path, should be %q, flag map is fine, but curious if %v for flag map, logs it in a nice way

pkg/csi_driver/utils.go
```diff
@@ -473,22 +476,73 @@ func getSidecarContainerStatus(isInitContainer bool, pod *corev1.Pod) (*corev1.C
...
+	if _, err := f.WriteString(content); err != nil {
+		return fmt.Errorf("failed to write defaulting-flag file: %w", err)
+	}
+
+	f.Close()
+
+	return nil
+}
+
+func prepareFileContentFromFlagMap(flagMap map[string]string) string {
```
amacaskill, line 524: Can you please add a comment here for what the format will look like?

test/e2e/specs/specs.go
```diff
@@ -1237,3 +1248,18 @@ func DeployIstioSidecar(namespace string) {
...
+	stdout, stderr, err := e2ekubectl.RunKubectlWithFullOutput(namespace, "logs", t.pod.Name, "-c", "gke-gcsfuse-sidecar")
+	framework.ExpectNoError(err,
+		"Error accessing logs from pod %v, but failed with error message %q\nstdout: %s\nstderr: %s",
+		t.pod.Name, err, stdout, stderr)
+
+	expectedDisableAutoconfigFlagString := fmt.Sprintf(`"DisableAutoconfig":%t`, expectedDisableAutoconfigFlag)
+	expectedMachineTypeFlagString := fmt.Sprintf(`"MachineType":"%s"`, expectedMachineTypeFlag)
+
+	gomega.Expect(stdout).To(gomega.ContainSubstring(expectedDisableAutoconfigFlagString),
+		"Should find DisableAutoconfig flag string %q in stdout:\n%s", expectedDisableAutoconfigFlagString, stdout)
```
amacaskill: Do you want to log all of the stdout in the test output? I feel like that will make it a little unreadable when you run this test

test/e2e/testsuites/mount.go
```diff
@@ -183,6 +185,40 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
...
+			tPod.VerifyDefaultingFlagsArePassed(f.Namespace.Name, machineType /* disableAutoconfig */, true)
+		} else {
+			tPod.VerifyDefaultingFlagsArePassed(f.Namespace.Name, machineType /* disableAutoconfig */, false)
+		}
+
+		ginkgo.By("Deleting pod")
+		tPod.Cleanup(ctx)
+	}
+
+	ginkgo.It("should pass --machine-type and --disable-autoconfig=false from driver to gcsfuse ", func() {
```
amacaskill, line 214: I remembered you set you are going to add a test that GCSFuse is actually setting the correct defaults based on this machine type, this probably belongs in GCSFuse integration tests somewhere, but we probably need to enable that GCSFuse integration test. Can you reach out to Mauyresh to make sure they have a test for that, so we can enable it to run on GKE?

cmd/sidecar_mounter/main.go
hungnguyen243: Do you have any suggestions? I noticed that the absolute paths are slightly different from the driver vs the sidecar's perspective, so we cannot really share one absolutePath value for both. For example, the absolute path we're writing to in NodePublishVolume for one of my test runs is /var/lib/kubelet/pods/29b6af40-ddc2-4df6-a057-3f6e4537b509/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp/.volumes/flags-for-defaulting, whereas the corresponding path when read from sidecar container is just /gcsfuse-tmp/.volumes/flags-for-defaulting

test/e2e/specs/specs.go
hungnguyen243: I've found having access to the stdout as part of the test output helpful in case the test fails, as I can easily check the gcsfuse config map right from the test's output instead of having to do the extra step of searching through the sidecar container's logs. Here's an example test log from one of my test runs: https://paste.googleplex.com/4996591494365184. Given that it doesnt't cause too much trouble, I wonder if we could keep this in light of the debuggability aspect?

amacaskill: "I don't see "Should find DisableAutoconfig flag string" in the output you shared? Is that the full output or only a part of it?

cmd/sidecar_mounter/main.go
amacaskill: I think just have FlagFileForDefaulting be the file name, then add the / in where needed. You call it a file name in the variable, so it should just be file  name.

test/e2e/specs/specs.go
hungnguyen243: Ah you're right. I realized this logging is not part of ginkgo.Expect(), which only logs in case of test failures but from the implementation of e2ekubectl.RunKubectlWithFullOutput: https://github.com/kubernetes/kubernetes/blob/03a3c0c89161935bc2338f5754ebb1104f779af1/test/e2e/framework/kubectl/builder.go#L157. That's why we're seeing the logs even in successful runs. I updated the code to remove this extra logging since we already have stdout logged anw.

test/e2e/testsuites/mount.go
hungnguyen243: Resolved offline; gcsfuse don't have integration tests due to test infra limitations but they did unit testing + manual testing.

amacaskill (approved):

pkg/csi_driver/node.go
```diff
@@ -182,6 +183,20 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+	machineType, ok := node.Labels[clientset.MachineTypeKey]
+	if ok {
+		flagMap := map[string]string{"machine-type": machineType, "disable-autoconfig": strconv.FormatBool(shouldDisableAutoConfig)}
+		if err := PutFlagsFromDriverToTargetPath(flagMap, targetPath, FlagFileForDefaultingPath); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	} else {
+		klog.Warningf("Unable to fetch target node %v's machine type", node.Name)
+	}
+	// }
```
amacaskill: I think we can remove this

hungnguyen243: Uncommented. This was left commented out during testing.

cmd/sidecar_mounter/gcsfuse_binary
```diff
@@ -1 +1 @@
-gs://gke-release-staging/gcsfuse/v2.11.1-gke.0/gcsfuse_bin
```
amacaskill: Please leave new line so this shows no change

PR #559, opened 2025-04-09
Allow multiple volume handles with different mount options

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
 /kind feature
> /kind flake

When users try to use the same bucket in two different persistent volumes (same value for bucket handle field) and mount these pvs to the same node kubelet does not call node publish volume a second time it treats both volumes as the same which prevents containers from starting (the second volume is not mounted at the expected path). This is a limitation of the gcsfuse csi driver as described [here](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/test/e2e/testsuites/multivolume.go#L237)
Fixes #395976718

```release-note
This release introduces support for mounting two persistent volumes to the same node by allowing users to append a custom suffix to the volume handle (e.g., my-bucket:suffix). The suffix is ignored during NodePublishVolume to ensure the correct bucket is mounted, but is retained in earlier stages to let kubelet treat the volumes as unique. Bucket handles a:b:c are valid but recall the [bucket naming conventions](https://cloud.google.com/storage/docs/buckets#:~:text=Bucket%20names%20can%20only%20contain%20lowercase%20letters%2C%20numeric%20characters%2C%20dashes%20(%2D)%2C%20underscores%20(_)%2C%20and%20dots%20(.).%20Spaces%20are%20not%20allowed.%20Names%20containing%20dots%20require%20verification.). In the case of bucket handle "a:b:c" the csi driver will treat it as a bucket name "a"

This unblocks use cases where workloads need to mount the same GCS bucket with distinct persistent volumes. 
For example usage please review PV1 and PV2 under How was this test/validated. Note the different volume handles being used in each persistent volume - volumeHandle: fuechr_fuse_csi_test:1, volumeHandle: fuechr_fuse_csi_test
these will access the same bucket 

To mount the same GCS Bucket with different persistent volumes, use the following syntax for the volumeHandle field in the PersistentVolume object: "BUCKET_NAME:RANDOM_SUFFIX" , where RANDOM_SUFFIX is a random string of your choice and BUCKET_NAME is your bucket name.

For example : "my_bucket:xyz123"
```
**Root Cause - suspected**
This behavior originates from a limitation in the GCS Fuse CSI driver, which currently does not support mounting multiple PVs with the same volume handle (i.e., pointing to the same bucket) on a single node. This restriction aligns with what is documented in the driver’s [test suite](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/test/e2e/testsuites/multivolume.go#L237).

**Problem Verification**

To reproduce the issue, a cluster was deployed using the latest version of the GCS Fuse CSI driver. All the resources listed under Resources were applied. During testing, it was observed that NodePublishVolume was invoked only once, resulting in one of the pods failing to start due to the absence of the mounted volume.

To verify that the root cause was the reuse of the same bucket handle across both PVs, a new test was conducted using the same setup, but this time with pv1 and pv2 pointing to different buckets (i.e., using volume handles fuechr_fuse_csi_test and fuechr_fuse_csi_test2, respectively). In this scenario, NodePublishVolume was called twice, and both pods started successfully, confirming that the issue only arises when the same bucket is used in both PVs.

**Proposed Solution**
To resolve the issue where NodePublishVolume is not invoked due to duplicate volume handles referencing the same bucket, we adopted a strategy similar to that used by ParallelStore. This approach introduces a [special suffix](https://cloud.google.com/parallelstore/docs/connect-from-kubernetes-engine-existing-instance#same-instance-different-mount-options) to the volume handle, making each volume unique even if they point to the same underlying bucket.

During the NodePublishVolume operation, the CSI driver strips this suffix from the bucket name before performing the mount. This logic has been implemented in the parseVolumeId function, available in utils.go.

The function works as follows:

Checks for the presence of a suffix, delimited by a colon (:),

Removes the suffix if found,

Returns the base bucket name (volume handle).

Unit tests have been added to verify correct parsing and behavior for volume handles both with and without a suffix.
Importantly, this implementation assumes that bucket handles will contain at most one colon (:). This assumption is safe, as Google Cloud bucket names do not support colons, ensuring that any colon present is introduced solely by the CSI driver for suffixing purposes.

**Solution Verification**

To validate the fix, a custom driver image built from the PR was deployed to a cluster in a Google-managed project. The test pods were re-applied using the same bucket for both PVs, but with unique suffixed volume handles.

As expected:

NodePublishVolume was invoked twice—once for each suffixed volume handle.

Both pods successfully transitioned to the Running state.

Driver logs confirmed that the bucket handle suffixes were properly stripped during mount operations, and the correct underlying bucket was used in both cases.

This confirms that the suffix-handling logic works as intended and allows multiple PVs backed by the same bucket to mount successfully on the same node.

**Resources**:
**PV1:** 
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: pv1
  annotations:
    gke-gcsfuse/volumes: "true"
spec:
  capacity:
    storage: 1Gi
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain
  csi:
    **volumeHandle: fuechr_fuse_csi_test:1**
  claimRef:
    name: pvc1  
    namespace: default
``` 

**PVC1:** 
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc1
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  volumeName: pv1
``` 
**PV2:** 
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: pv2
  annotations:
    gke-gcsfuse/volumes: "true"
spec:
  capacity:
    storage: 1Gi
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain
  csi:
    **volumeHandle: fuechr_fuse_csi_test**
  claimRef:
    name: pvc2    
    namespace: default
``` 
**PVC2:** 
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc2
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  volumeName: pv2
``` 
**Pod spec:** 
``` 
apiVersion: v1
kind: Pod
metadata:
  name: pod
  annotations:
    gke-gcsfuse/volumes: "true"
spec:
  containers:
    - name: gcsfuse-container
      image: google/cloud-sdk:alpine
      command: ["sleep", "3600"]
      volumeMounts:
        - name: gcsfuse-volume-1
          mountPath: /mnt/gcsfuse-1
          subPath: gcsfuse-mount
        - name: gcsfuse-volume-2
          mountPath: /mnt/gcsfuse-2
          subPath: gcsfuse-mount
  volumes:
    - name: gcsfuse-volume-1
      persistentVolumeClaim:
        claimName: pvc1
    - name: gcsfuse-volume-2
      persistentVolumeClaim:
        claimName: pvc2
```

### DIFF
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -68,6 +68,8 @@ const (
 	tokenServerSidecarMinVersion        = "v1.12.2-gke.0" // #nosec G101
 )
 
+var volumeIDRegEx = regexp.MustCompile(`:.*$`)
+
 func NewVolumeCapabilityAccessMode(mode csi.VolumeCapability_AccessMode_Mode) *csi.VolumeCapability_AccessMode {
 	return &csi.VolumeCapability_AccessMode{Mode: mode}
 }
@@ -271,14 +273,13 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	}
 
 	vc := req.GetVolumeContext()
-	bucketName := req.GetVolumeId()
+	bucketName := parseVolumeID(req.GetVolumeId())
 	if vc[VolumeContextKeyEphemeral] == util.TrueStr {
 		bucketName = vc[VolumeContextKeyBucketName]
 		if len(bucketName) == 0 {
 			return "", "", nil, false, false, fmt.Errorf("NodePublishVolume VolumeContext %q must be provided for ephemeral storage", VolumeContextKeyBucketName)
 		}
 	}
-
 	fuseMountOptions := []string{}
 	if req.GetReadonly() {
 		fuseMountOptions = joinMountOptions(fuseMountOptions, []string{"ro"})
@@ -300,6 +301,13 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	}
 
 	return targetPath, bucketName, fuseMountOptions, skipCSIBucketAccessCheck, enableMetricsCollection, nil
+}
+
+// The format allows customers to specify a fake volume handle for static provisioning,
+// enabling multiple PVs in the same pod to mount the same bucket. This prevents Kubelet from
+// skipping mounts of volumes with the same volume handle, which can cause the pod to be stuck in container creation.
+func parseVolumeID(bucketHandle string) string {
+	return volumeIDRegEx.ReplaceAllString(bucketHandle, "")
 }
 
 func putExitFile(pod *corev1.Pod, targetPath string) error {
--- a/pkg/csi_driver/utils_test.go
+++ b/pkg/csi_driver/utils_test.go
@@ -29,6 +29,56 @@ const (
 	TraceStr = "trace"
 )
 
+func TestRemoveBucketSuffixIfPresentAndReturnVolumeId(t *testing.T) {
+	t.Parallel()
+	t.Run("removing bucket suffix if present and returning volume id", func(t *testing.T) {
+		t.Parallel()
+		testCases := []struct {
+			name          string
+			bucketName    string
+			expectedValue string
+		}{
+			{
+				name:          "should return bucket name without suffix, no special character",
+				bucketName:    "bucket-name:1234567890123456789",
+				expectedValue: "bucket-name",
+			},
+			{
+				name:          "should return bucket name without suffix, special character",
+				bucketName:    "bucket-name:1234567890123456789@us-central1",
+				expectedValue: "bucket-name",
+			},
+			{
+				name:          "should return bucket name without suffix, no colon with special character",
+				bucketName:    "bucket-name@us-central1",
+				expectedValue: "bucket-name@us-central1",
+			},
+			{
+				name:          "should return bucket name without suffix, two colons",
+				bucketName:    "bucket-name:1234567890123456789:12",
+				expectedValue: "bucket-name",
+			},
+			{
+				name:          "should return bucket name without suffix, no special character and no colon",
+				bucketName:    "bucket-name",
+				expectedValue: "bucket-name",
+			},
+			{
+				name:          "should return bucket name without suffix, no bucket name",
+				bucketName:    ":xyz",
+				expectedValue: "",
+			},
+		}
+
+		for _, tc := range testCases {
+			t.Logf("test case: %s", tc.name)
+			actual := parseVolumeID(tc.bucketName)
+			if actual != tc.expectedValue {
+				t.Errorf("Got value %v, but expected %v", actual, tc.expectedValue)
+			}
+		}
+	})
+}
 func TestJoinMountOptions(t *testing.T) {
 	t.Parallel()
 	t.Run("joining mount options into one", func(t *testing.T) {
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -76,6 +76,9 @@ func (s *controllerServer) ControllerGetCapabilities(_ context.Context, _ *csi.C
 func (s *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req *csi.ValidateVolumeCapabilitiesRequest) (*csi.ValidateVolumeCapabilitiesResponse, error) {
 	// Validate arguments
 	volumeID := req.GetVolumeId()
+	if req.GetVolumeContext()[VolumeContextKeyEphemeral] != util.TrueStr {
+		volumeID = parseVolumeID(volumeID)
+	}
 	if len(volumeID) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "ValidateVolumeCapabilities volumeID must be provided")
 	}


hime:
Adding @uriel-guzman to take a look as he implemented this change in an internal driver.

pkg/csi_driver/utils.go
```diff
@@ -302,6 +303,12 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	return targetPath, bucketName, fuseMountOptions, skipCSIBucketAccessCheck, enableMetricsCollection, nil
 }
 
+// removes bucket suffix if is present, else leaves it alone and returns it
+func removeBucketSuffixIfPresentAndReturnVolumeId(bucketHandle string) string {
+	regEx := regexp.MustCompile(`:.*$`)
```
uriel-guzman: What would happen if the customer introduces the following string: `bucketName:x:y:z`? Can we add a test case for that?

pkg/csi_driver/utils.go
```diff
@@ -302,6 +303,12 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
...
+	return regEx.ReplaceAllString(bucketHandle, "")
```
uriel-guzman: Have we validated that a bucket can't accept `:` as part of the bucket name?

pkg/csi_driver/utils.go
```diff
@@ -302,6 +303,12 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	return targetPath, bucketName, fuseMountOptions, skipCSIBucketAccessCheck, enableMetricsCollection, nil
 }
 
+// removes bucket suffix if is present, else leaves it alone and returns it
```
uriel-guzman: Can we specify exactly why we're doing this? What problem is it trying to solve?

Example:
```
// The format allows customers to specify a fake volume handle for static provisioning,
// enabling multiple PVs in the same pod to mount the same bucket. This prevents Kubelet from
// skipping mounts of volumes with the same volume handle, which can cause the pod to be stuck in container creation.
```

uriel-guzman (commented):
Can we update the PR description to include the full pod spec and have we validated that without this change it doesn't work?

pkg/csi_driver/utils.go
chrisThePattyEater: see above

pkg/csi_driver/utils.go
chrisThePattyEater: This is not a valid case - Use only lowercase letters, numbers, hyphens (-), and underscores (_). Dots (.) may be used to form a valid domain name.

uriel-guzman: That's for the bucket name. But what if they add a double suffix like my example?

If customer introduces `bucketName:x:y:z`, will the bucket be parsed to `bucketName:x:y` or as `bucketName`?

chrisThePattyEater: added test case to cover

pkg/csi_driver/utils.go
uriel-guzman: The above comment was for validating if `x:y:z` gets transformed to `x` or `x:y`. We still need to make sure if the buckets don't accept `:`. Could we link a reference that makes sure of that in the PR description?

pkg/csi_driver/utils.go
```diff
@@ -302,6 +303,16 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	return targetPath, bucketName, fuseMountOptions, skipCSIBucketAccessCheck, enableMetricsCollection, nil
 }
 
+/*
+Removes the bucket suffix (after the last colon) if present.
+This allows users to specify multiple volume handles for the same bucket.
+If no suffix is present, the original bucket handle is returned unchanged.
+*/
+func removeBucketSuffixIfPresentAndReturnVolumeId(bucketHandle string) string {
```
uriel-guzman: Can we change `removeBucketSuffixIfPresentAndReturnVolumeId` to `parseVolumeID`. I think this function's name is extremely specific on the implementation details.

uriel-guzman:
The PR description is outdated. It seems to be using ephemeral volume mounting instead of PV mounting.

pkg/csi_driver/utils.go
```diff
@@ -271,9 +271,10 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	}
 
 	vc := req.GetVolumeContext()
-	bucketName := req.GetVolumeId()
+	bucketName := removeBucketSuffixIfPresentAndReturnVolumeId(req.GetVolumeId())
+
 	if vc[VolumeContextKeyEphemeral] == util.TrueStr {
-		bucketName = vc[VolumeContextKeyBucketName]
+		bucketName = removeBucketSuffixIfPresentAndReturnVolumeId(vc[VolumeContextKeyBucketName])
```
hime: We can just call this function once after the if statement to keep things simple and readable.

pkg/csi_driver/utils.go
```diff
@@ -302,6 +303,16 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	return targetPath, bucketName, fuseMountOptions, skipCSIBucketAccessCheck, enableMetricsCollection, nil
 }
 
+/*
+Removes the bucket suffix (after the last colon) if present.
+This allows users to specify multiple volume handles for the same bucket.
+If no suffix is present, the original bucket handle is returned unchanged.
+*/
+func removeBucketSuffixIfPresentAndReturnVolumeId(bucketHandle string) string {
+	regEx := regexp.MustCompile(`:.*$`)
```
hime: Can you can make this regex a local variable for the entire package? wondering if running this over and over is required.

pkg/csi_driver/utils_test.go
```diff
@@ -29,6 +29,51 @@ const (
...
+	t.Run("removing bucket suffix if present and returning volume id", func(t *testing.T) {
+		t.Parallel()
+		testCases := []struct {
+			name          string
+			bucketName    string
+			expectedValue string
+		}{
+			{
+				name:          "should return bucket name without suffix",
+				bucketName:    "bucket-name:1234567890123456789",
```
hime, line 43: Are all these valid GCS bucket names? Could you add comments explaining what the rules are when it comes to naming a valid GCS bucket?

pkg/csi_driver/utils.go
uriel-guzman: If you do this outside the if statement, the line below will not emit an error on `:x`, while it will be parsed to an empty string, which will create a new bug. I also think this scenario should be added to the unit tests.

pkg/csi_driver/utils.go
uriel-guzman: And also add the link in the comments.

chrisThePattyEater: https://cloud.google.com/storage/docs/buckets#:~:text=Bucket%20names%20can%20only%20contain%20lowercase%20letters%2C%20numeric%20characters%2C%20dashes%20(%2D)%2C%20underscores%20(_)%2C%20and%20dots%20(.).%20Spaces%20are%20not%20allowed.%20Names%20containing%20dots%20require%20verification.

pkg/csi_driver/utils.go
```diff
@@ -302,6 +305,15 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	return targetPath, bucketName, fuseMountOptions, skipCSIBucketAccessCheck, enableMetricsCollection, nil
 }
 
+/*
+Removes the bucket suffix (after the last colon) if present.
+This allows users to specify multiple volume handles for the same bucket.
```
uriel-guzman: Can you explain why this is needed in more detail? What happens if you don't specify it? There's no mention of Kubelet nor NodePublishVolume nor PV, which are the primary reason of this change.

pkg/csi_driver/utils.go
uriel-guzman: Can you just use this description instead? It's more informative to the reader on exactly why it's needed (e.g. it mentions Kubelet, static provisioning, persistent volume, mounting, end effect, etc.

pkg/csi_driver/utils.go
uriel-guzman: Can you add this link in the comment? https://cloud.google.com/storage/docs/buckets#naming

uriel-guzman:
It's also hard to understand what is being tested in the PR description. I just see 5 specs that were pasted. Could we be more  descriptive? (e.g. "observed behavior", "root cause", "solution", "verification")

pkg/csi_driver/utils.go
```diff
@@ -278,7 +280,7 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 			return "", "", nil, false, false, fmt.Errorf("NodePublishVolume VolumeContext %q must be provided for ephemeral storage", VolumeContextKeyBucketName)
 		}
 	}
-
+	bucketName = parseVolumeID(bucketName)
```
uriel-guzman: As I mentioned in the previous comment, moving this here is going to cause a potential bug.

1. Customer introduces `:x` string.
2. Line `279` is going to evaluate the `len(bucketName) == 0` to be false.
3. Line `280` is not going to correctly emit the error.
4. Line `283` is going to strip the bucket and make it an empty string.
5. Best case scenario, this is caught somewhere else. Worst case scenario, this causes unexpected behavior.

Can we return this parsing back to line `278`?

uriel-guzman:
The link to the Parallelstore documentation seems to not work. I think you want: https://cloud.google.com/parallelstore/docs/connect-from-kubernetes-engine-existing-instance#same-instance-different-mount-options

uriel-guzman:
> During the NodePublishVolume operation, the CSI driver strips this suffix from the bucket name before performing the mount. This logic has been implemented in the parseVolumeId function, available in [this PR](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/559/files).

Why is this PR description providing a link to itself? Could we remove this?

uriel-guzman:
Can we add code markdown to your PV specs to make the description more readable?

pkg/csi_driver/utils.go
```diff
@@ -271,14 +273,13 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	}
 
 	vc := req.GetVolumeContext()
-	bucketName := req.GetVolumeId()
+	bucketName := parseVolumeID(req.GetVolumeId())
 	if vc[VolumeContextKeyEphemeral] == util.TrueStr {
-		bucketName = vc[VolumeContextKeyBucketName]
+		bucketName = parseVolumeID(vc[VolumeContextKeyBucketName])
```
uriel-guzman: This seems to be applying the parsing to the ephemeral volume ID as well. I don't think we should scope this outside of the PV `volumeHandle` use-case, as ephemeral mounting does not have this problem and already supports mounting the same instance with different mount-options. In the upcoming documentation, let's make it clear that this workaround is for PV `volumeHandle` only as well.

pkg/csi_driver/utils.go
```diff
@@ -271,14 +273,13 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest) (string, string, [
 	}
 
 	vc := req.GetVolumeContext()
-	bucketName := req.GetVolumeId()
+	bucketName := parseVolumeID(req.GetVolumeId())
```
uriel-guzman, line 276: Have we exhausted all of the references to `GetVolumeId()` or `VolumeID`? We need to make sure that the volumeID is always parsed. 

What about https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/controller.go#L78?

Or what about https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/controller.go#L190?

I'm not familiar if the GCS Fuse CSI Driver controller is used or not. We should add @saikat-royc as a reviewer here as well to make sure.

pkg/csi_driver/utils_test.go
```diff
@@ -29,6 +29,51 @@ const (
...
+				name:          "should return bucket name without suffix",
+				bucketName:    "bucket-name@us-central1",
+				expectedValue: "bucket-name@us-central1",
+			},
+			{
+				name:          "should return bucket name without suffix",
+				bucketName:    "bucket-name:1234567890123456789:12",
+				expectedValue: "bucket-name",
+			},
+			{
```
uriel-guzman, line 61: Can we add a test case for `:suffix`? It should return an empty string.

uriel-guzman:
After these final changes, I will LGTM. The only part I'm not sure about is the controller level changes (e.g. this is missing the parsing in the controller.go for `ControllerGetCapabilities` and I think the change should be applied there as well). Adding @saikat-royc as the final reviewer.

pkg/csi_driver/utils.go
uriel-guzman: As discussed offline with @saikat-royc, let's add the parsing to the `ValidateVolumeCapabilities` call (and everywhere else that `GetVolumeId()` is referenced as the parsing should be a no-op to a correct volume ID.

uriel-guzman (approved):

pkg/csi_driver/utils.go
saikat-royc: +1 to ensure any reference of volumeID needs to handle suffix. 

E.g https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/controller.go#L94C34-L94C95 this would break CheckBucketExists, if we dont handle suffix. and also update any controller unit test

uriel-guzman (approved):

pkg/csi_driver/utils_test.go
```diff
@@ -29,6 +29,56 @@ const (
...
+				name:          "should return bucket name without suffix",
+				bucketName:    "bucket-name",
+				expectedValue: "bucket-name",
+			},
+			{
+				name:          "should return bucket name without suffix",
```
uriel-guzman: These test names should all be different.

uriel-guzman (changes requested):

uriel-guzman (approved):

PR #582, opened 2025-04-28
Hash socket name to prevent long socket names from causing issues.

uriel-guzman:
**What type of PR is this?**

/kind bug

Currently, the fuse socket consists of the following format: `/sockets/PodUID_VolumeID/sockets`. Customers specifying long volume names can make the path exceed 104 chars, which will lead to`bind: invalid argument` errors: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blame/eeeb147f62994c559bd423221f2cef72bac68658/pkg/csi_mounter/csi_mounter.go#L211-L215. 

A solution ot this is to change  `/sockets/x/sockets` to be `/sockets/sha1Hash(x)/sockets`. The sha1Hash will always be of 40 chars, which makes the entire string never be longer than 66 chars, fixing this problem.

Fixes #

```release-note
Apply hashing to the Fuse socket path to avoid long volumes from exceeding 104 char limit and causing `bind: invalid argument` errors.
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -18,6 +18,7 @@ limitations under the License.
 package util
 
 import (
+	"crypto/sha1"
 	"fmt"
 	"net/url"
 	"os"
@@ -159,8 +160,21 @@ func PrepareEmptyDir(targetPath string, createEmptyDir bool) (string, error) {
 	return emptyDirBasePath, nil
 }
 
+// sha1Hash function takes a string as input and returns its SHA1 hash as a hexadecimal string.
+func sha1Hash(str string) string {
+	return fmt.Sprintf("%x", sha1.Sum([]byte(str)))
+}
+
+// GetSocketBasePath constructs the base path for a Unix domain socket.
+// It takes the target path and the directory where Fuse sockets are stored as input.
 func GetSocketBasePath(targetPath, fuseSocketDir string) string {
 	podID, volumeName, _ := ParsePodIDVolumeFromTargetpath(targetPath)
 
-	return filepath.Join(fuseSocketDir, podID+"_"+volumeName)
+	// We hash the combined pod ID and volume name because Unix domain socket paths
+	// have a length limitation (often around 104 characters). Directly using potentially
+	// long pod IDs and volume names could easily violate this limit. SHA1 provides a
+	// fixed-length, short representation (40 hexadecimal characters) that ensures
+	// we stay within these constraints while still providing a unique identifier
+	// based on the pod and volume.
+	return filepath.Join(fuseSocketDir, sha1Hash(podID+"_"+volumeName))
 }
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -18,7 +18,10 @@ limitations under the License.
 package util
 
 import (
+	"crypto/sha1"
 	"fmt"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


hime:
Could you add unit tests or e2e tests showing support for longer volume names? Please also add release notes.

Also wondering (non-blocking):
- Is there a limit to the length of the volume?
- Do we lose visibility when it comes to debugging if we hash the volume name?

uriel-guzman:
@hime

> Could you add unit tests

Done.

> Please also add release notes

Done.

> run e2e tests locally showing no regressions

Running. I'll come back with results.

> Is there a limit to the length of the volume in k8s?

IIUC, there isn't a hard, universally fixed limit to the size of a Kubernetes Persistent Volume Claim (PVC) in terms of the length of the name. However, whatever that limit is (i.e. if it's 253 chars), this bug in the GCS Fuse CSI Driver will limit that to 51 characters, because the len(`/sockets/<PodUID>_<VolumeName>/socket`) needs to be <= 104. `<podUID>` is fixed to 36 characters. This fix is to remove that unintended constraint.

> Do we lose visibility when it comes to debugging if we hash the volume name?

Maybe, it depends. Is there any useful information that would be gained by preserving the socket name containing the PodUID or the volume name in the logs? And is that hypothetical gain better worth preserving than unblocking customers hitting this issue because they specified a long volume name? IIUC, this socket is for internal use only.

uriel-guzman:
I ran a small subset of tests:
```
[ReportAfterSuite] PASSED [24.849 seconds]
autogenerated by Ginkgo

Ran 4 of 422 Specs in 37.602 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 418 Skipped
```

uriel-guzman:
Small manual test for the fix:

1. Before this fix:
```
/csi.v1.Node/NodePublishVolume failed with error: rpc error: code = Internal desc = failed to mount volume "uriel-bucket" to target path "/var/lib/kubelet/pods/94d0fca3-0195-465a-9459-c55eb55abe8b/volumes/kubernetes.io~csi/pvc-1111111111111111111111111111111111111111111111111111/mount": failed to create a listener using the socket: listen unix /sockets/94d0fca3-0195-465a-9459-c55eb55abe8b_pvc-1111111111111111111111111111111111111111111111111111/socket: bind: invalid argument
```
2. After this fix:
```
NodePublishVolume succeeded on volume "uriel-bucket" to target path "/var/lib/kubelet/pods/722e7572-6249-4580-9bda-a8adacb0d498/volumes/kubernetes.io~csi/pvc-1111111111111111111111111111111111111111111111111111/mount"
```
Confirmed pod is running and bucket is readable / writable by SSH'ing into container.

Here's the test spec used:
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: pvc-1111111111111111111111111111111111111111111111111111
spec:
  storageClassName: ""
  accessModes:
    - ReadWriteMany
  capacity:
    storage: 10Gi
  csi:
    volumeHandle: uriel-bucket
  persistentVolumeReclaimPolicy: Retain
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-gcs-fuse-pvc
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 10Gi
  volumeName: pvc-1111111111111111111111111111111111111111111111111111
  storageClassName: ""
---
apiVersion: v1
kind: Pod
metadata:
  name: gcs-fuse-test-pod
  annotations:
    gke-gcsfuse/volumes: "true"
spec:
  containers:
    - name: my-container
      image: busybox
      command: [ "sleep", "infinity" ]
      volumeMounts:
        - name: gcs-volume
          mountPath: /mnt/gcs
  serviceAccountName: my-ksa
  volumes:
    - name: gcs-volume
      persistentVolumeClaim:
        claimName: my-gcs-fuse-pvc
```

amacaskill (approved):

PR #589, opened 2025-05-06
Skip E2E tests for gcsfuse's defaulting flags while min sidecar version is not known

hungnguyen243:
…on is unknown

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Skip E2E tests for gcsfuse's defaulting flags while we don't have a released sidecar version with gcsfuse v2.12 (starting version for the new flags' support). 
We have a sidecar version filter for this feature. The version is currently using a placeholder sidecar version of v1.99.0-gke.0, so related E2E tests will fail expectedly. We'll run these E2E tests once we've finished the first round of release and have a min sidecar version to replace the placeholder.

I invoked these E2E tests locally and made sure they're skipped.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/testsuites/mount.go
+++ b/test/e2e/testsuites/mount.go
@@ -212,10 +212,14 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
 	}
 
 	ginkgo.It("should pass --machine-type and --disable-autoconfig=false from driver to gcsfuse ", func() {
+		// TODO: remove this skip after we've updated the minimum sidecar version that supports the feature
+		e2eskipper.Skipf("skipping machine type defaulting test while new sidecar version is not known")
 		testDefaultingFlags()
 	})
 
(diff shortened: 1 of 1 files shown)


amacaskill:
Testskips will eventually need to go here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/c3523230a04bbfeac0d393d237311682ef97401e/test/e2e/utils/handler.go#L223. 

Note, we need to skip this test for GKE cluster version less than X. Its ok if you don't know this yet, but we need to fill this in eventually, so keep this in mind when adding test skip. This test should never run on managed testgrid for GCSFUse CSI Driver versions targeting previous GKE minors, this feature only goes in the latest GKE minor

amacaskill (approved):

PR #592, opened 2025-05-07
update dependencies, and add go mod tidy to test make commands

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

update dependencies, and add go mod tidy to test make commands. This prevents the GCSFUse presubmits from failing with "you need to run go mod tidy" caused by the go module in test package, being out of date. I basically just copied the changes from  https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/591, but then had to skip the upgrade to  
github.com/google/cel-go  , since 0.25.0 has a breaking change that isn't compatible with k8s.io/apiserver v0.30.10. Perhaps I also could have bumped k8s.io/apiserver. to find version that works for both, but I will do that later. 

Fixes #

```release-note
NONE
```


mattcary:
Sigh, seems to be some problem still with the vendor versions.

mattcary (approved):

PR #605, opened 2025-05-15
Add retry logic while fetching STS token in sidecar mounter

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature
> /kind flake

The PR performs bucket access check in GCS Fuse sidecar. This check is currently performed in CSI node driver but that consumes additional STS quota hence it is being moved to sidecar. The STS quota benefits can be seen only for Workload Identity pods while HostNetwork pods will see ad additional STS quota request per pod.
Since the bucket access check is being performed in sidecar with this PR it can be skipped in node by setting `--skipCSIBucketAccessCheck=true`

***Workload Identity***
1. Performs bucket access check in sidecar
2. STS quota savings as WI serves any recurring request from same pod through its cached token.
3. Ensure `--skipCSIBucketAccessCheck=true` to save STS quota

***Host Network***
1. Performs bucket access check in sidecar
2. Sidecar makes an additional STS request to the token server to perform bucket access check.

***Metrics to view STS requests and quota***
STS quota consumption can be found in Metrics explorer page `service=sts.googleapis.com` and metric `Consumed API - Request Count`
image

***Testing***
Ran full test suite locally, some of the failing tests succeeded on re-run with E2E_TEST_FOCUS, the tests were failing in earlier attempt due to unrelated issues (tag naming - needed `prow-gob-internal-boskos` in tag name)
```
============================= gcsfuseIntegrationTests focused =============================

[ReportAfterSuite] PASSED [0.081 seconds]
autogenerated by Ginkgo

Ran 76 of 434 Specs in 2376.885 seconds
SUCCESS! -- 76 Passed | 0 Failed | 0 Pending | 358 Skipped


==================== Volumes focused ===============
make e2e-test E2E_TEST_FOCUS="volumes.should.*" E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false STAGINGVERSION=$STAGINGVERSION  REGISTRY=$REGISTRY
PROJECT is snehaaradhey-gke-dev
OVERLAY is stable
STAGINGVERSION is test-bucket-access-check-debugging-prow-gob-internal-boskos-29
DRIVER_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver
SIDECAR_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver-sidecar-mounter
WEBHOOK_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver-webhook
........


Ran 12 of 434 Specs in 32.093 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 422 Skipped

===================== Mount focused ================
OVERLAY is stable
STAGINGVERSION is test-bucket-access-check-debugging-prow-gob-internal-boskos-29
DRIVER_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver
SIDECAR_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver-sidecar-mounter
WEBHOOK_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver-webhook

.......
Ran 7 of 434 Specs in 19.492 seconds
SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 427 Skipped


========================== All the failing tests were re-run with E2E_TEST_FOCUS captured above, the tests were failing due to unrelated issues (tag naming)  ====================
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false STAGINGVERSION=$STAGINGVERSION  REGISTRY=$REGISTRY
PROJECT is snehaaradhey-gke-dev
OVERLAY is stable
STAGINGVERSION is test-bucket-access-check-debugging-29
DRIVER_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver
SIDECAR_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver-sidecar-mounter
WEBHOOK_IMAGE is us-central1-docker.pkg.dev/snehaaradhey-gke-dev/csi-dev/gcs-fuse-csi-driver-webhook
.......
Ran 324 of 434 Specs in 8333.466 seconds
FAIL! -- 307 Passed | 17 Failed | 8 Flaked | 0 Pending | 110 Skipped
```

Fixes #

```release-note
Enhances sidecar to perform bucket access check before mounting any bucket
```

### DIFF
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -40,28 +40,30 @@ import (
 )
 
 var (
-	endpoint                     = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
-	nodeID                       = flag.String("nodeid", "", "node id")
-	runController                = flag.Bool("controller", false, "run controller service")
-	runNode                      = flag.Bool("node", false, "run node service")
-	kubeconfigPath               = flag.String("kubeconfig-path", "", "The kubeconfig path.")
-	kubeAPIQPS                   = flag.Float64("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
-	kubeAPIBurst                 = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
-	identityPool                 = flag.String("identity-pool", "", "The Identity Pool to authenticate with GCS API.")
-	identityProvider             = flag.String("identity-provider", "", "The Identity Provider to authenticate with GCS API.")
-	enableProfiling              = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
-	enableBucketScanner          = flag.Bool("enable-bucket-scanner", false, "enable the bucket scanner feature")
-	datafluxParallelism          = flag.Int("dataflux-parallelism", 0, "number of go routines for Dataflux lister. Defaults to 0 (10X number of available vCPUs)")
-	datafluxBatchSize            = flag.Int("dataflux-batch-size", 25000, "batch size for Dataflux lister. Defaults to 25000")
-	datafluxSkipDirectoryObjects = flag.Bool("dataflux-skip-directory-objects", false, "set to true to skip Dataflux listing objects that include files with names ending in '/'")
-	informerResyncDurationSec    = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
-	retryIntervalStart           = flag.Duration("retry-interval-start", time.Second, "Initial retry interval for a failed PV processing operation. It doubles with each failure, up to retry-interval-max.")
-	retryIntervalMax             = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval for a failed PV processing operation.")
-	fuseSocketDir                = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
-	metricsEndpoint              = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
-	maximumNumberOfCollectors    = flag.Int("max-metric-collectors", -1, "Maximum number of prometheus metric collectors exporting metrics at a time, less than 0 (e.g -1) means no limit.")
-	disableAutoconfig            = flag.Bool("disable-autoconfig", false, "Disable gcsfuse's defaulting based on machine type")
-	wiNodeLabelCheck             = flag.Bool("wi-node-label-check", true, "Workload Identity node label check")
+	endpoint                       = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	nodeID                         = flag.String("nodeid", "", "node id")
+	runController                  = flag.Bool("controller", false, "run controller service")
+	runNode                        = flag.Bool("node", false, "run node service")
+	kubeconfigPath                 = flag.String("kubeconfig-path", "", "The kubeconfig path.")
+	kubeAPIQPS                     = flag.Float64("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
+	kubeAPIBurst                   = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
+	identityPool                   = flag.String("identity-pool", "", "The Identity Pool to authenticate with GCS API.")
+	identityProvider               = flag.String("identity-provider", "", "The Identity Provider to authenticate with GCS API.")
+	enableProfiling                = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
+	enableBucketScanner            = flag.Bool("enable-bucket-scanner", false, "enable the bucket scanner feature")
+	datafluxParallelism            = flag.Int("dataflux-parallelism", 0, "number of go routines for Dataflux lister. Defaults to 0 (10X number of available vCPUs)")
+	datafluxBatchSize              = flag.Int("dataflux-batch-size", 25000, "batch size for Dataflux lister. Defaults to 25000")
+	datafluxSkipDirectoryObjects   = flag.Bool("dataflux-skip-directory-objects", false, "set to true to skip Dataflux listing objects that include files with names ending in '/'")
+	informerResyncDurationSec      = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
+	retryIntervalStart             = flag.Duration("retry-interval-start", time.Second, "Initial retry interval for a failed PV processing operation. It doubles with each failure, up to retry-interval-max.")
+	retryIntervalMax               = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval for a failed PV processing operation.")
+	fuseSocketDir                  = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
+	metricsEndpoint                = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
+	maximumNumberOfCollectors      = flag.Int("max-metric-collectors", -1, "Maximum number of prometheus metric collectors exporting metrics at a time, less than 0 (e.g -1) means no limit.")
+	disableAutoconfig              = flag.Bool("disable-autoconfig", false, "Disable gcsfuse's defaulting based on machine type")
+	wiNodeLabelCheck               = flag.Bool("wi-node-label-check", true, "Workload Identity node label check")
+	enableSidecarBucketAccessCheck = flag.Bool("enable-sidecar-bucket-access-check", false, "Enable bucket access check on sidecar, this does not disable bucket access check in node driver")
+
 	enableCloudProfilerForDriver = flag.Bool("enable-cloud-profiler-for-driver", false, "Enable cloud profiler to collect analysis data")
 	// These are set at compile time.
 	version = "unknown"
@@ -163,19 +165,20 @@ func main() {
 	}
 
 	config := &driver.GCSDriverConfig{
-		Name:                  driver.DefaultName,
-		Version:               version,
-		NodeID:                *nodeID,
-		RunController:         *runController,
-		RunNode:               *runNode,
-		StorageServiceManager: ssm,
-		TokenManager:          tm,
-		Mounter:               mounter,
-		K8sClients:            clientset,
-		MetricsManager:        mm,
-		DisableAutoconfig:     *disableAutoconfig,
-		WINodeLabelCheck:      *wiNodeLabelCheck,
-		FeatureOptions:        featureOptions,
+		Name:                           driver.DefaultName,
+		Version:                        version,
+		NodeID:                         *nodeID,
+		RunController:                  *runController,
+		RunNode:                        *runNode,
+		StorageServiceManager:          ssm,
+		TokenManager:                   tm,
+		Mounter:                        mounter,
+		K8sClients:                     clientset,
+		MetricsManager:                 mm,
+		DisableAutoconfig:              *disableAutoconfig,
+		WINodeLabelCheck:               *wiNodeLabelCheck,
+		EnableSidecarBucketAccessCheck: *enableSidecarBucketAccessCheck,
+		FeatureOptions:                 featureOptions,
 	}
 
 	gcfsDriver, err := driver.NewGCSDriver(config)
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -29,6 +30,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/profiler"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
 	driver "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/csi_driver"
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
@@ -36,9 +38,17 @@ import (
 )
 
 var (
-	gcsfusePath    = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
-	volumeBasePath = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
-	_              = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
+	gcsfusePath                           = flag.String("gcsfuse-path", "/gcsfuse", "gcsfuse path")
+	volumeBasePath                        = flag.String("volume-base-path", webhook.SidecarContainerTmpVolumeMountPath+"/.volumes", "volume base path")
+	_                                     = flag.Int("grace-period", 0, "grace period for gcsfuse termination. This flag has been deprecated, has no effect and will be removed in the future.")
+	kubeconfigPath                        = flag.String("kubeconfig-path", "", "The kubeconfig path.")
+	informerResyncDurationSec             = flag.Int("informer-resync-duration-sec", 1800, "informer resync duration in seconds")
+	storageServiceAndBucketAccessCap      = flag.Duration("storage-service-check-retry-cap", 60*time.Minute, "storage service creation and bucket access check exponential retry cap")
+	storageServiceAndBucketAccessSteps    = flag.Int("storage-service-check-retry-steps", math.MaxInt, "storage service creation and bucket access check exponential retry cap") // Effectively infinite retry steps
+	storageServiceAndBucketAccessFactor   = flag.Float64("storage-service-check-retry-factor", 2.0, "storage service creation and bucket access check exponential retry factor")
+	storageServiceAndBucketAccessJitter   = flag.Float64("storage-service-check-retry-jitter", 0.1, "storage service creation and bucket access check exponential retry jitter")
+	storageServiceAndBucketAccessDuration = flag.Duration("storage-service-check-retry-duration", 5*time.Second, "storage service creation and bucket access check exponential retry initial duration")
+
 	// This is set at compile time.
 	version = "unknown"
 )
@@ -55,9 +65,12 @@ func main() {
 		klog.Fatalf("failed to look up socket paths: %v", err)
 	}
 
+	clientset, err := clientset.New(*kubeconfigPath, *informerResyncDurationSec)
+	if err != nil {
+		klog.Fatalf("Failed to configure k8s client: %v", err)
+	}
 	mounter := sidecarmounter.New(*gcsfusePath)
 	ctx, cancel := context.WithCancel(context.Background())
-
 	flagsFromDriver := map[string]string{}
 	defaultingFlagFilePath := *volumeBasePath + "/" + driver.FlagFileForDefaultingPath
 	klog.Infof("Checking if defaulting-flag file %q exists", defaultingFlagFilePath)
@@ -69,7 +82,6 @@ func main() {
 		fileContent := string(machineTypeBytes)
 		flagsFromDriver = driver.ParseFlagMapFromFlagFile(fileContent)
 	}
-
 	for _, sp := range socketPaths {
 		klog.V(4).Infof("in sidecar mounter, found socket path %s", sp)
 		// sleep 1.5 seconds before launch the next gcsfuse to avoid
@@ -88,6 +100,19 @@ func main() {
 			}
 		}
 		if mc != nil {
+			if mc.EnableSidecarBucketAccessCheck {
+				tm, ssm, err := mounter.SetupTokenAndStorageManager(clientset, mc)
+				if err != nil {
+					klog.Errorf("Failed to fetch identity pool and identity provider details required for bucket access check, got error %v", err)
+				}
+				mounter.TokenManager = tm
+				mounter.StorageServiceManager = ssm
+				mc.SidecarRetryConfig.Cap = *storageServiceAndBucketAccessCap
+				mc.SidecarRetryConfig.Steps = *storageServiceAndBucketAccessSteps
+				mc.SidecarRetryConfig.Factor = *storageServiceAndBucketAccessFactor
+				mc.SidecarRetryConfig.Duration = *storageServiceAndBucketAccessDuration
+				mc.SidecarRetryConfig.Jitter = *storageServiceAndBucketAccessJitter
+			}
 			if err := mounter.Mount(ctx, mc); err != nil {
 				mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to mount bucket %q for volume %q: %v\n", mc.BucketName, mc.VolumeName, err))
 			}
--- a/pkg/cloud_provider/auth/token_sources.go
+++ b/pkg/cloud_provider/auth/token_sources.go
@@ -29,6 +29,8 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/metadata"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 	"golang.org/x/oauth2"
 	"google.golang.org/api/option"
 	sts "google.golang.org/api/sts/v1"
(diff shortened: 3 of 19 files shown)


Sneha-at:
Ran E2E tests successfully
```

[ReportAfterSuite] PASSED [332.561 seconds]
autogenerated by Ginkgo

Ran 76 of 432 Specs in 2471.130 seconds
SUCCESS! -- 76 Passed | 0 Failed | 0 Pending | 356 Skipped
```

amacaskill:
Please add PR description for everything this PR does, Add release note, and tag the PR with PR type

cmd/csi_driver/main.go
```diff
@@ -40,25 +40,27 @@ import (
...
+	retryIntervalMax                   = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval for a failed PV processing operation.")
+	fuseSocketDir                      = flag.String("fuse-socket-dir", "/sockets", "FUSE socket directory")
+	metricsEndpoint                    = flag.String("metrics-endpoint", "", "The TCP network address where the Prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means that the metrics endpoint is disabled.")
+	maximumNumberOfCollectors          = flag.Int("max-metric-collectors", -1, "Maximum number of prometheus metric collectors exporting metrics at a time, less than 0 (e.g -1) means no limit.")
+	disableAutoconfig                  = flag.Bool("disable-autoconfig", false, "Disable gcsfuse's defaulting based on machine type")
+	wiNodeLabelCheck                   = flag.Bool("wi-node-label-check", true, "Workload Identity node label check")
+	enableSidecarBucketAccessCheckFlag = flag.Bool("enable-sidecar-bucket-access-check-flag", false, "Enable bucket access check on sidecar, this does not disable bucket access check in node driver")
```
saikat-royc: enable-sidecar-bucket-access-check

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -38,51 +40,60 @@ import (
...
+	var err error
+	if !ts.fetchTokenfromFile {
+		k8sSAToken, err = ts.fetchK8sSAToken(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("k8s service account token fetch error: %w", err)
+		}
+	} else {
+		k8sSAToken, err = util.FetchK8sTokenFromFile(webhook.SidecarContainerSATokenVolumeMountPath + "/" + webhook.K8STokenPath)
+		if err != nil {
+			return nil, fmt.Errorf("k8s service account token fetch error: %w", err)
```
saikat-royc, line 68: please log the file path where the token is being fetched from

pkg/webhook/sidecar_spec.go
```diff
@@ -104,7 +104,7 @@ func GetSidecarContainerSpec(c *Config) corev1.Container {
 	limits, requests := prepareResourceList(c)
 
 	volumeMounts := []corev1.VolumeMount{TmpVolumeMount, buffVolumeMount, cacheVolumeMount}
-	if c.PodHostNetworkSetting && c.ShouldInjectSAVolume {
+	if c.ShouldInjectSAVolume {
```
saikat-royc, line 126: Double check the existing tests, if there is some strict checks for existence of this token volume only for hostnetwork

pkg/csi_driver/node.go
```diff
@@ -88,7 +94,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	// Validate arguments
-	targetPath, bucketName, userSpecifiedIdentityProvider, fuseMountOptions, skipBucketAccessCheck, disableMetricsCollection, optInHostnetworkKSA, enableCloudProfilerForSidecar, err := parseRequestArguments(req)
+	targetPath, bucketName, userSpecifiedIdentityProvider, identityPool, fuseMountOptions, skipBucketAccessCheck, disableMetricsCollection, optInHostnetworkKSA, enableCloudProfilerForSidecar, err := parseRequestArguments(req)
```
saikat-royc: userSpecifiedIdentityPool


// so that we know it comes from the volume attributes

pkg/csi_driver/utils.go
```diff
@@ -277,33 +280,37 @@ func parseVolumeAttributes(fuseMountOptions []string, volumeContext map[string]s
 
 				mountOptionWithValue = mountOption + strconv.Itoa(intVal)
 			} else {
-				return nil, userSpecifiedIdentityProvider, skipCSIBucketAccessCheck, disableMetricsCollection, optInHostnetworkKSA, enableCloudProfilerForSidecar, fmt.Errorf("volume attribute %v only accepts a valid int value, got %q", volumeAttribute, value)
+				return nil, userSpecifiedIdentityProvider, identityPool, skipCSIBucketAccessCheck, disableMetricsCollection, optInHostnetworkKSA, enableCloudProfilerForSidecar, fmt.Errorf("volume attribute %v only accepts a valid int value, got %q", volumeAttribute, value)
 			}
 		case VolumeContextKeyIdentityProvider:
 			userSpecifiedIdentityProvider = value
+		case VolumeContextKeyIdentityPool:
```
saikat-royc: is there a strict requirement to introduce this volume attribute? is it to handle the custom image scenario for the bucket checks (identity pool used by SetupTokenAndStorageManager). If yes, we need to update any external facing docs

pkg/csi_driver/node.go
```diff
@@ -138,16 +144,30 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+	}
+
+	if s.driver.config.EnableSidecarBucketAccessCheckFlag {
+		if identityProvider == "" {
+			identityProvider = s.driver.config.TokenManager.GetIdentityProvider()
+			fuseMountOptions = joinMountOptions(fuseMountOptions, []string{TokenServerIdentityProviderConst + "=" + identityProvider})
+		}
+		klog.Infof("Got identity provider %s", identityProvider)
+
+		identityPool = s.driver.config.TokenManager.GetIdentityPool()
```
saikat-royc: in L97 of node.go we are parsing the pool, but here we are unconditionally overriding the default pool.

not sure I follow the logic?

pkg/csi_mounter/csi_mounter.go
```diff
@@ -162,6 +162,7 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 		klog.V(4).Infof("%v closing the socket and fd", logPrefix)
 		listener.Close()
 		syscall.Close(fd)
+
```
saikat-royc: remove new line

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -64,14 +75,36 @@ func New(mounterPath string) *Mounter {
 func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	// Start the token server for HostNetwork enabled pods.
 	// For managed sidecar, the token server identity provider is only populated when host network pod ksa feature is opted in.
+	var tokenSource oauth2.TokenSource
+	var audience string
+	var err error
+	if mc.EnableSidecarBucketAccessCheckFlag {
+		if mc.HostNetworkKSAOptIn {
```
saikat-royc: what about the case where HostNetworkKSAOptIn is false?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -309,64 +350,7 @@ func scrapeMetrics(ctx context.Context, metricEndpoint string, w http.ResponseWr
...
-		projectID,
-		identityProvider,
-	)
-	klog.Infof("audience: %s", audience)
-
-	return audience, nil
-}
-
-func StartTokenServer(ctx context.Context, tokenURLBasePath, tokenSocketName, identityProvider string) {
+func StartTokenServer(ctx context.Context, tokenURLBasePath, tokenSocketName string, tm auth.TokenManager, tokenSource oauth2.TokenSource, identityProvider string) {
```
saikat-royc: where is the tokenSource used in this function StartTokenServer

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -427,3 +413,97 @@ func StartTokenServer(ctx context.Context, tokenURLBasePath, tokenSocketName, id
 		klog.Errorf("Server for %q returns unexpected error: %v", tokenURLSocketPath, err)
 	}
 }
+
+// checkBucketAccessWithRetry prepares the GCS Storage Service using the Kubernetes Service Account from VolumeContext and validates bucket access.
+func (m *Mounter) checkBucketAccessWithRetry(ctx context.Context, storageServiceManager storage.ServiceManager, tokenSource oauth2.TokenSource, tm auth.TokenManager, bucketName string, tokenProvider string) error {
+	backoff := wait.Backoff{
+		Duration: 5 * time.Second,
```
saikat-royc: make this some constant (preferably flags in sidecar with default values), so we have a fallback to tweak them

pkg/sidecar_mounter/sidecar_mounter.go
saikat-royc: I think i understand the logic now , for the non host network case, we want to use an empty token source, so that we use the default credentials to the token exchange. is this correct? If yes, please put some comments, as it is difficult to navigate the code path.


Also I think we can flip the if checks. we already have a if mc. HostNetworkKSAOptIn {} in L93. we can put the audience and tokensource inside that block. But i still dont understand why we gate the audence and tokensource behind the EnableSidecarBucketAccessCheckFlag for the host network code path?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -427,3 +413,97 @@ func StartTokenServer(ctx context.Context, tokenURLBasePath, tokenSocketName, id
...
+		Factor:   2.0,
+		Cap:      60 * time.Minute,
+		Steps:    math.MaxInt32, // Effectively infinite retry steps
+		Jitter:   0.1,           // Adds randomness, this will give +/- 10% of the current delay
+	}
+
+	storageServiceCreationFunc := func(ctx context.Context) (bool, error) {
```
saikat-royc: ssCreateAndBucketCheckFunc -> to reflect what the func actually does

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -36,21 +37,31 @@ import (
 
 	"cloud.google.com/go/compute/metadata"
 	credentials "cloud.google.com/go/iam/credentials/apiv1"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/auth"
+	csimetadata "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/metadata"
```
saikat-royc: csimetadata is not the right name. cpmeta maybe?

pkg/sidecar_mounter/sidecar_mounter.go
saikat-royc: where is the tm, and tokenSource used in this function StartTokenServer?

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -38,51 +40,60 @@ import (
...
+	k8sSAName          string
+	k8sSANamespace     string
+	k8sSAToken         string
+	k8sClients         clientset.Interface
+	audience           string
+	fetchTokenfromFile bool
 }
 
 // Token exchanges a GCP IAM SA Token with a Kubernetes Service Account token.
 func (ts *GCPTokenSource) Token() (*oauth2.Token, error) {
```
amacaskill, line 53: Please add comment detailing the flow for how the STS token exchange differs based on ts.fetchTokenFromFile,  when fetchTokenFromFile is supposed to be set, and why the identity binding token itself is used to authenticate to GCS instead of the GCP SA token when fetchTokenFromFile is true...  The type of token (identity binding token, vs GCP SA token) that is returned from this function differs based on the value of  fetchTokenFromFile, so this is important.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -64,14 +75,36 @@ func New(mounterPath string) *Mounter {
...
+	var tokenSource oauth2.TokenSource
+	var audience string
+	var err error
+	if mc.EnableSidecarBucketAccessCheckFlag {
+		if mc.HostNetworkKSAOptIn {
+			audience, err = getAudienceFromContextAndIdentityProvider(ctx, mc.TokenServerIdentityProvider)
+			if err != nil {
+				return fmt.Errorf("failed to get audience from the context: %w", err)
+			}
+			err, tokenSource = m.fetchTokenSource(mc.PodNamespace, mc.ServiceAccountName, audience)
```
amacaskill: nit: swap tokenSource and err returns

pkg/cloud_provider/storage/storage.go
```diff
@@ -384,3 +385,21 @@ func isBucketAZonalBucket(ctx context.Context, client *storage.Client, bucketNam
...
+func (manager *gcsServiceManager) SetupStorageServiceForSidecar(ctx context.Context, ts oauth2.TokenSource) (Service, error) {
+	var err error
+	var storageClient *storage.Client
+
+	if ts != nil {
+		client := oauth2.NewClient(ctx, ts)
+		storageClient, err = storage.NewClient(ctx, option.WithHTTPClient(client))
+	} else {
+		storageClient, err = storage.NewClient(ctx)
+	}
```
amacaskill, line 400: The way we gate logic between the two flows is hard to follow. Can we have one flag that does the new flow vs old flow, or more comments everywhere?  Doing checks like  X resource =="" or X resource != nil ,where X resource keeps changing, it makes it hard to follow, especially when there isn't comments letting you know when X resource should be nil vs not.

pkg/csi_driver/node.go
Sneha-at: Token managers metadata service is created using the passed Identity Pool value and it doesn't change even for Hostnetwork setting. So s.driver.config.TokenManager.GetIdentityPool() is the same value just different approach for fetching.
```
meta, err := csimetadata.NewMetadataService(mc.TokenServerIdentityPool, mc.TokenServerIdentityProvider)
```

pkg/csi_driver/node.go
Sneha-at: In this feature I am only passing the identity pool from node to sidecar, this does not add support for custom identity pool, it might be misleading to name it as userSpecified? hence the deviance from userSpecifiedIdentityProvider naming convention.

pkg/csi_driver/utils.go
Sneha-at: Yeah, we need identity pool to setup the Token Manager in sidecar

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: > for the non host network case, we want to use an empty token source, so that we use the default credentials to the token exchange. is this correct?
That's correct, adding some comments


>But i still don't understand why we gate the audience and tokensource behind the EnableSidecarBucketAccessCheckFlag for the host network code path? 

The flag is to just disable all the related changes in case on any issues. We don't intend to share this flag for individual customer use case right? I have gated all the code changes behind this flag.

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: Thanks! this was stale code.

pkg/csi_driver/node.go
Sneha-at: I missed this is for node and not sidecar, you are right we can just fetch it directly. No need to parse it through Volume Attributes. I can modify this as we will have a separate PR to support custom identity pool

pkg/cloud_provider/storage/storage.go
Sneha-at: Added some comments, let me know if this helps clarify

pkg/cloud_provider/auth/token_sources.go
Sneha-at: Added some comments

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: That was stale code, thanks for catching.

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: Updated to cpmetadata

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: Added flags for 2 parameters I thought we would want to change in future - Steps and Cap. Let me know if we need flags for any other params too.

pkg/csi_driver/node.go
Sneha-at: Removed passing identity pool through volume attributes, it is not required for this feature

cmd/csi_driver/main.go
Sneha-at: Done

pkg/cloud_provider/auth/token_sources.go
Sneha-at: Done

pkg/webhook/sidecar_spec.go
Sneha-at: I ran our existing tests, all are passing. Our managed component uses GCW so will check guitar tests with GCW webhook update

pkg/sidecar_mounter/sidecar_mounter.go
amacaskill: Since all flags work together, I think we should add all flags not just 2. If we need to change something about this, it very well could be the factor / initial duration. I would just add flags for everything, and just make it clear what the flags are for. You can add a comment section saying these are all for the sidecar retry

pkg/webhook/sidecar_spec.go
amacaskill: Can you also just confirm with Siyan why we added this check in the first place? Maybe she knows of an edge case we are missing?

test/e2e/specs/specs.go
```diff
@@ -159,7 +159,7 @@ func NewTestPod(c clientset.Interface, ns *corev1.Namespace) *TestPod {
 				},
 				RestartPolicy:                corev1.RestartPolicyAlways,
 				Volumes:                      make([]corev1.Volume, 0),
-				AutomountServiceAccountToken: ptr.To(false),
+				AutomountServiceAccountToken: ptr.To(true),
```
amacaskill, line 162: WHat about current implementation requires this, and this is run for managed and non managed, so what impact does changing this have on tests run on versions before your change?

cmd/sidecar_mounter/main.go
```diff
@@ -88,7 +96,15 @@ func main() {
 			}
 		}
 		if mc != nil {
-			if err := mounter.Mount(ctx, mc); err != nil {
+			if mc.EnableSidecarBucketAccessCheck {
+				tm, ssm, err := mounter.SetupTokenAndStorageManager(clientset, mc)
+				if err != nil {
+					klog.Errorf("Failed to fetch identity pool and identity provider details required for bucket access check, got error %q", err)
```
amacaskill: got error %q", err) -> got error %v", err)

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -38,51 +40,65 @@ import (
...
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
 	defer cancel()
-
-	k8sSAToken, err := ts.fetchK8sSAToken(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("k8s service account token fetch error: %w", err)
+	var k8sSAToken string
+	var err error
+	// If fetchTokenfromFile is set we assume the request is coming from sidecar, as the sidecar (user pod) doesn't have access to get service account details we fetch the service account from file.
```
amacaskill, line 58: The comment describing case when fetchTokenfromFile is set is above the if block which handles where fetchTokenfromFile is false, Can we swap if statement to reflect the coment? 

like 
```
if ts.fetchTokenfromFile {` 
// 		tokenPath := webhook.SidecarContainerSATokenVolumeMountPath + "/" + webhook.K8STokenPath
} else {
		k8sSAToken, err = ts.fetchK8sSAToken(ctx)
```

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -38,51 +40,65 @@ import (
...
+	identityBindingToken, err := ts.FetchIdentityBindingToken(ctx, k8sSAToken, ts.audience)
 	if err != nil {
 		return nil, fmt.Errorf("identity binding token fetch error: %w", err)
 	}
-
-	token, err := ts.fetchGCPSAToken(ctx, identityBindingToken)
-	if err != nil {
-		return nil, fmt.Errorf("GCP service account token fetch error: %w", err)
+	// GCP SA Token was needed in pre WI world where we linked GCP SA Token and K8s SA token through annotation.
+	// In the fetchGCPSAToken too we check if there's any annotation on tha SA and return IdentityBindingToken if no annotation specified.
```
amacaskill: // In the fetchGCPSAToken too we check if there's any annotation on tha SA and return IdentityBindingToken if no annotation specified. -> 

// In the fetchGCPSAToken we check if there's any annotation on the SA and return IdentityBindingToken if no annotation specified.

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -38,51 +40,65 @@ import (
...
+	// If however fetchTokenfromFile is set we assume the request is coming from sidecar, as it doesn't have access to Get SA API we skip the check assuming the sidecar follows new setup.
```
amacaskill, line 78: Same comment as above comment is directly above the opposite of the case it describes. Can we change this to: 

```
// If however fetchTokenfromFile is set we assume the request is coming from sidecar, as it doesn't have access to Get SA API we skip the check assuming the sidecar follows new setup.	The new setup is the default with WI

if ts.fetchTokenfromFile {
   return identityBindingToken
} 
token, err := ts.fetchGCPSAToken(ctx, identityBindingToken)
if err != nil {
	return nil, fmt.Errorf("GCP service account token fetch error: %w", err)
}
return token, nil

```

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -61,16 +71,37 @@ func New(mounterPath string) *Mounter {
 	}
 }
 
-func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
+func (m *Mounter) Mount(ctx context.Context, mc *MountConfig, storageServiceAndBucketAccessCap time.Duration, storageServiceAndBucketAccessSteps int) error {
```
amacaskill: We can probably just add all of the storageServiceAndBucketAccessCap, steps, factor, etc into a type within MountConfig.

siyanshen:
If there is no STS quota benefit for Host Network pods at all, can we keep the bucket check in the node for host network pods?

pkg/webhook/sidecar_spec.go
siyanshen: I added this so that OSS webhook only injects SA volume for host network pods. It was one of the switches for the token server. We have many other switches as defined in this function: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/817d1d66839b5c9aad8988391e354217f790526c/pkg/csi_driver/node.go#L342

As long as hostnetwork related tests are passing, I am fine with this.

pkg/csi_driver/utils.go
Sneha-at: Removed this, we don't need to pass the user provided identity pool

pkg/cloud_provider/auth/token_sources.go
Sneha-at: Done

pkg/cloud_provider/auth/token_sources.go
Sneha-at: Done

pkg/cloud_provider/auth/token_sources.go
Sneha-at: Done

cmd/sidecar_mounter/main.go
Sneha-at: Done

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: Done

pkg/webhook/sidecar_spec.go
Sneha-at: Thanks!, closing this as no further comments

pkg/sidecar_mounter/sidecar_mounter.go
Sneha-at: Done

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -237,16 +249,35 @@ func (mc *MountConfig) prepareMountArgs() {
 			value = argPair[1]
 		}
 
-		if flag == identityProviderFlag {
```
amacaskill, line 240: all these if flag == is probably simpler as a switch statement

amacaskill:
LGTM besides the open comments I still have. I'm waiting on this PR to merge my custom audience PR since I will need to make additional changes, so the sooner we can merge this, the better :)

test/e2e/specs/specs.go
Sneha-at: Currently for our tests SA token is mounted in sidecar only for Host network workloads. This updates to mount SA token by default. For tests running on older versions too there shouldn't be any issue due to this mount, this is the default behavior for a pod.

siyanshen:
> If there is no STS quota benefit for Host Network pods at all, can we keep the bucket check in the node for host network pods?

Synced with Sneha offline. We will enable sidecar bucket check for host network pods for now for code simplicity, but a scale testing is needed to make sure no significant amount of STS quota is used up because of this feature.

pkg/cloud_provider/storage/storage.go
```diff
@@ -384,3 +385,24 @@ func isBucketAZonalBucket(ctx context.Context, client *storage.Client, bucketNam
 	}
 	return attrs.StorageClass == "RAPID", nil
 }
+
+func (manager *gcsServiceManager) SetupStorageServiceForSidecar(ctx context.Context, ts oauth2.TokenSource) (Service, error) {
+	var err error
+	var storageClient *storage.Client
+	// For workload identity enabled resources we need to create the storage service with default credentials so as to not consume more STS quota.
+	// The token source thus is only shared for host network enabled workload. If token source is nil then create storage client with default credentials else use the tokenSource.
+	// This is needed as the storage API checks calls TokenSource.Token() function (defined above) which leads to increased STS quota since we are directly hittingthe STS API.
```
siyanshen, line 394: nit: hitting the

pkg/sidecar_mounter/sidecar_mounter_config.go
Sneha-at: Done

pkg/csi_driver/node.go
Sneha-at: This is no longer relevant since we are not passing identitypool through volume attributes

siyanshen (approved):

PR #629, opened 2025-05-28
Revert "Fix dependabot bumps in root directory + in /test"

hungnguyen243:
Reverts GoogleCloudPlatform/gcs-fuse-csi-driver#609, since test directory is [failing build](https://screenshot.googleplex.com/cxJsmg8W27mxh75)


hime (approved):

hime:
Are we able to catch this locally, or do we have to wait for it to run on prow?

hungnguyen243:
We can try to run e2e tests locally probably. It just completely slipped my mind that a change like this could break e2e tests...

PR #657, opened 2025-05-29
Add support for node restarts.

hime:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR allows the GCSFuse CSI Driver to handle Node restarts. We are currently seeing more demand for this feature as the restarts become increasingly common the scale of the cluster increases. Certain nodes (e.g. TPU and GPU VMs) perform restarts for more scenarios compared to a traditional GKE node/nodepool (e.g. host error, maintenance preemption, etc.)

**Restart scenarios tested**:
We've manually verified select features of the GCSFuse CSI Driver against node restarts. 
Each feature was verified against both graceful and ungraceful node restart. The testing steps are as follows:
1. Deploy workload on GKE Cluster
2. Verify feature works by: running IOPS against volume, check logs for errors
3. Perform node restart
5. Verify feature works by: running IOPS against volume, check logs for errors

Graceful node restart command
`sudo systemctl reboot`

Ungraceful node restart command (node crash)
`sudo su \ echo 1 > /proc/sys/kernel/sysrq & echo "c" > /proc/sysrq-trigger`

**Verified the following features:**
- Defaulting flags for GCSFuse
- HostNetworking
- Metrics (only addressing metrics endpoint health)

**Breakdown of PR**
Stale sockets that are cleaned up:
- Mount
  - Socket used to transfer the /dev/fuse file descriptor, bucket name, and GCSFuse configurations from csi driver to sidecar mounter
  - Stale socket is reached when the Node crashes after the NodePublishVolume call succeeds, but before the sidecar starts up.
  - image
- Hostnetwork
  - Socket used by token server to process requests from GCSFuse
  - image

- Metrics
  - Socket used by metrics process running in sidecar mounter to scrape metrics from GCSFuse
  - image

Files used by GCSFuse CSI Driver:
- `config.yaml`
  - This file is created by sidecar and it is read by GCSFuse process.
  - cleaned up to avoid permission issues while reading file after restart.
- `error`
  - created by gcsfuse process when failure occurs
  - cleaned up to avoid storing stale errors after node restart
- `flags-for-defaulting`
  - created by csi driver and used by sidecar mounter at startup to create `config.yaml`
  - Did not encounter permission or startup issues by having stale file after restart.
- `exit`
  - Only used on clusters that don't support native sidecar
  - This file is placed on sidecar tmp volume once all workload containers terminate.
  - No action required. Once exit file is placed, pod is not expected to restart. Kubelet does not restart pods that are in terminating state in the event of a restart.

**Local tests:**
image

Fixes #

This PR also adds retries to the WaitForLog function used in e2e testing, as too many tests calling this function can cause temporary failures.

```release-note
Add node restart support to GCSFuse CSI driver.
```

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -437,6 +437,13 @@ See GKE documentation [Storage for GKE clusters overview](https://cloud.google.c
 
 If the sidecar container (`gke-gcsfuse-sidecar`) hosting the gcsfuse process, crashes for some reason (e.g OOM Kill, or fatals), and kubelet decides to inplace restart the container, gcsfusecsi driver does not handle this scenario gracefully. This is because in the CSI NodePublishVolume the driver [checks the last known state](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/v1.15.4/pkg/csi_driver/utils.go#L431) of the container and fails fast. To mitigate this, delete and create the pod.
 
+### Node level restarts
+
+GCSFuse CSI Driver can now manage node restarts without impacting pods using GCSFuse-backed volumes. The driver ensures a clean slate by clearing stale files or sockets, and re-establishing the gcsfuse /dev/fuse file descriptor during the NodePublishVolume phase after a restart.
+
+Known issues after a node restart:
+- Metrics will not be visible on Cloud Monitoring after node restart. If they become visible, assume that they will be malformed.
+- Metrics after a restart will only look at the time period after the node restart.
 
 ### enabling debug logs
 
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -197,33 +197,34 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 	}
 
-	// Check if there is any error from the gcsfuse
-	code, err := checkGcsFuseErr(isInitContainer, pod, targetPath)
-	if code != codes.OK {
-		if code == codes.Canceled {
-			klog.V(4).Infof("NodePublishVolume on volume %q to target path %q is not needed because the gcsfuse has terminated.", bucketName, targetPath)
-
-			return &csi.NodePublishVolumeResponse{}, nil
-		}
-
-		return nil, status.Error(code, err.Error())
-	}
-
-	// Check if there is any error from the sidecar container
-	code, err = checkSidecarContainerErr(isInitContainer, pod)
-	if code != codes.OK {
-		return nil, status.Error(code, err.Error())
-	}
-
-	// TODO: Check if the socket listener timed out
-
 	// Check if the target path is already mounted
 	mounted, err := s.isDirMounted(targetPath)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to check if path %q is already mounted: %v", targetPath, err)
 	}
 
+	// Skip mount if the mount on target path already exists.
 	if mounted {
+		// Check if there is any error from the gcsfuse
+		code, err := checkGcsFuseErr(isInitContainer, pod, targetPath)
+		if code != codes.OK {
+			if code == codes.Canceled {
+				klog.V(4).Infof("NodePublishVolume on volume %q to target path %q is not needed because the gcsfuse has terminated.", bucketName, targetPath)
+
+				return &csi.NodePublishVolumeResponse{}, nil
+			}
+
+			return nil, status.Error(code, err.Error())
+		}
+
+		// Check if there is any error from the sidecar container
+		code, err = checkSidecarContainerErr(isInitContainer, pod)
+		if code != codes.OK {
+			return nil, status.Error(code, err.Error())
+		}
+
+		// TODO: Check if the socket listener timed out
+
 		klog.V(4).Infof("NodePublishVolume succeeded on volume %q to target path %q, mount already exists.", bucketName, targetPath)
 
 		return &csi.NodePublishVolumeResponse{}, nil
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -392,11 +392,17 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 		return code, fmt.Errorf("failed to get emptyDir path: %w", err)
 	}
 
+	klog.V(4).Infof("checkGcsFuseErr read file %s", emptyDirBasePath+"/error")
+
 	errMsg, err := os.ReadFile(emptyDirBasePath + "/error")
 	if err != nil && !os.IsNotExist(err) {
 		return code, fmt.Errorf("failed to open error file %q: %w", emptyDirBasePath+"/error", err)
 	}
 	if err == nil && len(errMsg) > 0 {
+		// TODO: We need a standard for scraping errors from GCSFuse.
+		// A change in string format in GCSFuse would break this function.
+		// If we are aware of such change, its also tedious to catch and
+		// update tests that are hardcoded with this strings.
 		errMsgStr := string(errMsg)
 		code := codes.Internal
 		if strings.Contains(errMsgStr, "Incorrect Usage") ||
@@ -510,6 +516,7 @@ func PutFlagsFromDriverToTargetPath(flagMap map[string]string, targetPath string
 	absolutePath := filepath.Dir(emptyDirBasePath) + "/" + fileName
 	klog.V(4).Infof("Writing flags needed for gcsfuse defaulting logic to file %q: %v", absolutePath, flagMap)
 
+	// This file is truncated/cleared if it already exists.
 	f, err := os.Create(absolutePath)
 	if err != nil {
 		return fmt.Errorf("failed to create defaulting-flag file: %w", err)
--- a/pkg/csi_mounter/csi_mounter.go
+++ b/pkg/csi_mounter/csi_mounter.go
@@ -106,6 +106,7 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 	klog.V(4).Infof("%v mounting the fuse filesystem", logPrefix)
 	err = m.MountSensitiveWithoutSystemdWithMountFlags(source, target, fstype, csiMountOptions, nil, []string{"--internal-only"})
 	if err != nil {
+		klog.V(4).ErrorS(err, "MountSensitiveWithoutSystemdWithMountFlags failed with error %v", err)
 		return fmt.Errorf("failed to mount the fuse filesystem: %w", err)
 	}
 
@@ -134,8 +135,26 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 		return err
 	}
 
+	// Clean up stale /error file for the fresh gcsfuse process to use.
+	// We put this cleanup step here because:
+	// 1. We can keep visibility into sidecar failures through subsequent NodePublishVolume calls.
+	// 2. SLOs report sidecar failures.
+	// 3. File gets cleaned up before sidecar starts to prevent false positives. These are cases where we report NodePublishVolume (NPV)
+	//    failures (due to stale error file) in scenarios where sidecar takes a while to start up after first NPV suceeded.
+	emptyDirBasePath, err := util.PrepareEmptyDir(target, false /* m.createSocket creates emptydir */)
+	if err != nil {
+		klog.Errorf("failed to get emptyDir path: %v", err)
+	}
+
+	err = util.CheckAndDeleteStaleFile(emptyDirBasePath, "/error")
+	if err != nil {
+		klog.Errorf("failed to check and delete stale /error file: %v", err)
+	}
+
 	// Close the listener and fd after 1 hour timeout
 	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
+	// Since we close the fd, we don't support standalone sidecar restarts.
+	// We support node restarts because a new fd is opened during first NodePublishVolme after restart.
 	go func() {
 		<-ctx.Done()
 		klog.V(4).Infof("%v closing the socket and fd", logPrefix)
@@ -210,6 +229,18 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
 	}
 
 	// Create socket base path.
+	// To handle node restarts we need to:
+	// 1. Delete any stale socket
+	// 2. Create symlink (reasoning explained below)
+	// 3. Create socket
+
+	// Step 1: Attempt to remove a stale mount socket file before creating listener.
+	err = util.CheckAndDeleteStaleFile(emptyDirBasePath, socketName)
+	if err != nil {
+		klog.Errorf("failed to check and delete stale fd mount socket file: %v", err)
+	}
+
+	// Step 2: Create the symlink.
 	// Need to create symbolic link of emptyDirBasePath to socketBasePath,
 	// because the socket absolute path is longer than 104 characters,
 	// which will cause "bind: invalid argument" errors.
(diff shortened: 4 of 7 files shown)


pkg/csi_driver/node.go
```diff
@@ -197,32 +197,29 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
-	// Check if there is any error from the gcsfuse
-	code, err := checkGcsFuseErr(isInitContainer, pod, targetPath)
-	if code != codes.OK {
-		if code == codes.Canceled {
-			klog.V(4).Infof("NodePublishVolume on volume %q to target path %q is not needed because the gcsfuse has terminated.", bucketName, targetPath)
-
-			return &csi.NodePublishVolumeResponse{}, nil
-		}
+	// Check if the target path is already mounted.
+	mounted, err := s.isDirMounted(targetPath)
```
msau42, line 201: Does this check hang if the gcsfuse process crashed?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -104,6 +111,7 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	// so it is safe to force kill the gcsfuse process.
 	go func(cmd *exec.Cmd) {
 		<-ctx.Done()
+		klog.V(4).Infof("context marked as done, force killing gcsfuse process %d", cmd.Process.Pid)
```
saikat-royc: klog.V(4).Infof("context marked as done, sleep for 5 seconds, to evaluate gcsfuse process %d exit state)

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -239,6 +247,17 @@ func collectMetrics(ctx context.Context, port, tempDir string) {
 
 	// Create a unix domain socket and listen for incoming connections.
 	socketPath := filepath.Join(tempDir, metrics.SocketName)
+
+	// Cleanup any stale metrics path to handle container restart scenario.
+	if _, err := os.Stat(socketPath); err == nil {
+		klog.V(4).Infof("found stale socket path for metrics collector %s, attempting delete of the path", socketPath)
```
saikat-royc: please create a playbook entry for metrics collector where the metrics may be malformed across node reboots

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -239,6 +247,17 @@ func collectMetrics(ctx context.Context, port, tempDir string) {
...
+		if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
+			klog.Errorf("error removing stale socket file for metrics collector, %s: %v, ", socketPath, err)
+		}
+		klog.Infof("successfully removed stale socket for metrics collector at %s", socketPath)
```
saikat-royc: klog.V(4).Infof

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -239,6 +247,17 @@ func collectMetrics(ctx context.Context, port, tempDir string) {
...
+	}
+
+	klog.Infof("Start listen on metrics collector socket %s", socketPath)
```
saikat-royc: klog.V(4).Infof

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -349,6 +368,17 @@ func getAudienceFromContextAndIdentityProvider(ctx context.Context, identityProv
 }
 
 func StartTokenServer(ctx context.Context, tokenURLSocketPath string, identityProvider string) {
+	// Clean up any stale socket file before creating a new one.
+	if _, err := os.Stat(tokenURLSocketPath); err == nil {
```
saikat-royc: please create a helper function to check a socket path and delete the socket. this same block is used to delete metrics socket

pkg/csi_mounter/csi_mounter.go
```diff
@@ -136,6 +137,8 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 
 	// Close the listener and fd after 1 hour timeout
 	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
+	// Since we close the fd, we don't support standalone sidecar restarts.
+	// We support node restarts because a new fd is opened during first NodePublishVolme after restart.
```
saikat-royc, line 157: please create a playbook entry to update to add a section for node restarts https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/docs/troubleshooting.md#container-restarts

pkg/csi_mounter/csi_mounter.go
```diff
@@ -213,36 +216,110 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
...
+	// To handle node restarts we need to:
+	// 1. Delete any stale socket
+	// 2. Create symlink
+	// 3. Create socket
+
 	socketBasePath := util.GetSocketBasePath(target, m.fuseSocketDir)
+	socketPath := filepath.Join(socketBasePath, socketName)
+
+	// Step 1: Attempt to remove a stale socket file before creating listener.
+	if _, err := os.Stat(socketPath); err == nil {
```
saikat-royc: same block of checking socket path and delete. use a helper function

pkg/csi_mounter/csi_mounter.go
```diff
@@ -213,36 +216,110 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
...
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return nil, fmt.Errorf("failed to remove stale socket file %q: %w", socketPath, err)
+			}
+		} else {
+			klog.Infof("stale mount socket %s deleted successfully", socketPath)
+		}
+	}
+
+	// Step 2: Create the symlink.
```
saikat-royc, line 243: please add a detailed comment on why we do a symlink

pkg/csi_mounter/csi_mounter.go
```diff
@@ -213,36 +216,110 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
...
-	// Change the socket ownership
+	// We also need to change the socket ownership in 3 steps.
+	// 1. Change ownership of base of emptyDirBasePath.
+	// 2. Change ownership of sockets directory.
+	// 3. Change ownership of the mount socket itself.
+
 	targetSocketPath := filepath.Join(emptyDirBasePath, socketName)
-	if err = os.Chown(filepath.Dir(emptyDirBasePath), webhook.NobodyUID, webhook.NobodyGID); err != nil {
+	// First changing ownership of parent directory where socket resides.
+	// Full path: .../volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp/.volumes/
```
saikat-royc: please add the full path template. /var/lib/kubelet/... in the comment

pkg/csi_mounter/csi_mounter.go
```diff
@@ -213,36 +216,110 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
...
 	}
 
+	if err = m.checkAndDeleteConfigFile(emptyDirBasePath, "config.yaml"); err != nil {
+		klog.Errorf("checkAndDeleteConfigFile err %s", err)
+	}
+
 	return l, nil
 }
 
+func (m *Mounter) checkAndDeleteConfigFile(dirPath, fileName string) error {
```
saikat-royc: this function can be reused to delete a socket file as well. we can create a generic function to delete file paths

pkg/csi_driver/node.go
siyanshen, line 201: Wondering why you switched the order of isDirMounted check and checkGcsFuseErr

test/e2e/specs/specs.go
```diff
@@ -267,10 +266,46 @@ func (t *TestPod) WaitForPodNotFoundInNamespace(ctx context.Context) {
...
+		framework.ExpectNoError(err)
+
+		return podLogs
+	})
+}
+
+// lookForString looks for the given string in the output of fn, repeatedly calling fn until
+// the timeout is reached or the string is found. Returns last log and possibly
+// error if the string was not found.
+func lookForString(expectedString string, timeout time.Duration, fn func() string) (result string, err error) {
```
siyanshen: nit: `pollForString` or `waitForString` might more clearly convey the polling/waiting behavior.

test/e2e/specs/specs.go
```diff
@@ -267,10 +266,46 @@ func (t *TestPod) WaitForPodNotFoundInNamespace(ctx context.Context) {
 }
 
 func (t *TestPod) WaitForLog(ctx context.Context, container string, expectedString string) {
-	_, err := e2epodooutput.LookForStringInLogWithoutKubectl(ctx, t.client, t.namespace.Name, t.pod.Name, container, expectedString, pollTimeout)
+	_, err := LookForStringInLogWithoutKubectl(ctx, t.client, t.namespace.Name, t.pod.Name, container, expectedString, pollTimeout)
```
siyanshen: These changes in test/ seems unrelated to node restart problem. Why are we modifying WaitForLog?

hime: I left a note in PR description but in short, is using this function too much causes test failures. Was thinking of just making a separate PR for this specific issue.

hime: Created #659

test/e2e/specs/specs.go
hime: left notes on #659

pkg/csi_driver/node.go
hime: This check is using [mounter.List()](https://github.com/kubernetes/mount-utils/blob/1384034db5f7b236bb35c82029b3c13289d33245/mount_linux.go#L426) which checks the proc mounts file. This way of checking mounts is not blocking when fuse client (e.g. gcsfuse) crashes. The mount point still shows up as mounted if gcsfuse crashes.

pkg/csi_mounter/csi_mounter.go
hime: ack

pkg/csi_mounter/csi_mounter.go
hime: done.

pkg/csi_driver/node.go
hime: The ordering does not make a difference here so I've reverted this change.

pkg/csi_mounter/csi_mounter.go
hime: done. moved to util package.

pkg/csi_driver/node.go
siyanshen: are checkGcsFuseErr and checkSidecarContainerErr only necessary when mounted? Are they not needed when isDirMounted returns false?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -65,7 +66,7 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	if mc.TokenServerIdentityProvider != "" {
 		tp := filepath.Join(mc.TempDir, TokenFileName)
```
siyanshen: Removed tp definition since StartTokenServer is rewritten.

pkg/csi_driver/node.go
hime: We only care about the GCSFuse/sidecar failures when we expect the sidecar to be running. When the mountpoint is not setup, this means we haven't made the filedescriptor for the sidecar to pickup. This means that after a node restart, we ignore any stale errors from the previous abrupt termination.

siyanshen (approved):

pkg/sidecar_mounter/sidecar_mounter.go
hime: done

leiyiz (commented):
/lgtm

docs/troubleshooting.md
```diff
@@ -437,6 +437,13 @@ See GKE documentation [Storage for GKE clusters overview](https://cloud.google.c
 
 If the sidecar container (`gke-gcsfuse-sidecar`) hosting the gcsfuse process, crashes for some reason (e.g OOM Kill, or fatals), and kubelet decides to inplace restart the container, gcsfusecsi driver does not handle this scenario gracefully. This is because in the CSI NodePublishVolume the driver [checks the last known state](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/v1.15.4/pkg/csi_driver/utils.go#L431) of the container and fails fast. To mitigate this, delete and create the pod.
 
+### Node level restarts
+
+GCSFuse CSI Driver can now manage node restarts without impacting pods using GCSFuse-backed volumes. The driver ensures a clean slate by clearing stale files or sockets, and re-establishing the gcsfuse /dev/fuse file descriptor during the NodePublishVolume phase after a restart.
+
+Known issues after a node restart:
+- Metrics will not be visible on Cloud Monitoring after node restart. If they become visible, assume that they will be malformed.
```
saikat-royc, line 445: what is the actionable guidance for the user? Do they need to disable metrics? 
how can the user detect the metrics are malformed due to node restarts? same question applies to the oncaller

pkg/csi_driver/node.go
```diff
@@ -197,33 +197,34 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
-
 	// Check if the target path is already mounted
 	mounted, err := s.isDirMounted(targetPath)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to check if path %q is already mounted: %v", targetPath, err)
 	}
 
+	// Skip mount if the mount on target path already exists.
 	if mounted {
+		// Check if there is any error from the gcsfuse
```
saikat-royc, line 208: please add some comment on the reasoning behind moving the checks only if mounted is reported as true (which means the first half of the mount has been setup and gcsfuse is ready to proceed). preserve the original behavior for node publish reporting error on gcsfuse failures

pkg/csi_driver/utils.go
```diff
@@ -392,11 +392,17 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
...
 	}
 
+	klog.V(4).Infof("checkGcsFuseErr read file %s", emptyDirBasePath+"/error")
+
 	errMsg, err := os.ReadFile(emptyDirBasePath + "/error")
 	if err != nil && !os.IsNotExist(err) {
 		return code, fmt.Errorf("failed to open error file %q: %w", emptyDirBasePath+"/error", err)
 	}
 	if err == nil && len(errMsg) > 0 {
+		// TODO: We need a standard for scraping errors from GCSFuse.
```
saikat-royc, line 402: please create internal bugs for the TODO. this can be picked up as chores and tech debt scraping @hime

pkg/csi_mounter/csi_mounter.go
```diff
@@ -106,6 +106,7 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 	klog.V(4).Infof("%v mounting the fuse filesystem", logPrefix)
 	err = m.MountSensitiveWithoutSystemdWithMountFlags(source, target, fstype, csiMountOptions, nil, []string{"--internal-only"})
 	if err != nil {
+		klog.V(4).ErrorS(err, "MountSensitiveWithoutSystemdWithMountFlags failed with error %v", err)
```
saikat-royc, line 109: isin't klog.Error sufficient? why log only at v(4)?

pkg/csi_mounter/csi_mounter.go
```diff
@@ -134,8 +135,26 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 		return err
 	}
 
+	// Clean up stale /error file for the fresh gcsfuse process to use.
```
saikat-royc, line 138: please be more explicit in the description. a new reader cannot understand what "fresh" means. e.g. clean up stale/error file, when the csi setup the mount on the node publish path, and the sidecar is about to invoke the gcsfuse process for the first time to complete the bottom half (registration of the fd for events) of the mount.

pkg/util/util.go
```diff
@@ -178,3 +178,30 @@ func GetSocketBasePath(targetPath, fuseSocketDir string) string {
...
+	}
+
+	// Stale file exists.
+	klog.V(4).Infof("File '%s' found in '%s'. Attempting to delete...", fileName, dirPath)
+
+	// Delete the file.
+	if deleteErr := os.Remove(filePath); deleteErr != nil {
+		return fmt.Errorf("failed to delete file '%s': %w", filePath, deleteErr)
+	}
+	klog.Infof("Stale file '%s' successfully deleted.", fileName)
```
saikat-royc, line 204: nit: 
" ... successfully deleted" 

'.' not needed

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -82,6 +82,15 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 		}
 	}
 
+	// Cleanup stale /error file. This check is relevant in sidecar mounter because it:
+	// - Acts as in cases where NodePublishVolume fails to clean up /error file.
```
saikat-royc, line 86: I am curious if we have faced any issues? seems like an overkill to add the delete in 2 places (csi mounter and sidecar mounter)?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -348,8 +366,15 @@ func getAudienceFromContextAndIdentityProvider(ctx context.Context, identityProv
 	return audience, nil
 }
 
-func StartTokenServer(ctx context.Context, tokenURLSocketPath string, identityProvider string) {
+func StartTokenServer(ctx context.Context, tokenURLBasePath, tokenSocketName, identityProvider string) {
+	// Clean up any stale socket file before creating a new one.
+	err := util.CheckAndDeleteStaleFile(tokenURLBasePath, tokenSocketName)
+	if err != nil {
+		klog.Errorf("failed to check and delete stale token server socket file: %v", err)
```
saikat-royc, line 373: should we return here to have a deterministic behavior?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -239,6 +249,14 @@ func collectMetrics(ctx context.Context, port, tempDir string) {
 
 	// Create a unix domain socket and listen for incoming connections.
 	socketPath := filepath.Join(tempDir, metrics.SocketName)
+
+	// Cleanup any stale metrics path to handle container restart scenario.
+	err := util.CheckAndDeleteStaleFile(tempDir, metrics.SocketName)
+	if err != nil {
+		klog.Errorf("failed to check and delete stale metrics socket file: %v", err)
```
saikat-royc, line 256: should we return here for deterministic behavior?

pkg/util/util.go
```diff
@@ -178,3 +178,30 @@ func GetSocketBasePath(targetPath, fuseSocketDir string) string {
 	// based on the pod and volume.
 	return filepath.Join(fuseSocketDir, sha1Hash(podID+"_"+volumeName))
 }
+
+func CheckAndDeleteStaleFile(dirPath, fileName string) error {
```
saikat-royc, line 182: how about an unit test for this?

hime: Done in #668.

pkg/sidecar_mounter/sidecar_mounter.go
hime: Done in #668.

pkg/sidecar_mounter/sidecar_mounter.go
hime: Done in #668.

pkg/sidecar_mounter/sidecar_mounter.go
hime: We did not face any issues, this is overkill. Wanted to have check in case we ever supported container restarts, but that should be applied if/when we get there. Fixed in #668.

pkg/util/util.go
hime: Done in #668.

pkg/csi_mounter/csi_mounter.go
hime: Done in #668.

pkg/csi_mounter/csi_mounter.go
hime: You're right. If its an error log, its worth to always log. Done in #668.

pkg/csi_driver/utils.go
hime: Done in #internal and cc'ed

pkg/csi_driver/node.go
hime: Done in #668.

docs/troubleshooting.md
hime: Updated in #668, although I think it needs another round of feedback from you.

PR #659, opened 2025-05-30
Allow retries for e2e helper function WaitForLog.

hime:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature

/kind flake

This PR adds retries to the WaitForLog function used in e2e testing, as too many tests calling this function can cause temporary failures. Overloading the server with requests from e2e testsuite causes flaky tests that show as `400 bad request` errors.

Fixes #

```release-note
None
```

### DIFF
--- a/test/e2e/specs/specs.go
+++ b/test/e2e/specs/specs.go
@@ -43,7 +43,6 @@ import (
 	e2ejob "k8s.io/kubernetes/test/e2e/framework/job"
 	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
-	e2epodooutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
 	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
 	imageutils "k8s.io/kubernetes/test/utils/image"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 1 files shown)


test/e2e/specs/specs.go
```diff
@@ -267,10 +266,48 @@ func (t *TestPod) WaitForPodNotFoundInNamespace(ctx context.Context) {
...
+// Based on the following kubernetes function: https://github.com/kubernetes/kubernetes/blob/4832b57e48b98af0b0d558fa9810dc77b906ea62/test/e2e/framework/pod/output/output.go#L125
+func LookForStringInLogWithoutKubectl(ctx context.Context, client clientset.Interface, ns string, podName string, container string, expectedString string, timeout time.Duration) (result string, err error) {
+	return lookForString(expectedString, timeout, func() string {
+		// Implementing function with retries to circumvent server overload failures.
+		var podLogs string
+		var err error
+
+		maxRetries := 10
+		for retries := 0; retries < maxRetries; retries++ {
+			podLogs, err = e2epod.GetPodLogs(ctx, client, ns, podName, container)
```
siyanshen, line 283: This retry mechanism can potentially overwhelm the server too, correct? Maybe add an exponential backoff mechanism here. Like `time.Sleep(time.Second * math.Pow(5, retries))`

hime:
Node restart support #657 was able to keep original e2e test behavior, which means this fix is no longer relevant.

PR #759, opened 2025-06-18
update to gcsfuse v3, and fix metrics and local_file tests

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

- This bumps the GCSFuse binary to GCSFuse v3.0.0
-  Passes an additional flag, write-global-max-blocks=-1 to the streaming_writes tests, which is required now that streaming writes is enabled by default in GCSFuse v3. 
- Fixes the gcsfuse metrics tests, which were failing because the number of GCS methods we call differs between GCSFuse v2 and v3.
   - In v3, GCSFuse now calls CreateObjectChunkWriter and FinalizeUpload, but in GCSFusev2 UpdateObject exists but not those other 2.  This makes the sum of GCS methods called in v3 be 8, but in v2 it was 7. Counting the number of methods is not a good way to test metrics, and can break easily as new features are added, so I removed this. 
- This also fixes local_dir tests, which now needs an additional flag --enable-streaming-writes=false, to pass. See comments in code for details.
- This sets the branch to clone gcsfuse Integration tests from for v3.0.0 to the master branch. This is needed because GCSFuse made test fixes after cutting the v3.0.0, and we don't want to block the release on waiting for them to create a new patch release for this.     

The cloud_profiler tests will be added in a https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/740, as they are only for debugging purposes, and aren't release blocking. 

Manually tested:

```
gcloud container clusters create test-gcsfusev3 \
    --zone=us-central1-a \
    --workload-pool=amacaskill-gke-dev.svc.id.goog --machine-type=n2-standard-4
gcloud container clusters get-credentials test-gcsfusev3 --zone us-central1-a
# build and push image to registry
make build-image-and-push-multi-arch REGISTRY=us-central1-docker.pkg.dev/amacaskill-gke-dev/csi-dev STAGINGVERSION=v999.999.999
# install non-managed driver on GKE cluster
make install REGISTRY=us-central1-docker.pkg.dev/amacaskill-gke-dev/csi-dev STAGINGVERSION=v999.999.999 PROJECT=amacaskill-gke-dev

#Confirm that the driver is up and running
kubectl get CSIDriver,Deployment,DaemonSet,Pods -n gcs-fuse-csi-driver


# run tests on that cluster.
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false E2E_TEST_SKIP="should.succeed.in.performance.test" STAGINGVERSION=v999.999.999 REGISTRY=us-central1-docker.pkg.dev/amacaskill-gke-dev/csi-dev
```

Fixes #

```release-note
Updates GCSFuse to v3.0.0
```


test/e2e/e2e_test.go
```diff
@@ -115,10 +116,10 @@ var _ = ginkgo.Describe("E2E Test Suite", func() {
 		testsuites.InitGcsFuseCSIMetadataPrefetchTestSuite,
 		testsuites.InitGcsFuseMountTestSuite,
 	}
-
+	enableZB := false
```
hime: we set enableZB through [CLI flag](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/744de0269d8a3c47a8927618c8cfa31819f6aee7/test/e2e/main.go#L44). Is there a reason to override this?

amacaskill: I just need to rebase, this is in master right now. ZB flag was returning true even when we passed false to the flag, which broke the testgrid so I disabled it.

PR #762, opened 2025-06-20
Added support for implicit dir creation using go client

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This pr replaces the current implementation of implicit_dir creation (implemented with gcs cp commands) with go client implementation. The functionality remains the same but is needed to support zb bucket testing. ZB currently do not support gcloud commands, so to handle test cases we must use go client.

How was this tested - 
e2e tests were ran using a custom image containing the changes -v

```release-note
make build-image-and-push-multi-arch REGISTRY=gcr.io/prow-gob-internal-boskos-02 STAGINGVERSION=v999.999.999

make install REGISTRY=gcr.io/prow-gob-internal-boskos-02 STAGINGVERSION=v999.999.999 PROJECT=gcr.io/prow-gob-internal-boskos-02
```

Tests were ran with zb enabled -v

```release-note
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false E2E_TEST_SKIP="should.succeed.in.performance.test" STAGINGVERSION=v999.999.999 REGISTRY=gcr.io/prow-gob-internal-boskos-02 E2E_TEST_FOCUS="rename_dir_limit" ENABLE_ZB=true GCSFUSE_CLIENT_PROTOCOL=grpc
```
and without -v

```release-note
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false E2E_TEST_SKIP="should.succeed.in.performance.test" STAGINGVERSION=v999.999.999 REGISTRY=gcr.io/prow-gob-internal-boskos-02
```
focus was put on tests from subpath module due to the change in location of implicit dir method
```release-note
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false E2E_TEST_SKIP="should.succeed.in.performance.test" STAGINGVERSION=v999.999.999 REGISTRY=gcr.io/prow-gob-internal-boskos-02 E2E_TEST_FOCUS="subPath"
```

Fixes #

```release-note
NONE
```


pkg/util/goclientobjectcommands/storage_client_utils.go
```diff
@@ -0,0 +1,115 @@
...
+	"io"
+	"log"
+	"os"
+
+	"cloud.google.com/go/storage"
+	"cloud.google.com/go/storage/experimental"
+)
+
+// NewGCSClient creates and returns a new Google Cloud Storage client.
+func NewGCSClient(ctx context.Context, isZBEnabled bool) (*storage.Client, error) {
```
amacaskill: storage client here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1fc57d253888397895a74af8649af67fb27ff67b/pkg/cloud_provider/storage/storage.go#L58-L93 Can we add this functionality to that, so we don't have to create a new client every time we call  CreateImplicitDirInBucket?

pkg/util/goclientobjectcommands/storage_client_utils.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log"
+	"os"
+
+	"cloud.google.com/go/storage"
+	"cloud.google.com/go/storage/experimental"
```
amacaskill: I don't think we want to use experimental packages: Package experimental is a collection of experimental features that might have some rough edges to them. Housing experimental features in this package results in a user accessing these APIs as `experimental.Foo`, thereby making it explicit that the feature is experimental and using them in production code is at their own risk.

amacaskill:
I can't really review this as it changes too many files and crashes my browser. when I try to leave comments on the files. Can you please try to not use experimental package, and try to use the Storage go client we already have. I think that would mean you don't need to update so many package versions which will cause the thousands of vendor files to be updated.

amacaskill:
Please fix the `What type of PR is this?` to add a blank line so it is visible. 

Please add a release note. 

Also, can you please add which test you ran (a ZB case and non-ZB case), to confirm it is working and won't break testgrids?

pkg/util/goclientobjectcommands/storage_client_utils.go
amacaskill: Specifically, can we use the EnableZB flag within the SetupService function: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1fc57d253888397895a74af8649af67fb27ff67b/pkg/cloud_provider/storage/storage.go#L87C1-L93, and create the GRPC client with WithGRPCBidiReads() , and assign that to be the &gcsService{storageClient: storageClient}, nil.

Then in CreateImplicitDirInBucket, we can just pass our existing storageClient to CreateImplicitDirInBucket, because it will be the one with the experimental part since EnableZB = true, that way we don't need to create a  new GCS client every time we call CreateImplicitDirInBucket?

pkg/util/goclientobjectcommands/storage_client_utils.go
amacaskill: Discussed offline, the reason why experimental is needed is because ZB functionality like `WithGRPCBidiReads` is only in the experimental package: storage.NewGRPCClient(ctx, experimental.WithGRPCBidiReads())

pkg/util/goclientobjectcommands/storage_client_utils.go
chrisThePattyEater: Ask was to prevent multiple instances of the client being generated by creating one and reusing it as needed.

To address this we have added a flag "enableZB" to the SetUpServiceWithDefaultCredential method. Based on the flag we handle the required zb client creation.

SetUpServiceWithDefaultCredential is called during testing driver initialization and this service is passed to CreateImplicitDirInBucket

UploadGcsObject (the method called during createImplicitDir that needs the client) is moved into storage.go so that it can now access the gcsservice and its client

test/e2e/specs/specs.go
```diff
@@ -1184,10 +1186,20 @@ func CreateImplicitDirInBucket(dirPath, bucketName string) {
...
 
-	//nolint:gosec
-	if output, err := exec.Command("gsutil", "cp", bucketName, fmt.Sprintf("gs://%v/%v/", bucketName, dirPath)).CombinedOutput(); err != nil {
-		framework.Failf("Failed to create a implicit dir in GCS bucket: %v, output: %s", err, output)
+	// ZB does not currently support gcloud commands, so we use the go GCS client directly.
+	if zbEnabled {
+		dstPrefix := fmt.Sprintf("%s/", dirPath)
+
+		if err := (*storageService).UploadGcsObject(ctx, bucketName, bucketName, dstPrefix, false, zbEnabled); err != nil {
+			klog.Fatalf("CopyBucketToPath failed: %v", err)
```
amacaskill: Why do we use klog.Fatalf here, but framework.Failf for non-zonal buckets. 

Should this say UploadGcsObject failed? And not CopyBucketToPath?

test/e2e/specs/testdriver.go
```diff
@@ -48,7 +48,8 @@ type GCSFuseCSITestDriver struct {
 	ClientProtocol              string
 	skipGcpSaTest               bool
 	EnableHierarchicalNamespace bool
-	EnableZB                    bool // Enable Zonal Buckets
+	EnableZB                    bool             // Enable Zonal Buckets
+	storageService              *storage.Service // Client to interact with GCS
```
amacaskill, line 52: Why do we need another storageService? I believe storageServiceManager wraps the service

test/e2e/specs/testdriver.go
```diff
@@ -66,6 +67,7 @@ type gcsVolume struct {
 // InitGCSFuseCSITestDriver returns GCSFuseCSITestDriver that implements TestDriver interface.
 func InitGCSFuseCSITestDriver(c clientset.Interface, m metadata.Service, bl string, skipGcpSaTest, enableHierarchicalNamespace bool, clientProtocol string, enableZB bool) storageframework.TestDriver {
 	ssm, err := storage.NewGCSServiceManager()
+	storageService, err := ssm.SetupServiceWithDefaultCredential(context.Background(), enableZB)
```
amacaskill: We shouldn't need to call  SetupServiceWithDefaultCredential here. This is called already in prepareStorageService : https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/d86772ad7602e17a54e9661c233fefc60d380eb9/test/e2e/specs/testdriver.go#L393

And prepareStorageService, should be called in your UploadGCSObject helper method, which also sets up the ServiceBucket which should be passed to UploadGCSObject as the param. Please follow other  examples in testdriver.go: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/4f5486741e244f7bd8b68fd329fef85e51abe449/test/e2e/specs/testdriver.go#L436-L462

This means we don't need to store an additional storage service in GCSFuseCSITestDriver. The storageService is the same thing that storageServiceManager wraps, so we are adding two clients to GCSFuseCSITestDriver, when  we should just add one. So please remove storageService from GCSFuseCSITestDriver after you fix this

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +302,51 @@ func ParseErrCode(err error) codes.Code {
 
 	return code
 }
+
+// UploadGcsObject uploads a local file to a specified GCS bucket and object.
+// Handles gzip compression if requested.
+func (service *gcsService) UploadGcsObject(ctx context.Context, localPath, bucketName, objectName string, uploadGzipEncoded bool, zbEnabled bool) error {
```
amacaskill: Since we always set uploadGzipEncoded to false, lets remove this functionality / param to simplify the function.

test/e2e/specs/specs.go
```diff
@@ -1184,10 +1186,20 @@ func CreateImplicitDirInBucket(dirPath, bucketName string) {
...
 	}()
 
-	//nolint:gosec
-	if output, err := exec.Command("gsutil", "cp", bucketName, fmt.Sprintf("gs://%v/%v/", bucketName, dirPath)).CombinedOutput(); err != nil {
-		framework.Failf("Failed to create a implicit dir in GCS bucket: %v, output: %s", err, output)
+	// ZB does not currently support gcloud commands, so we use the go GCS client directly.
+	if zbEnabled {
+		dstPrefix := fmt.Sprintf("%s/", dirPath)
+
+		if err := (*storageService).UploadGcsObject(ctx, bucketName, bucketName, dstPrefix, false, zbEnabled); err != nil {
```
amacaskill: UploadGcsObject(ctx, bucketName, bucketName will result in failures. One should be the dir path. Please make sure you manually test your changes with test cases that exercise this case.

pkg/util/goclientobjectcommands/storage_client_utils.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+
+	"cloud.google.com/go/storage"
+)
+
+// NewWriter is a wrapper over storage.NewWriter which
+// extends support to zonal buckets.
+func NewWriter(ctx context.Context, o *storage.ObjectHandle, client *storage.Client, zbEnabled bool) (wc *storage.Writer, err error) {
+	wc = o.NewWriter(ctx)
+	wc.FinalizeOnClose = true
```
amacaskill: Why do we need FinalizeOnClose set to true? Is this needed for ZB? If so, please put it into a if ZB is true statement.

It seems like what this function does, is it just sets append to true for ZB. What case would we actually hit the case were zbEnabled bool, doesn't match attrs.StorageClass )? If there aren't any, lets simplify this function to just rely on one. We can call it NewZBWriter or something, then just do
wc = o.NewWriter(ctx)
wc.append=true, then return.

pkg/util/goclientobjectcommands/storage_client_utils.go
amacaskill: Thanks, I reviewed and left comments. Please let me know when you manually tested with the new changes, and  are ready for another review

test/e2e/specs/specs.go
chrisThePattyEater: i did test this, upload gcsobject uses bucket name as an actual bucket name and uses objectname as the path for the destination,
service.storageClient.Bucket(bucketName).Object(objectName)
object name in this case being dst prefix from line 1191
dstPrefix := fmt.Sprintf("%s/", dirPath)
so the path we are copying to becomes
gs://bucketName/objectName

amacaskill: If we aren't using localPath for anything and setting it to the same thing as bucketName, then can we remove that variable? Its just used by gzip, which we should remove

chrisThePattyEater: well no local path is used as the path from which we copy, this -> gs://bucketName/objectName is the path we upload to

pkg/util/goclientobjectcommands/storage_client_utils.go
chrisThePattyEater: making this a zb only function as recommended -"FinalizeOnClose indicates whether the Writer should finalize an object when closing the write stream. This only applies to Writers where Append is true"

pkg/util/goclientobjectcommands/file_operations.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Creates a temporary file (name-collision-safe) in /tmp with given content.
+// If gzipCompress is true, output file is a gzip-compressed file.
+// Caller is responsible for deleting the created file when done using it.
+// Failure cases:
+// 1. os.CreateTemp() returned error or nil handle
+// 2. gzip.NewWriter() returned nil handle
+// 3. Failed to write the content to the created temp file
+func CreateLocalTempFile(content string, gzipCompress bool) (string, error) {
```
amacaskill: I don't see where this is used. Can we remove this?

amacaskill: And remove writeTextToFile and writeGzipToFile

test/e2e/testsuites/subpath.go
```diff
@@ -135,6 +135,7 @@ func (t *gcsFuseCSISubPathTestSuite) DefineTests(driver storageframework.TestDri
 		bucketName := l.config.Prefix
 
 		// Create sub-paths using gsutil
+		// Passing zb as false because this test does not run with HNS enabled buckets and ZB is only relevant when HNS is enabled.
```
amacaskill, line 138: by HNS do you mean hierarchal namespace buckets? Why is HNS a prereq to ZB?

test/e2e/specs/testdriver.go
```diff
@@ -485,3 +487,31 @@ func prepareGcpSAName(ns string) string {
...
+			if err != nil {
+				e2eframework.Failf("Failed to delete the empty data file: %v", err)
+			}
+		}()
+		// ZB does not currently support gcloud commands, so we use the go GCS client directly.
+		if err := storageService.UploadGcsObject(ctx, bucketName, bucketName, fmt.Sprintf("%s/", dirPath), zbEnabled); err != nil {
+			e2eframework.Failf("UploadGcsObject failed: %v", err)
+		}
+	} else {
+		CreateImplicitDirInBucket(dirPath, bucketName)
```
amacaskill: What's stopping us from just using golang client for all of the test cases. It should be equivalent, and I think this would simplify things and it would be nice to not have multiple implicitDir functions because in the future, people might get confused which one to use.

If n.enableZB is true, then  prepareStorageService will use a storageService  with a storageClient := storage.NewGRPCClient(ctx, experimental.WithGRPCBidiReads()). If n.enableZB is false, then  prepareStorageService will use a storageService with a storageClient  := storage.NewClient(ctx)



```
func (n *GCSFuseCSITestDriver) createImplicitDirInBucket(ctx context.Context, dirPath, bucketName string) {
		storageService, err := n.prepareStorageService(ctx)
		if err != nil {
			e2eframework.Failf("failed to prepare storage service: %w", err)
			return
		}
		// Use bucketName as the name of a temp file since bucketName is unique.
		f, err := os.Create(bucketName)
		if err != nil {
			e2eframework.Failf("Failed to create an empty data file: %v", err)
		}
		f.Close()
		defer func() {
			err = os.Remove(bucketName)
			if err != nil {
				e2eframework.Failf("Failed to delete the empty data file: %v", err)
			}
		}()
		// ZB does not currently support gcloud commands, so we use the go GCS client directly.
		if err := storageService.UploadGcsObject(ctx, bucketName, bucketName, fmt.Sprintf("%s/", dirPath), zbEnabled); err != nil {
			e2eframework.Failf("UploadGcsObject failed: %v", err)
		}
```


The only thing I'm not positive about is within UploadGCSBucket, you might need to add an if/else dependent on zbEnabled, because You call ZBWriter specifically. 

```
// NewZBWriter is a wrapper over storage.NewWriter which extends support to zonal buckets.
func NewZBWriter(ctx context.Context, o *storage.ObjectHandle) (wc *storage.Writer) {
	wc = o.NewWriter(ctx)
	// Changes specific to zonal bucket
	wc.Append = true
	wc.FinalizeOnClose = true
	return wc
}
```

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,29 @@ func ParseErrCode(err error) codes.Code {
 
 	return code
 }
+
+// UploadGcsObject uploads a local file to a specified GCS bucket and object.
+// Handles gzip compression if requested.
+func (service *gcsService) UploadGcsObject(ctx context.Context, filePathToUpload, bucketName, objectName string, zbEnabled bool) error {
+	// Create a writer to upload the object.
+	obj := service.storageClient.Bucket(bucketName).Object(objectName)
+	w := foUtils.NewZBWriter(ctx, obj)
```
amacaskill: What about ZB requires a ZB writer that adds append and FinalizeOnClose? Can we make this a writer that takes some options so we can use this UploadGcsObject function for non ZB cases also?

pkg/util/goclientobjectcommands/file_operations.go
amacaskill: I guess I see some  tests like "gcsfuseIntegration should succeed in gzip test 1", but these are gcsfuse integration tests, not our e2e tests, so I don't think we need gzip correct?

test/e2e/testsuites/subpath.go
amacaskill: ZB requires HNS. Just curious, why don't we run this on HNS?

pkg/cloud_provider/storage/storage.go
chrisThePattyEater: ZB requires two additional settings on top of the traditional writer, 
w.Append = true
w.FinalizeOnClose = true

test/e2e/specs/testdriver.go
chrisThePattyEater: acked, moved all implicit dir creation to use go client and ran reran specific tests in subpath module to ensure code works as expected

test/e2e/testsuites/subpath.go
chrisThePattyEater: this sub path test suite is not part of the tests initialized for non hns test suites.

pkg/util/goclientobjectcommands/file_operations.go
chrisThePattyEater: acked, removed unused methods from file_ops

pkg/util/goclientobjectcommands/storage_client_utils.go
chrisThePattyEater: this method has since been removed and isntead we conditionally add the required w.Append = true and w.FinalizeOnClose = true if the writer being inited is for zb. This happens in UploadGcsObject within storage.go

test/e2e/specs/testdriver.go
chrisThePattyEater: acked, this has since been removed

pkg/cloud_provider/storage/storage.go
amacaskill: Can you point me to where I can find these requirements?

pkg/util/goclientobjectcommands/storage_client_utils.go
chrisThePattyEater: Did some checks with people who with more knowledge on the subject than me, tldr is the zonal buckets api requires append for zb and while finalizeOnClose is not required without it gcsfuse has been known to be problematic - "it has some limitations which we are trying to figure out"

test/e2e/specs/testdriver.go
```diff
@@ -485,3 +487,26 @@ func prepareGcpSAName(ns string) string {
...
 	return ns
 }
+
+func (n *GCSFuseCSITestDriver) CreateImplicitDirInBucket(ctx context.Context, dirPath, bucketName string, zbEnabled bool) {
+	storageService, err := n.prepareStorageService(ctx)
+	if err != nil {
+		e2eframework.Failf("failed to prepare storage service: %w", err)
+		return
+	}
+	// Use bucketName as the name of a temp file since bucketName is unique.
```
gargnitingoogle, line 497: nit: You could instead use os.CreateTemp here, for which you don't need to think of a unique name.

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,33 @@ func ParseErrCode(err error) codes.Code {
 
 	return code
 }
+
+// UploadGcsObject uploads a local file to a specified GCS bucket and object.
+// Handles gzip compression if requested.
```
gargnitingoogle, line 306: Remove this line. It's irrelevant here.

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,33 @@ func ParseErrCode(err error) codes.Code {
...
+func (service *gcsService) UploadGcsObject(ctx context.Context, filePathToUpload, bucketName, objectName string, zbEnabled bool) error {
+	// Create a writer to upload the object.
+	obj := service.storageClient.Bucket(bucketName).Object(objectName)
```
gargnitingoogle, line 309: Is the bucket with name bucketName guaranteed to exist ? If not, this will fail here without properly returning error .

gargnitingoogle: Also, to be cautious, could you also confirm that the bucket here is indeed a zonal bucket (when zbEnabled is true) by checking that its storage-class is "RAPID" like I'm doing [here](https://github.com/GoogleCloudPlatform/gcsfuse/blob/5123c98fbc406d68d1efd8ec06eb660881e53eb1/tools/integration_tests/util/client/storage_client.go#L148) ?

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,33 @@ func ParseErrCode(err error) codes.Code {
...
+	w := obj.NewWriter(ctx)
+	if zbEnabled {
+		w.Append = true
+		w.FinalizeOnClose = true
```
gargnitingoogle, line 322: Add a note/todo to remember that in long term, FinalizeOnClose will be set to false for ZB.

```suggestion
        // TODO: Switch this to false when this is supported by gcsfuse.
		w.FinalizeOnClose = true
```

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,33 @@ func ParseErrCode(err error) codes.Code {
...
+	}
```
gargnitingoogle, line 323:
```suggestion
	} else {
	  w.FinalizeOnClose = true
	}
```

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,33 @@ func ParseErrCode(err error) codes.Code {
...
+	defer func() {
+		if err := w.Close(); err != nil {
```
gargnitingoogle: Just so you know, if you intend to return this error returned from `w.close()` to the caller of this function, that won't happen here, because err is a different temporary variable. Here, this error value will only be used in the log below, but not returned. If you want to also return this variable, then you need to change your return-type to a named variable, e.g. 

```go
func (service *gcsService) UploadGcsObject(ctx context.Context, filePathToUpload, bucketName, objectName string, zbEnabled bool) (err error) {
```

... and then update that returned variable `err` in the defer func definition.

```go
defer func() {
  if err = w.close(); err != nil {
```

test/e2e/specs/testdriver.go
```diff
@@ -208,11 +210,11 @@ func (n *GCSFuseCSITestDriver) CreateVolume(ctx context.Context, config *storage
 		case InvalidMountOptionsVolumePrefix:
 			mountOptions += ",invalid-option"
 		case ImplicitDirsVolumePrefix:
-			CreateImplicitDirInBucket(ImplicitDirsPath, bucketName)
+			n.CreateImplicitDirInBucket(ctx, ImplicitDirsPath, bucketName, n.EnableZB)
```
gargnitingoogle: If we're always passing n.EnableZB as the last parameter, why pass it ? Why not directly use n.EnableZB in the function implementation ?

gargnitingoogle: never mind. I saw a different value being passed at a later place.

gargnitingoogle (commented):
Added comments

pkg/cloud_provider/storage/storage.go
```diff
@@ -291,3 +301,33 @@ func ParseErrCode(err error) codes.Code {
 
 	return code
 }
+
+// UploadGcsObject uploads a local file to a specified GCS bucket and object.
+// Handles gzip compression if requested.
+func (service *gcsService) UploadGcsObject(ctx context.Context, filePathToUpload, bucketName, objectName string, zbEnabled bool) error {
```
amacaskill: nit: this should be UploadGCSObject as GCS is an initialism  go/go-style/decisions#initialisms

test/e2e/testsuites/subpath.go
```diff
@@ -135,8 +135,14 @@ func (t *gcsFuseCSISubPathTestSuite) DefineTests(driver storageframework.TestDri
 		bucketName := l.config.Prefix
 
 		// Create sub-paths using gsutil
-		specs.CreateImplicitDirInBucket("subpath1", bucketName)
-		specs.CreateImplicitDirInBucket("subpath2", bucketName)
+		// Passing zb as false because this test does not run with HNS enabled buckets and ZB is only relevant when HNS is enabled.
+
+		if driver, ok := driver.(*specs.GCSFuseCSITestDriver); ok {
+			driver.CreateImplicitDirInBucket(ctx, "subpath1", bucketName, false)
```
amacaskill: optional, but I like to put a comment explaining what the random boolean is so its more clear to the reader: 

 driver.CreateImplicitDirInBucket(ctx, "subpath1", bucketName, false /* enableZB */)  so its clear

pkg/util/goclientobjectcommands/file_operations.go
```diff
@@ -0,0 +1,44 @@
...
+func OpenFileAsReadonly(filepath string) (*os.File, error) {
+	f, err := os.OpenFile(filepath, os.O_RDONLY|syscall.O_DIRECT, FilePermission_0400)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file %s as readonly: %v", filepath, err)
+	}
+
+	return f, nil
+}
+
+// Deprecated: please use CloseFileShouldNotThrowError instead.
```
amacaskill: We don't need to change this now, so maybe just add a TODO to change to CloseFileShouldNotThrowError or if there is a reason why we can't maybe just remove the `// Deprecated: please use CloseFileShouldNotThrowError instead.` .   But in the future, if a method is already deprecated, maybe just use the new method if its possible.

amacaskill (approved):
Left a couple nit comments,  please address before submitting, but other than that LGTM! Great job!

test/e2e/testsuites/subpath.go
chrisThePattyEater: done

PR #865, opened 2025-06-23
Make host network token server feature opt-in only via mount options

siyanshen:
**What type of PR is this?**
>
> /kind api-change
/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Adds opt-in mount option tests.
How I inject SA: set injectSAVol as true in OSS webhook.
How I mock manager driver: tag my image with “prow-gob-internal-boskos”

Success: hnw + inject sa + opt in

make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false E2E_TEST_FOCUS=should.successfully.mount.for.hostnetwork.enabled.pods STAGINGVERSION=ssy-prow-gob-internal-boskos REGISTRY=gcr.io/shensiyan-joonix](http://gcr.io/shensiyan-joonix) 

Failure: hnw + inject sa + not opt in

make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false E2E_TEST_FOCUS=should.fail.when.the.KSA.feature.is.not.opted.in STAGINGVERSION=ssy-prow-gob-internal-boskos REGISTRY=gcr.io/shensiyan-joonix

Failure: hnw + inject insufficient sa: test case already exists

Fixes #

```release-note

```


pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -63,8 +63,8 @@ func New(mounterPath string) *Mounter {
 
 func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	// Start the token server for HostNetwork enabled pods.
-	if mc.TokenServerIdentityProvider != "" {
-		klog.Infof("Pod has hostNetwork enabled and token server feature is turned on. Starting Token Server on %s/%s", mc.TempDir, TokenFileName)
+	if mc.TokenServerIdentityProvider != "" && mc.HostNetworkKSAOptIn {
```
saikat-royc: these are CSI specific knobs which typically get captured in CSI volume attributes. While mount options are meant for gcsfuse configurations.

Given that feature is rolled back, my recommendation is to use csi volume attributes for tokenServerIdentityProvider and usePodKSAForHostNetwork.



Changes needed:
Add support to parse these new flags `tokenServerIdentityProvider` and `usePodKSAForHostNetwork` in 
[parseVolumeAttributes](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/utils.go#L193C6-L193C27)


Sample usage of volumeAttributes [skipCSIBucketAccessCheck](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-perf#reduce-quota-consumption)

Thoughts @siyanshen ?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -63,8 +63,8 @@ func New(mounterPath string) *Mounter {
...
+		klog.Infof("Pod has hostNetwork enabled and token server feature is supported and opted in. Starting Token Server on %s/%s", mc.TempDir, TokenFileName)
```
saikat-royc: klog.V().Infof()

pkg/sidecar_mounter/sidecar_mounter.go
saikat-royc, line 66: for the scenario where user does not explicitly provide the pod KSA config, we still need to let mount succeed, else we will break such customers who are already using node SA + pod host network

hime: Since hostnetwork support is a pod level feature, would it make more sense to add this override in pod leve annotations? Doesn't make sense to have multiple volumes assigned to a single pod have this flag set.

Not only that, but the behavior with this flag approach is misleading. If a pod with multiple GCSFuse volumes is running in hostnetwork, how does this volumeattribute get applied?

siyanshen: > the behavior with this flag approach is misleading. If a pod with multiple GCSFuse volumes is running in hostnetwork, how does this volumeattribute get applied?

That will result in multiple volumes, each with their own volume attribute configuration. No?

> Since hostnetwork support is a pod level feature, would it make more sense to add this override in pod leve annotations? Doesn't make sense to have multiple volumes assigned to a single pod have this flag set.

Why not? What if a pod needs to mount a gcsfusecsi volume and a filestorecsi volume, then the ksa opt in flag will be obviously irrelevant to the filestorecsi volume. It makes sense to keep this at a volume level granularity.

siyanshen: @saikat-royc sounds like it involves a bit of cosmetic change. I will keep using mount option internally .

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -35,11 +35,12 @@ import (
...
-	TempDir              = "/temp-dir"
-	unixSocketBasePath   = "unix://"
-	TokenFileName        = "token.sock" // #nosec G101
-	identityProviderFlag = "token-server-identity-provider"
+	GCSFuseAppName          = "gke-gcs-fuse-csi"
+	TempDir                 = "/temp-dir"
+	unixSocketBasePath      = "unix://"
+	TokenFileName           = "token.sock" // #nosec G101
+	identityProviderFlag    = "token-server-identity-provider"
+	hostNetworkKSAOptInFlag = "hnw-ksa"
```
saikat-royc, line 43: hostNetworkPodKSAFlag = host-network-pod-ksa

pkg/sidecar_mounter/sidecar_mounter.go
saikat-royc: +1 to volume level option (and csi volume attribute). If a pod uses multiple volumes, it makes sense to have the config per volume. By the way I think we should think though if the current implementation works for multiple volumes for a pod using hostnetwork and add a test case for that. e.g Our gcsfuse best practice calls out using 2 unique volume - one for reads and one for writes like checkpoints.

siyanshen: Done refactoring. Also tested two volume cases- one with PV one with ephemeral. Worked as expected. Details in PR description.

pkg/sidecar_mounter/sidecar_mounter_config.go
siyanshen: Reworded the customer facing flag in Volume Attributes

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: Is it ok if V(4)? I don't expect this to be logged more than once per pod

pkg/csi_driver/utils.go
```diff
@@ -495,7 +509,7 @@ func isSidecarVersionSupportedForGivenFeature(imageName string, sidecarMinSuppor
 	isManagedSidecar := re.MatchString(imageName)
 
 	if !isManagedSidecar {
-		klog.Infof("mountOptions should not be passed because this is a private sidecar image %q", imageName)
+		klog.Infof("mountOptions should not be passed unless identity provider is specified by user, because this is a private sidecar image %q", imageName)
```
saikat-royc: klog.Warningf

saikat-royc: this function isSidecarVersionSupportedForGivenFeature  is also used for other feature checks https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L357

We can make the klog message generic, "{sidecarMinSupportedVersion} check skipped since this {imageName} is not a GKE managed image"

pkg/csi_driver/node.go
```diff
@@ -138,9 +138,14 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		return nil, status.Errorf(codes.NotFound, "failed to get pod: %v", err)
 	}
 
-	if s.shouldStartTokenServer(pod) && pod.Spec.HostNetwork {
-		identityProvider := s.driver.config.TokenManager.GetIdentityProvider()
-		fuseMountOptions = joinMountOptions(fuseMountOptions, []string{"token-server-identity-provider=" + identityProvider})
+	if s.shouldPopulateIdentifyProvider(pod, optInHostnetworkKSA, userSpecifiedIdentityProvider != "") {
+		identityProvider := ""
```
saikat-royc, line 142: minor nit (avoid if else):

```
identityProvider = s.driver.config.TokenManager.GetIdentityProvider()
if userSpecifiedIdentityProvider != "" {
			identityProvider = userSpecifiedIdentityProvider
}
fuseMountOptions = ...
```

test/e2e/utils/cluster.go
```diff
@@ -29,8 +29,9 @@ import (
 )
 
 var (
-	nativeSidecarMinimumVersion                 = version.MustParseGeneric("1.29.0")
-	saTokenVolInjectionMinimumVersion           = version.MustParseGeneric("1.33.0")
+	nativeSidecarMinimumVersion = version.MustParseGeneric("1.29.0")
+	// TODO(@siyanshen): to enable hostnetwork tests, update min version when GCW feature flag is on.
```
saikat-royc: What about the e2e test coverage with the OSS CSI driver and OSS webhook (i.e testParams.UseGKEManagedDriver = false)? Without this coverage, we will miss out on PR presubmits catching regression early. 

We can now test the oss driver with identityProvider, and hostNetworkPodKSA combination

pkg/csi_driver/node.go
siyanshen: I want to avoid calling unnecessary functions to minimize weird bugs

test/e2e/utils/cluster.go
siyanshen: Modify the code so that we do not skip OSS driver.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -63,9 +63,14 @@ func New(mounterPath string) *Mounter {
 
 func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	// Start the token server for HostNetwork enabled pods.
-	if mc.TokenServerIdentityProvider != "" {
-		klog.Infof("Pod has hostNetwork enabled and token server feature is turned on. Starting Token Server on %s/%s", mc.TempDir, TokenFileName)
-		go StartTokenServer(ctx, mc.TempDir, TokenFileName, mc.TokenServerIdentityProvider)
+	// For managed sidecar, the token server identity provider is only populated when host network pod ksa feature is opted in.
```
saikat-royc, line 66: this comment can be removed. since now a user can also provide an identity provider with a managed sidecar version also, 

```
shouldPopulateIdentifyProvider() {
   ...
   tokenVolumeInjected && (sidecarVersionSupported || userInput) // returns true
}
```
```
	if s.shouldPopulateIdentifyProvider(pod, optInHostnetworkKSA, userSpecifiedIdentityProvider != "") {
		identityProvider := ""
siyanshen marked this conversation as resolved.
		if userSpecifiedIdentityProvider != "" {
			identityProvider = userSpecifiedIdentityProvider
		} else {
			identityProvider = s.driver.config.TokenManager.GetIdentityProvider()
		}
		fuseMountOptions = joinMountOptions(fuseMountOptions, []string{"hnw-ksa=true", "token-server-identity-provider=" + identityProvider})
	}
```

saikat-royc (approved):

PR #867, opened 2025-06-30
Update token server sidecar min version

siyanshen:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

We need to specify the minimum sidecar version that supports the host network ksa token feature. It is needed because sidecars with versions lower than the min is not compatible with the (internal only) mount options.

This PR also skips OSS tests for clusters without support for native sidecars.

Fixes #

Ran e2e tests on 1.28 cluster and confirmed it is not failing any more.
```
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=true E2E_TEST_FOCUS=should.successfully.mount.for.hostnetwork.enabled.pods STAGINGVERSION=prow-gob-internal-boskos-0630-4 REGISTRY=gcr.io/shensiyan-joonix
```

STEP: Skipping the hostnetwork test for cluster version < 1.100.0 or cluster without init container support @ 06/30/25 22:05:22.694

```release-note
NA
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -139,6 +139,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	if s.shouldPopulateIdentifyProvider(pod, optInHostnetworkKSA, userSpecifiedIdentityProvider != "") {
+		klog.V(4).Infof("NodePublishVolume populating identity provider in mount options")
 		identityProvider := ""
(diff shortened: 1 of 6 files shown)


hime (approved):

hime:
we should verify guitar passes before submit/release

PR #896, opened 2025-07-23
Add filter for killed process from GCSFuse.

halimsam:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
GCSFuse by default is always force-killed in GKE environment. We want to degrade the severity of the error log caused by killed signal to a warning log.
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #907, opened 2025-07-30
ZB Test refactoring

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test
> /kind feature
> /kind flake

This pr implements a change relating to failing tests introduced during the enablement of gcsfusexzonal buckets tests.

The change introduced in commit -  7753a057a0dce11582d8453414712a99d7bd26e0
This commit disabled the only-dir mount option (a flag used to communicate to gcsfuse that we have mounted our bucket in a subdirectory) for zb tests, this is needed because as mentioned in this ticket gcsfuse team does not currently support it 
https://github.com/GoogleCloudPlatform/gcsfuse/pull/3612

These changes have been tested both manually as mentioned above and by running the e2e tests for "gcsfuseIntegration.should.succeed.in.imp
licit_dir" both with gcsfuse zb enabled and disabled

### DIFF
--- a/test/e2e/specs/testdriver.go
+++ b/test/e2e/specs/testdriver.go
@@ -214,8 +214,12 @@ func (n *GCSFuseCSITestDriver) CreateVolume(ctx context.Context, config *storage
 			mountOptions += ",implicit-dirs"
 		case SubfolderInBucketPrefix:
 			dirPath := uuid.NewString()
-			n.CreateImplicitDirInBucket(ctx, dirPath, bucketName)
-			mountOptions += ",only-dir=" + dirPath
+
+			//Temporarily disable only-dir mount option zonal bucket tests as it is not supported in gcsfuse tests yet.
+			if !n.EnableZB {
+				n.CreateImplicitDirInBucket(ctx, dirPath, bucketName)
+				mountOptions += ",only-dir=" + dirPath
+			}
 		case EnableFileCachePrefix, EnableFileCacheForceNewBucketPrefix:
 			v.fileCacheCapacity = "100Mi"
 		case EnableFileCacheAndMetricsPrefix, EnableFileCacheForceNewBucketAndMetricsPrefix:


pkg/csi_driver/utils.go
```diff
@@ -447,6 +447,11 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 			code = codes.NotFound
 		}
 
+		// Special case for harmless error message from gcsfuse stderr.
+		if strings.Contains(errMsgStr, "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.") {
+			return codes.OK, fmt.Errorf("gcsfuse failed with harmless error: %v", errMsgStr)
```
amacaskill: Returning codes.OK and an error at the same time doesn't make sense because it will just be ignored and not even logged. 

```
		code, err := checkGcsFuseErr(isInitContainer, pod, targetPath)
		if code != codes.OK {
			if code == codes.Canceled {
				klog.V(4).Infof("NodePublishVolume on volume %q to target path %q is not needed because the gcsfuse has terminated.", bucketName, targetPath)

				return &csi.NodePublishVolumeResponse{}, nil
			}

			return nil, status.Error(code, err.Error())
		}

```

We either should return a different error code + error wrapped in a different message, or return OK then log the error. As for wrapping the error, I think "gcsfuse failed with harmless error" because that is misleading. GCSFuse didn't fail, this is just another error written to standard error, correct? Its not what caused gcsfuse to fail, and gcsfuse could still be running. Also consider explaining why the error can be ignored rather than saying harmless. 

Can we filter this out to a log similar to what Sam did with the `signal:killed` in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/896. If we don't do that, I think the sidecar will still log this as an error over and over and confuse the user. 

After we make those changes and decide what error code or how to handle this, I would also like us to test this manually to see what the customer will see when they do kubectl describe on the Pod. If it still looks odd, we will need to add this to our playbooks for launch. 

Please consider putting these in separate PRS. The error filtering requires cherry picking, but the test fix does not.

amacaskill (approved):

PR #908, opened 2025-07-31
Downgrade bootstrap config error to warning

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test
> /kind feature
> /kind flake

This pr implements two changes that both relate to failing tests introduced during the enablement of gcsfusexzonal buckets tests.

This pr downgrades the error "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables." which is logged in stderr from an error level notification. This is needed because during manual testing of gcsfue with zb (following the [public docs](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-ephemeral) but using a zb) we observed that after the initial success of node publish volume subsequent calls resulted in NPV failing due to the previously mentioned error. This can be seen below
```
ERROR 2025-07-25T16:37:47.343806891Z [resource.labels.containerName: gcs-fuse-csi-driver] /csi.v1.Node/NodePublishVolume failed with error: rpc error: code = Internal desc = gcsfuse failed with error: 2025/07/25 16:37:44 ERROR: [xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.
```

After implementing this change we no longer see the failure and the multiple NPV calls from kubelet succeed v
image

**Note that althought the NPV no longer fails one of our goals was to completely downgrade this error, it has been observed that even when downgrading the log in the sidecar_mounter.go file similar to the work done here https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/900, we still encounter the issue while mounting the volume. This is because this filtering is only taken into consideration when the error being filtered causes the gcsfuse process to fail, in this case the error is benign so we still encounter it - 
image

### DIFF
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -447,6 +447,11 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 			code = codes.NotFound
 		}
 
+		// Special case for harmless error message from gcsfuse stderr, gcsfuse cannot mute this because it stems from a needed library.
+		if strings.Contains(errMsgStr, "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.") {
+			return codes.Unavailable, fmt.Errorf("benign grpc error: %s", errMsgStr)
+		}
+
 		return code, fmt.Errorf("gcsfuse failed with error: %v", errMsgStr)
 	}
 


pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -149,6 +149,9 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 			errMsg := fmt.Sprintf("gcsfuse exited with error: %v\n", err)
 			if strings.Contains(errMsg, "signal: terminated") {
 				klog.Infof("[%v] gcsfuse was terminated.", mc.VolumeName)
+			} else if strings.Contains(errMsg, "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.") {
+				// Special case for harmless error message from gcsfuse stderr, gcsfuse cannot mute this because it stems from a needed library.
+				klog.Infof("%s", "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.")
```
amacaskill: Let's remove this, and put a sentence in the PR  description as to why it doesn't work.  I think you testing this out helped our understanding about what errors show up where.

pkg/csi_driver/utils.go
```diff
@@ -447,6 +447,11 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 			code = codes.NotFound
 		}
 
+		// Special case for harmless error message from gcsfuse stderr, gcsfuse cannot mute this because it stems from a needed library.
+		if strings.Contains(errMsgStr, "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.") {
+			return codes.Unavailable, nil
```
amacaskill: We should not be returning nil here. We discussed that we want to set the code to Unavailable, then wrap this error message str in something that indicates this is not a critical error. then return Unavailable and the error.

pkg/csi_driver/utils.go
```diff
@@ -447,6 +447,11 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 			code = codes.NotFound
 		}
 
+		// Special case for harmless error message from gcsfuse stderr, gcsfuse cannot mute this because it stems from a needed library.
+		if strings.Contains(errMsgStr, "[xds] Attempt to set a bootstrap configuration even though one is already set via environment variables.") {
+			return codes.Unavailable, fmt.Errorf("Benign GCSFUSE Error: %v", errMsgStr)
```
amacaskill: Nit: please make sure we conform with  go/go-style/decisions#error-strings

fmt.Errorf("Benign GCSFUSE Error: %v", errMsgStr) -> fmt.Errorf("benign grpc error: %s", errMsgStr)

PR #920, opened 2025-08-07
Adding go client updates and increasing sidecar limits for ZB testing

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test
> /kind feature
> /kind flake

This pr introduces 2 changes to our tests that are both being added for the purpose of fixing failing tests in our new rapid storage test grid. The first is related to flakey ooms that occur on our gcsfuse sidecar, in the past to get around the increased memory utilization of grpc we have increased the requests and limits. We are doing the same here to validate that the issue is simply a matter of increasing resources. 

The second change is additional required flags that were recently introduced as part of gcsfuses go client that is used to perform actions on zonal buckets. We are using the same workflow to perform actions on zb on our end because zb still do not support gcloud commands. So to stay in sync we are adding the bug fixing change. The mentioned changes that are being mirrored are here - https://github.com/GoogleCloudPlatform/gcsfuse/pull/3491

Fixes #

```release-note

```


amacaskill:
Your PR is setting requests not changing limits. Also, only increase what you are increasing for ZB if that is really needed not the rest of the tests. But before we do this, please run the test in isolation, record how much memory it is using on grpc vs non ZB. Ask GCSFuse is this is expected for that test, or if they aren't cleaning up files somewhere

amacaskill:
Please update the description and the title,  to only describe the goclient change

amacaskill (approved):

PR #953, opened 2025-08-27
Bucket scanner controller - PersistentVolume creation trigger

uriel-guzman:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR builds on top of the existing Bucket Scanner controller skeleton (https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/948) within the GCSFuse CSI driver. This controller watches PersistentVolumes (PVs) managed by `gcsfuse.csi.storage.gke.io` and, when relevant, performs simulated metadata scans on the associated GCS buckets. The results are stored as annotations on the PV.

Key features implemented in this PR:

*   **PV Event Handling**: Reacts to Add and Delete PV events.
*   **Relevance Checks**:
    *   Ensures the PV uses the `gcsfuse.csi.storage.gke.io` driver.
    *   Verifies the StorageClass has a valid `workloadType` parameter (`inference`, `training`, or `checkpointing`).
    *   Implements a Time-To-Live (TTL) logic: re-scans only if the time since the `bucket-scan-last-updated-time` annotation exceeds the configured TTL.
*   **Configurable Scan Behavior**:
    *   Scan Timeout and TTL can be customized via PV VolumeAttributes (`bucketScanTimeout`, `bucketScanTTL`). Defaults are 2 minutes and 7 days, respectively.
*   **Simulated Bucket Scan**:
    *   The current version simulates the scanning process. It does not yet interact with Google Cloud Storage.
    *   It respects context cancellation and the configured scan timeout, returning partial results upon timeout.
*   **PV Annotation Updates**:
    *   Patches the PV with annotations prefixed `gke-gcsfuse/bucket-scan-`:
        *   `status` (e.g., `completed`, `timeout`)
        *   `num-objects`
        *   `total-size-bytes`
        *   `hns-enabled`
        *   `last-updated-time`
*   **Kubernetes Events**: Emits events to the PV to indicate scan status and errors.
*   **Unit Tests**: Expands the existing test suite (`scanner_test.go`) to validate:
    *   PV relevance logic (including TTL).
    *   Queue processing and error handling in `syncPV`.
    *   PV event handlers (add/delete).
    *   Annotation patching.
    *   Timeout/TTL parsing from VolumeAttributes.

**Manual Verification**:
1. Scan timeout w/ custom `bucketScanTTL` and `bucketScanTimeout` volumeAttributes (this will always time out because the mock infinetely scans for objects) :
```
Name:            my-pv
Labels:          <none>
[...]

PR #955, opened 2025-09-04
Modify pod spec with gcsfuse profiles features if enabled

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This pr introduces storage profile labels and placeholder volumes to the oss webhook that are critical to our storage profile features. 

The main changes involved lie in two methods within Mutatingwbhook.go, the two methods are for identifying pod create requests that involve k8s resources that have sp enabled (), and a secondary method that applies all the required changes needed to support gcsfuse profiles if it IS enabled.

IsStorageProfilesEnabled
check if profiles annotation has been applied

ModifyPodSpecForStorageProfiles
This method performs all modifications to the gcsfuse profiles pod spec, if profiles are enabled all of these modifications are applied
1. the label gcsfuse.gke.io/profiles is set to true, this is used to identify the pod as a pod that uses gcsfuse profiles
2. the scheduling gate gke-gcsfuse/bucket-scan-pending is applied to prevent the pod from being scheduled before the bucket scan is completed as the results from the bucket scan are necessary to set best practice settings on cache sizes
3. volumes and volume mounts are added for ram disk file cache, and ephemeral disk file cache these are placeholder volumes to be used based on used selected file cache backing type (ssd vs ram)

This change will be a noop for all pod creation requests that do not involve gcsfuse profile enabled resources

How was this tested
There were three cases tested along with running e2e tests
Using a standard cluster using 3 nodes and n2-standard-8 we deployed 3 pods with different pv/pvc/pv/sc configurations
case 1: 
User creates a pod with no gcsfuse profiles annotation
result: pod follows regular flow
case 2:
User creates a pod with profiles annotation
result: pod is not scheduled due to scheduling gate
case 3: 
User creates a pod with random profile
result: pod follows regular flow
case 4:
pod has profiles annotation but has invalid value 
result: 
[hi on] fuechr@fuechr:~/Desktop/projects/gcs-fuse-csi-driver$ kubectl apply -f pod.yaml
Error from server: error when creating "pod.yaml": admission webhook "gcsfuse-sidecar-injector.csi.storage.gke.io" denied the request: the acceptable values for "gke-gcsfuse/profiles" are 'True', 'true', 'false' or 'False'

here is the link to the pod spec for a the final case (1 sp pvc and 1 non sp pvc) - https://paste.googleplex.com/6145033350807552

overwrites for profiles overlay 

and lastly unit tests were added for the mutating webhook 

Fixes #

```release-note

```


pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		}
+	}
+	for i, c := range pod.Spec.Containers {
+		if c.Name == "gke-gcsfuse-sidecar" {
+			pod.Spec.Containers[i].VolumeMounts = append(pod.Spec.Containers[i].VolumeMounts, epehemeralMount, ramMount)
+		}
+	}
+}
+
+// Checks if the pod has ANY pvc backed by a storage profile enabled storage class. storage profile enabled means the sc uses a valid workloadtype parameter.
```
uriel-guzman: workloadType. Also, we're not calling this feature storage profiles anymore. It's called "gcsfuse profiles" now.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	// Check if pod has volumes
+	if pod.Spec.Volumes == nil {
+		return false
+	}
+
+	for _, vol := range pod.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			pvcName := vol.PersistentVolumeClaim.ClaimName
+
+			// Fetch the PVC from API server
```
uriel-guzman: It's not technically fetched from the API server if you're using a StorageClass informer. Let's fix this comment so it's not misleading.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+			pvc, err := si.GetPVC(pod.Namespace, pvcName)
+			if err != nil {
+				fmt.Printf("failed to get PVC %s in ns %s: %v", pvcName, pod.Namespace, err)
```
uriel-guzman: ns -> namespace

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +128,104 @@
...
+			if ok {
+				if slices.Contains(workloadTypes, workloadType) {
+					return true
+				} else {
+					fmt.Printf("StorageClass %s for pvc %s has unsupported workloadType %s", *pvc.Spec.StorageClassName, pvcName, workloadType)
+					continue
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
uriel-guzman: I agree with this insight. Let's switch to `klog`.

pkg/webhook/mutatingwebhook.go
uriel-guzman: Add name of the Pod in log.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+			if pvc.Spec.StorageClassName == nil {
+				// If the PVC does not have a storage class, skip it.
+				continue
+			}
+			sc, error := si.GetSC(*pvc.Spec.StorageClassName)
+			if error != nil {
+				fmt.Printf("failed to get StorageClass %s for pvc %s : %v", *pvc.Spec.StorageClassName, pvcName, error)
+				continue
+			}
+			workloadType, ok := sc.Parameters["workloadType"]
```
uriel-guzman: Make workloadType a constant.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+			if ok {
+				if slices.Contains(workloadTypes, workloadType) {
+					return true
+				} else {
+					fmt.Printf("StorageClass %s for pvc %s has unsupported workloadType %s", *pvc.Spec.StorageClassName, pvcName, workloadType)
```
uriel-guzman: This is a customer misconfiguration. I think it would be better if the Pod creation was blocked by this and the customer fixed their Pod spec.

pkg/webhook/mutatingwebhook_test.go
```diff
@@ -946,3 +949,389 @@ func skewVersionNodes() []corev1.Node {
...
+				},
+				Spec: corev1.PodSpec{
+					Volumes: []corev1.Volume{
+						{Name: "vol1", VolumeSource: corev1.VolumeSource{PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{ClaimName: "pvc1"}}},
+						{Name: "vol2", VolumeSource: corev1.VolumeSource{PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{ClaimName: "pvc2"}}},
+					},
+				},
+			},
+			spEnabled: true,
+		},
```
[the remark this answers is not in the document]
uriel-guzman: +1 please remove

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+
 	marshaledPod, err := json.Marshal(pod)
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to marshal pod: %w", err))
 	}
 
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
+
+// Modifies the pod spec to add storage profile related features. This includes adding a label, scheduling gate, and placeholder file cache volumes.
```
uriel-guzman: gcsfuse profile

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+func (si *SidecarInjector) ModifyPodSpecForStorageProfiles(ctx context.Context, pod *corev1.Pod) {
```
uriel-guzman: gcsfuse profile

uriel-guzman: Here an everywhere else

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	// Always apply the storage profile label when the storage profile is enabled for pod informer's Kubernretes API efficient filtering.
```
uriel-guzman: Kubernetes, also gcsfuse profile

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	if pod.Labels == nil {
+		pod.Labels = make(map[string]string)
+	}
+	pod.Labels["gcsfuse.gke.io/profile-managed"] = "true"
```
uriel-guzman: gke-gcsfuse/profile-managed

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+
+	// Always apply the scheduling gate when the storage profile is enabled. The contoroller will handle the logistics if its not needed.
```
uriel-guzman: controller, gcsfuse profile

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	spGate := corev1.PodSchedulingGate{Name: "gke-gcsfuse/bucket-scan-pending"}
```
uriel-guzman: Make gke-gcsfuse/bucket-scan-pending a const.

pkg/webhook/mutatingwebhook.go
uriel-guzman: Also, make that a const.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	pod.Spec.SchedulingGates = append(pod.Spec.SchedulingGates, spGate)
+
+	// Inject placeholder file cache volumes.
+	pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
```
uriel-guzman: How does this handle the case where these volumes already exist? Ideally, that should be a no-op. It seems like here it would duplicate those volumes.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		Name:         "gcsfuse-file-cache-ephemeral-disk",
```
uriel-guzman: Make const.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+	})
+	pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
+		Name: "gcsfuse-file-cache-ram-disk",
```
uriel-guzman: Make const.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	ramMount := corev1.VolumeMount{
+		Name:      "gcsfuse-file-cache-ram-disk",
+		MountPath: "/mnt/data/gcsfuse-file-cache-ram-disk",
+	}
+	epehemeralMount := corev1.VolumeMount{
+		Name:      "gcsfuse-file-cache-ephemeral-disk",
+		MountPath: "/mnt/data/gcsfuse-file-cache-ephemeral-disk",
+	}
+	for i, c := range pod.Spec.InitContainers {
+		if c.Name == "gke-gcsfuse-sidecar" {
```
uriel-guzman: Make all of these strings consts.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +133,105 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+			pod.Spec.InitContainers[i].VolumeMounts = append(pod.Spec.InitContainers[i].VolumeMounts, epehemeralMount, ramMount)
```
uriel-guzman: Could we check for duplicates here?

pkg/webhook/mutatingwebhook.go
```diff
@@ -63,16 +67,13 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	if req.Operation != admissionv1.Create {
-		return admission.Allowed(fmt.Sprintf("No injection required for operation %v.", req.Operation))
-	}
-
 	enableGcsfuseVolumes, ok := pod.Annotations[GcsFuseVolumeEnableAnnotation]
 	if !ok {
 		return admission.Allowed(fmt.Sprintf("The annotation key %q is not found, no injection required.", GcsFuseVolumeEnableAnnotation))
```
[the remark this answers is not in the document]
uriel-guzman: This looks unintentional. Could you pleae fix.

pkg/webhook/client_test.go
```diff
@@ -413,3 +414,87 @@ func TestGetPVC(t *testing.T) {
...
+				t.Error("for test: ", testcase.testName, ", want: ", testcase.expectedError.Error(), " but got: ", err.Error())
+			}
+		} else if err != nil || testcase.expectedError != nil {
+			// if one of them is nil, both must be nil to pass
+			t.Error("for test: ", testcase.testName, ", want: ", testcase.expectedError, " but got: ", err)
+		}
+
+		if response.String() != testcase.expectedResponse.String() {
+			t.Error("for test: ", testcase.testName, ", want: ", testcase.expectedResponse, " but got: ", response)
+		}
```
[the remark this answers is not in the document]
uriel-guzman: Please address.

pkg/webhook/mutatingwebhook.go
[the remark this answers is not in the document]
chrisThePattyEater: this was already fixed not sure why it still shows up here

pkg/webhook/mutatingwebhook.go
chrisThePattyEater: tested and confirmed [hi on] fuechr@fuechr:~/Desktop/projects/gcs-fuse-csi-driver$ kubectl apply -f pod.yaml
Error from server: error when creating "pod.yaml": admission webhook "gcsfuse-sidecar-injector.csi.storage.gke.io" denied the request: failed to check gcsfuse profiles support: StorageClass: gcsfusecsi-inference for pvc: my-pvc2 has unsupported workloadType: bla

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
 	ephemeralStorageRequestAnnotation       = "gke-gcsfuse/ephemeral-storage-request"
 	metadataPrefetchMemoryLimitAnnotation   = "gke-gcsfuse/metadata-prefetch-memory-limit"
 	metadataPrefetchMemoryRequestAnnotation = "gke-gcsfuse/metadata-prefetch-memory-request"
+
+	//gcsfuse profiles constants
+	gcsfuseProfilesManaged        = "gke-gcsfuse/profile-managed"
```
uriel-guzman: gcsfuseProfilesManaged -> gcsfuseProfilesManagedLabel

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
 	ephemeralStorageRequestAnnotation       = "gke-gcsfuse/ephemeral-storage-request"
 	metadataPrefetchMemoryLimitAnnotation   = "gke-gcsfuse/metadata-prefetch-memory-limit"
 	metadataPrefetchMemoryRequestAnnotation = "gke-gcsfuse/metadata-prefetch-memory-request"
+
+	//gcsfuse profiles constants
```
uriel-guzman: Space after `//`

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	gcsfuseProfilesManaged        = "gke-gcsfuse/profile-managed"
+	bucketScanPending             = "gke-gcsfuse/bucket-scan-pending"
```
uriel-guzman: bucketScanPending -> bucketScanPendingSchedulingGate

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	fileCacheEphemeralDisk        = "gcsfuse-file-cache-ephemeral-disk"
```
uriel-guzman: fileCacheEphemeralDisk -> fileCacheEphemeralDiskVolumeName

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	fileCacheEphemeralDiskMntPath = "/mnt/data/gcsfuse-file-cache-ephemeral-disk"
```
uriel-guzman: Why are we mounting on `/mnt/data`? Does that even exist?

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	fileCacheRamDisk              = "gcsfuse-file-cache-ram-disk"
```
uriel-guzman: fileCacheRamDisk -> fileCacheRamDiskVolumeName

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	fileCacheRamDiskMntPath       = "/mnt/data/gcsfuse-file-cache-ram-disk"
```
uriel-guzman: Same question for `/mnt/data`

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	gkeGcsfuseSidecar             = "gke-gcsfuse-sidecar"
```
uriel-guzman: That already exists here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L29. No need to re-declare the same const.

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
+	workloadType                  = "workloadType"
```
uriel-guzman: workloadType -> workloadTypeParamKey

pkg/webhook/mutatingwebhook.go
```diff
@@ -44,8 +46,20 @@ const (
...
 )
 
+var workloadTypes = []string{"training", "checkpointing", "inference"}
```
uriel-guzman: workloadTypes -> workloadTypeParamValues

pkg/webhook/mutatingwebhook.go
```diff
@@ -63,6 +78,7 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 	// Validate injection request
 	pod := &corev1.Pod{}
 
+	// TODO: check if request is a pv and if it is perform iam check
```
uriel-guzman: Can we be more detailed with this comment for a future reader? The PV IAM check is only relevant for the gcsfuse profiles feature. In the future, someone may be confused as to what this TODO actually is.

pkg/webhook/mutatingwebhook.go
uriel-guzman: In https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L32 we always mount to `/`, let's use that pattern.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +143,149 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		pod.Labels = make(map[string]string)
+	}
+	pod.Labels[gcsfuseProfilesManaged] = "true"
+
+	// Always apply the scheduling gate when the gcsfuse profiles are enabled. The controller will handle the logistics if its not needed.
+	profilesGate := corev1.PodSchedulingGate{Name: bucketScanPending}
+	pod.Spec.SchedulingGates = append(pod.Spec.SchedulingGates, profilesGate)
+
+	// Inject placeholder file cache volumes.
+	if !volumeExists(pod.Spec.Volumes, fileCacheEphemeralDisk) {
```
uriel-guzman: Can we log a warning if the volume already exists? This should ideally never happen, but in the rare occasion that the customer introduces a volume with a conflict, it would be very hard to debug without an informational message.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +143,149 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+			Name: fileCacheRamDisk,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{
+					Medium: corev1.StorageMediumMemory,
+				},
+			},
+		})
+	}
+
+	// Apply file cache volume mounts to side car container.
```
uriel-guzman: Sidecar volume mounts are defined here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L73. Let's keep this same pattern.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +143,149 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	pod.Labels[gcsfuseProfilesManaged] = "true"
+
+	// Always apply the scheduling gate when the gcsfuse profiles are enabled. The controller will handle the logistics if its not needed.
+	profilesGate := corev1.PodSchedulingGate{Name: bucketScanPending}
+	pod.Spec.SchedulingGates = append(pod.Spec.SchedulingGates, profilesGate)
+
+	// Inject placeholder file cache volumes.
+	if !volumeExists(pod.Spec.Volumes, fileCacheEphemeralDisk) {
+		pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
+			Name:         fileCacheEphemeralDisk,
```
uriel-guzman: Sidecar volumes are defined here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L51. Let's preserve the same pattern.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +143,149 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
 	}
 
+	isUsingGcsfuseProfiles, err := si.IsGCSFuseProfilesEnabled(ctx, pod)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to check gcsfuse profiles support: %w", err))
+	}
+
+	// Handle gcsfuse profiles spec modifications if using gcsfuse profile enabled buckets.
+	if isUsingGcsfuseProfiles {
+		si.ModifyPodSpecForGCSFuseProfiles(ctx, pod)
```
uriel-guzman: I think we should keep following this pattern: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/mutatingwebhook.go#L111, similar to host network. 

It populates a feature config and is controlled by a feature flag: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/cmd/webhook/main.go#L56.

We should make sure that none of these new code paths are exercised unless the feature flag is turned on, and should be turned off by default, until we are ready to launch.

uriel-guzman (commented):
Left several comments.

pkg/webhook/mutatingwebhook.go
uriel-guzman: Can we put these together with the other gcsfuse sidecar volume related vars? https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L28

deploy/overlays/profiles/webhook_deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+            - --sidecar-ephemeral-storage-limit=0
+            - --sidecar-ephemeral-storage-request=5Gi
+            - --sidecar-image=$(SIDECAR_IMAGE)
+            - --metadata-sidecar-image=$(METADATA_SIDECAR_IMAGE)
+            - --sidecar-image-pull-policy=$(SIDECAR_IMAGE_PULL_POLICY)
+            - --cert-dir=/etc/tls-certs
+            - --port=22030
+            - --health-probe-bind-address=:22031
+            - --should-inject-sa-vol=true
+            - --enable-gcsfuse-profiles=true
```
uriel-guzman: Address (missing blank empty line YAML).

deploy/overlays/profiles/webhook_setup.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: gcs-fuse-csi-webhook-role
+rules:
+  - apiGroups: [""]
+    resources: ["nodes", "persistentvolumes", "persistentvolumeclaims"]
+    verbs: ["get","list","watch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
```
uriel-guzman: Address (missing blank empty line YAML).

pkg/webhook/mutatingwebhook.go
```diff
@@ -63,6 +66,7 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 	// Validate injection request
 	pod := &corev1.Pod{}
 
+	// GCSFuse Profiles TODO: if request is a PV, validate user has applied iam rules required for gcsfuse profiles.
```
uriel-guzman: // TODO("your username"): If the request is a PV, fail the request if the user didn't apply the required IAM permissions for gcsfuse profiles.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +131,151 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+// Modifies the pod spec to add gcsfuse profile related features. This includes adding a label, scheduling gate, and placeholder file cache volumes.
+func (si *SidecarInjector) ModifyPodSpecForGCSFuseProfiles(ctx context.Context, pod *corev1.Pod) {
+	// Always apply the gcsfuse profile label when gcsfuse profiles are enabled for pod informer's Kubernetes API efficient filtering.
+	if pod.Labels == nil {
+		pod.Labels = make(map[string]string)
+	}
+	pod.Labels[gcsfuseProfilesManagedLabel] = "true"
+
+	// Always apply the scheduling gate when the gcsfuse profiles are enabled. The controller will handle the logistics if its not needed.
+	profilesGate := corev1.PodSchedulingGate{Name: bucketScanPendingSchedulingGate}
```
uriel-guzman: What would happen if the gate already exists? Ideally, this should be idempotent.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +131,151 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+				klog.Warningf("failed to get PVC %s in namespace %s while mutating pod %s: %v", pvcName, pod.Namespace, pod.Name, err)
+				continue
+			}
+			if pvc.Spec.StorageClassName == nil {
+				// If the PVC does not have a storage class, skip it.
+				continue
+			}
+			sc, error := si.GetSC(*pvc.Spec.StorageClassName)
+			if error != nil {
+				klog.Warningf("failed to get StorageClass %s for pvc %s : %v", *pvc.Spec.StorageClassName, pvcName, error)
```
uriel-guzman: Add comment that "dummy" StorageClasses should be handled gracefully (customers are allowed to specify non-existent StorageClasses during static provisioning because the spec requires PV and PVC SC fields to be equal).

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +131,151 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+				continue
+			}
+			wt, ok := sc.Parameters[workloadTypeParamKey]
+			if ok {
+				if slices.Contains(workloadTypeParamValues, wt) {
+					return true, nil
+				} else {
+					klog.Errorf("StorageClass: %s for pvc: %s has unsupported workloadType: %s", *pvc.Spec.StorageClassName, pvcName, wt)
```
uriel-guzman: Print the supported workloadType acceptable values.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +131,151 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+					return true, fmt.Errorf("StorageClass: %s for pvc: %s has unsupported workloadType: %s", *pvc.Spec.StorageClassName, pvcName, wt)
```
uriel-guzman: Same here.

pkg/webhook/sidecar_spec.go
```diff
@@ -41,6 +41,15 @@ const (
 	// Webhook relevant volume attributes.
 	gcsFuseMetadataPrefetchOnMountVolumeAttribute = "gcsfuseMetadataPrefetchOnMount"
 
+	// gcsfuse profiles constants
+	gcsfuseProfilesManagedLabel      = "gke-gcsfuse/profile-managed"
+	bucketScanPendingSchedulingGate  = "gke-gcsfuse/bucket-scan-pending"
+	fileCacheEphemeralDiskVolumeName = "gcsfuse-file-cache-ephemeral-disk"
+	fileCacheEphemeralDiskMntPath    = "/gcsfuse-file-cache-ephemeral-disk"
+	fileCacheRamDiskVolumeName       = "gcsfuse-file-cache-ram-disk"
+	fileCacheRamDiskMntPath          = "/gcsfuse-file-cache-ram-disk"
```
uriel-guzman: I see that the consts in the block above are allways of the format `SidecarContainer` ... `VolumeName|VolumeMountPath`. Can we also change these so that they are in sync?

pkg/webhook/sidecar_spec.go
```diff
@@ -90,6 +99,19 @@ var (
 		Name:      SidecarContainerSATokenVolumeName,
 		MountPath: SidecarContainerSATokenVolumeMountPath,
 	}
+
+	ramFileCacheMount = corev1.VolumeMount{
```
uriel-guzman: ramFileCacheMount -> `ramFileCacheVolumeMount` to respect the existing formats.

pkg/webhook/sidecar_spec.go
```diff
@@ -90,6 +99,19 @@ var (
...
+		Name:      fileCacheRamDiskVolumeName,
+		MountPath: fileCacheRamDiskMntPath,
+	}
+
+	epehemeralFileCacheMount = corev1.VolumeMount{
```
uriel-guzman: epehemeralFileCacheMount -> `ephemeralFileCacheVolumeMount` to respect the existing formats.

pkg/webhook/sidecar_spec.go
```diff
@@ -90,6 +99,19 @@ var (
...
+		Name:      fileCacheEphemeralDiskVolumeName,
+		MountPath: fileCacheEphemeralDiskMntPath,
+	}
+
+	// gcsfuse profiles constants
```
uriel-guzman: This is duplicated with line 44. Can we just add this to the same block as the other ones?

chrisThePattyEater: no you cant have const slices in go, I made it a string and split it when using it. Its a small const string so shouldn't be an issue when it comes to overhead

deploy/overlays/profiles/webhook_always_pull.yaml
```diff
@@ -25,4 +25,5 @@ spec:
           imagePullPolicy: Always
           env:
             - name: SIDECAR_IMAGE_PULL_POLICY
-              value: 'Always'
\ No newline at end of file
+              value: 'Always'
+
```
uriel-guzman: I still see this error. I think it may be because you have an empty space in L28 instead of an EOF char.

deploy/overlays/profiles/webhook_deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+
```
uriel-guzman: Same issue.

deploy/overlays/profiles/webhook_setup.yaml
```diff
@@ -0,0 +1,27 @@
...
+
```
uriel-guzman: Same issue.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +132,166 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	gateExists := false
+	for _, gate := range pod.Spec.SchedulingGates {
+		if gate.Name == profilesGate.Name {
+			gateExists = true
+			break
+		}
+	}
+
+	// Only append the gate if it does not already exist.
+	if !gateExists {
```
uriel-guzman, line 172: If the gate already exists, this is unexpected. Add warning to inform about the pre-existing gate.

pkg/webhook/sidecar_spec.go
```diff
@@ -41,6 +41,16 @@ const (
...
 
+	// gcsfuse profiles constants
+	GcsfuseProfilesManagedLabel                           = "gke-gcsfuse/profile-managed"
+	BucketScanPendingSchedulingGate                       = "gke-gcsfuse/bucket-scan-pending"
+	SidecarContainerFileCacheEphemeralDiskVolumeName      = "gcsfuse-file-cache-ephemeral-disk"
+	SidecarContainerFileCacheEphemeralDiskVolumeMountPath = "/gcsfuse-file-cache-ephemeral-disk"
+	SidecarContainerFileCacheRamDiskVolumeName            = "gcsfuse-file-cache-ram-disk"
+	SidecarContainerFileCacheRamDiskVolumeMountPath       = "/gcsfuse-file-cache-ram-disk"
+	WorkloadTypeParamKey                                  = "workloadType"
+	WorkloadTypeParamValues                               = "training checkpointing inference"
```
uriel-guzman: I don't think this is a good practice. Please use slices of strings as you had previously.

pkg/webhook/sidecar_spec.go
uriel-guzman: That's fine, I just meant to group the vars together, but I hadn't realized one was const and the other group were vars.

pkg/webhook/client_test.go
```diff
@@ -413,3 +414,99 @@ func TestGetPVC(t *testing.T) {
...
+			t.Fatalf("expected an error but got nil:\n\twant: %v", testcase.expectedError)
+		}
+
+		// Case: Both errors exist, but their content doesn't match.
+		if err != nil && testcase.expectedError != nil {
+			if err.Error() != testcase.expectedError.Error() {
+				t.Fatalf("error mismatch:\n\twant: %q\n\tgot:  %q", testcase.expectedError.Error(), err.Error())
+			}
+			// If errors match, we're done with this test case.
+			return
```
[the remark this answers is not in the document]
uriel-guzman: Please address.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +132,166 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+				continue
+			}
+			wt, ok := sc.Parameters[WorkloadTypeParamKey]
+			if ok {
+				workloadTypeParamValuesAsSlice := strings.Split(WorkloadTypeParamValues, " ")
+				if slices.Contains(workloadTypeParamValuesAsSlice, wt) {
+					return true, nil
+				} else {
+					klog.Errorf("StorageClass: %s for pvc: %s has unsupported workloadType: %s, the supported workloadTypes are: %s", *pvc.Spec.StorageClassName, pvcName, wt, WorkloadTypeParamValues)
+					return true, fmt.Errorf("StorageClass: %s for pvc: %s has unsupported workloadType: %s, the supported workloadTypes are: %s", *pvc.Spec.StorageClassName, pvcName, wt, WorkloadTypeParamValues)
```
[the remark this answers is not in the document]
uriel-guzman: I agree. This looks like a bug. The `IsGCSFuseProfilesEnabled` shoult not be returning `true` if the `workloadType` param has invalid values.

pkg/webhook/sidecar_spec.go
```diff
@@ -90,6 +100,16 @@ var (
 		Name:      SidecarContainerSATokenVolumeName,
 		MountPath: SidecarContainerSATokenVolumeMountPath,
 	}
+
+	ramFileCacheVolumeMount = corev1.VolumeMount{
+		Name:      SidecarContainerFileCacheRamDiskVolumeName,
+		MountPath: SidecarContainerFileCacheRamDiskVolumeMountPath,
+	}
+
+	epehemeralFileCacheVolumeMount = corev1.VolumeMount{
```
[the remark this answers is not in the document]
uriel-guzman: Fix typo.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +132,166 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		}
+	}
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name == GcsFuseSidecarName {
+			// Found the sidecar as a container, now add mounts safely
+			addMounts(&pod.Spec.Containers[i])
+			return
+		}
+	}
+	klog.Warning("Could not find gcsfuse sidecar container to add gcsfuse profile file cache mounts.")
```
uriel-guzman: Make this a final error and wrap it to return it to the mutator code.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +132,166 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	if pod.Spec.Volumes == nil {
+		return false, nil
+	}
+
+	for _, vol := range pod.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			pvcName := vol.PersistentVolumeClaim.ClaimName
+
+			// Fetch the PVC from storage class informer
+			pvc, err := si.GetPVC(pod.Namespace, pvcName)
```
uriel-guzman: Check if this error is a 404 error. If it's a 404 not found, "continue", otherwise, return error. Here, and same for the GetSC call.

uriel-guzman: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/953/files#diff-6624cc90057fa23225d6bc72235d4f18ccb633f0dcf24848c8849258ba6601dcR297

chrisThePattyEater:
@gemini-code-assist review

chrisThePattyEater:
@gemini-code-assist review

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +130,122 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+}
+
+// Checks if the pod has gcsfuse profiles annotation. returns true if the annotation is present and set to "true" (case insensitive), false otherwise.
+func IsGCSFuseProfilesEnabled(pod *corev1.Pod) bool {
+	// Check if pod has gcsfuse profiles annotation set to true.
+	if pod.Annotations == nil {
+		return false
+	}
+	value, ok := pod.Annotations[GcsfuseProfilesLabel]
+	return strings.EqualFold(strings.TrimSpace(value), "true") && ok
```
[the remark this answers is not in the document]
chrisThePattyEater: doing so would make this return true in the case user sets to false which is wrong

chrisThePattyEater:
@gemini-code-assist review

pkg/webhook/mutatingwebhook.go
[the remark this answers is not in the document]
uriel-guzman: This is an AND statement, not an OR. If the customer sets the value to "false", it would return "false", not "true", because both statements must be true.

I think Gemini's assessment is correct, you typically want to check if the value even exists before evaluating it. The way that Gemini suggests is better, because Golang will not evaluate the second argument if the first one is false.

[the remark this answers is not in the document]
chrisThePattyEater: oops you are right

pkg/webhook/mutatingwebhook.go
```diff
@@ -63,6 +65,7 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 	// Validate injection request
 	pod := &corev1.Pod{}
 
+	// TODO(fuechr): If the request is a PV, fail the request if the user didn't apply the required IAM permissions for gcsfuse profiles.
```
uriel-guzman, line 68: I don't think we want to do this anymore. That would require knowledge about the StorageClass in the PV mutation event, which may not exist yet, which will lead to the same race condition we had before. So let's just forget about this improvement, even in GCW.

pkg/webhook/sidecar_spec.go
```diff
@@ -41,6 +41,16 @@ const (
...
 
+	// gcsfuse profiles constants
+	GcsfuseProfilesLabel                                  = "gke-gcsfuse/profiles"
+	GcsfuseProfilesManagedLabel                           = "gke-gcsfuse/profile-managed"
+	BucketScanPendingSchedulingGate                       = "gke-gcsfuse/bucket-scan-pending"
+	SidecarContainerFileCacheEphemeralDiskVolumeName      = "gcsfuse-file-cache-ephemeral-disk"
+	SidecarContainerFileCacheEphemeralDiskVolumeMountPath = "/gcsfuse-file-cache-ephemeral-disk"
+	SidecarContainerFileCacheRamDiskVolumeName            = "gcsfuse-file-cache-ram-disk"
+	SidecarContainerFileCacheRamDiskVolumeMountPath       = "/gcsfuse-file-cache-ram-disk"
+	WorkloadTypeParamKey                                  = "workloadType"
```
uriel-guzman: We're not using this anymore. Delete.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +129,122 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	}
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name == GcsFuseSidecarName {
+			// Found the sidecar as a container, now add mounts safely
+			addMounts(&pod.Spec.Containers[i])
+			return nil
+		}
+	}
+	errString := fmt.Sprintf("Could not find gcsfuse sidecar container in pod %s/%s to add gcsfuse profile file cache mounts.", pod.Namespace, pod.Name)
+	klog.Warning(errString)
```
uriel-guzman: Make this error, not warning.

uriel-guzman: Also, don't wrap this in `errString`. The returned error should not be capitalized. The inline error should be capitalized.

pkg/webhook/sidecar_spec.go
```diff
@@ -41,6 +41,16 @@ const (
 	// Webhook relevant volume attributes.
 	gcsFuseMetadataPrefetchOnMountVolumeAttribute = "gcsfuseMetadataPrefetchOnMount"
 
+	// gcsfuse profiles constants
+	GcsfuseProfilesLabel                                  = "gke-gcsfuse/profiles"
```
uriel-guzman: This is an annotation, not a label.

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +129,122 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+}
+
+// Checks if the pod has gcsfuse profiles annotation. returns true if the annotation is present and set to "true" (case insensitive), false otherwise.
+func IsGCSFuseProfilesEnabled(pod *corev1.Pod) bool {
+	// Check if pod has gcsfuse profiles annotation set to true.
+	if pod.Annotations == nil {
+		return false
+	}
+	value, ok := pod.Annotations[GcsfuseProfilesLabel]
+	return ok && strings.EqualFold(strings.TrimSpace(value), "true")
```
uriel-guzman: Let's use the same logic for evaluating `gke-gcsfuse/volumes: true`: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/dc2da44afa027b40b6936a778984258ae56077bd/pkg/webhook/mutatingwebhook.go#L83

pkg/webhook/sidecar_spec.go
uriel-guzman: Also, move that over here together with the rest of the annotations: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/dc2da44afa027b40b6936a778984258ae56077bd/pkg/webhook/mutatingwebhook.go#L37

pkg/webhook/mutatingwebhook.go
uriel-guzman: Also, remove the punctuation. Logs shouldn't end with a `.`. Here and everywhere else.

chrisThePattyEater:
@gemini-code-assist review

pkg/webhook/mutatingwebhook.go
```diff
@@ -127,10 +129,132 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		}
+	}
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name == GcsFuseSidecarName {
+			// Found the sidecar as a container, now add mounts safely
+			addMounts(&pod.Spec.Containers[i])
+			return nil
+		}
+	}
+	klog.Error(fmt.Sprintf("Could not find gcsfuse sidecar container in pod %s/%s to add gcsfuse profile file cache mounts.", pod.Namespace, pod.Name))
```
uriel-guzman: `klog.Errorf` instead of `klog.Error(fmt.Sprintf`

uriel-guzman:
Note that https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/957 is now merged.

Could you also please remove the following hardcoded values from the example spec and verify they are injected?

1. https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/dc2da44afa027b40b6936a778984258ae56077bd/examples/gcsfuse-profiles/pv-pvc-deployment.yaml#L68

2. https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/dc2da44afa027b40b6936a778984258ae56077bd/examples/gcsfuse-profiles/pv-pvc-deployment.yaml#L92

After this has been verified, please include the result in your PR description and I'll LGTM.

uriel-guzman (approved):

PR #948, opened 2025-08-22
Bucket scanner controller - Initial PR.

uriel-guzman:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR introduces the basic scaffolding for a new PersistentVolume Scanner controller within the GCSFuse CSI driver. The eventual goal of this controller is to watch PersistentVolumes managed by gcsfuse.csi.storage.gke.io and perform metadata scans on the associated GCS buckets.

This initial commit lays the groundwork by:
-   Defining the `Scanner` and `ScannerConfig` structs.
-   Initializing the Kubernetes client, shared informer factory (for PVs and SCs), and workqueue.
-   Setting up basic event handlers (`addPV`, `updatePV`, `deletePV`) for PersistentVolumes, which simply enqueue the PV key.
-   Including a skeleton `syncPV` function that currently just logs the key being processed.
-   Adding basic unit tests to verify scanner creation and event handler enqueuing.

This controller does not yet perform any scanning or modify any resources. Future PRs will build upon this foundation to add PV relevance checks, the bucket scanning logic, and annotation updates.

**Which issue(s) this PR fixes**:

Fixes #<insert-issue-number> 

**Special notes for your reviewer**:
-   This is the first in a series of PRs to implement the PV Scanner feature.
-   This PR focuses only on establishing the controller skeleton and the PV informer/workqueue mechanism.
-   No actual bucket scanning or PV annotation logic is included in this change.
-   The `syncPV` function is currently a no-op placeholder.
-   Tests in `scanner_test.go` cover the basic setup and event enqueuing.

**Does this PR introduce a user-facing change?**:

```release-note
NONE

PR #957, opened 2025-09-05
Bucket scanner controller - Pod creation trigger

uriel-guzman:
**What type of PR is this?**
> /kind feature

This PR enhances the Bucket Scanner controller within the GCSFuse CSI driver by introducing a Pod creation trigger and a scheduling gate mechanism. Instead of only reacting to PV events, the controller now watches for the creation of Pods labeled with `gke-gcsfuse/profile-managed: "true".`

This PRs is part of a chain of previous PRs to implement the bucket scanner controller:
* https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/948
* https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/953

New Pods with this label are expected to start with a scheduling gate `gke-gcsfuse/bucket-scan-pending`. The controller will:

* **Watch Pods**: Identify new Pods with the label and scheduling gate.
* **Resolve Volumes**: Trace through the Pod's volumes to find PersistentVolumeClaims and their bound PersistentVolumes.
* **Check PV Relevance**: Determine if any of the bound PVs require a bucket scan based on the StorageClass configuration and TTL (Time-To-Live).
* **Enqueue PVs**: Trigger scans for relevant PVs by enqueuing them.
* **Manage Scheduling Gate**: The syncPod function reconciles the Pod. It only removes the `gke-gcsfuse/bucket-scan-pending` gate after confirming that all relevant PVs associated with the Pod have been scanned at least once (i.e., they are no longer considered "relevant" for an initial scan). The Pod remains unschedulable until the gate is removed.
* **Requeue Pods**: Pods are requeued if PVCs are not yet bound or if any PVs still require scanning.

**Key Improvements & Changes**:

* **Pod-Tuned Scans**: Scans are now primarily driven by Pod usage, ensuring metadata is fresh when a workload starts.
* **Scheduling Gate**: Prevents Pods from starting until the bucket scan (if required) is completed, ensuring the GCS Fuse driver has the necessary bucket information.
* **Informer Optimizations**: The Pod informer uses a label selector (`gke-gcsfuse/profile-managed=true`) and a transform function (trimPodObject) to reduce memory consumption.
* **Robust TTL Handling**: Introduced an in-memory map (lastSuccessfulScan) to track the last scan time, mitigating potential race conditions caused by informer cache delays. This supplements the on-PV annotations.
* **Workqueue Prefixes**: Items in the workqueue are now prefixed with pv/ or pod/ to distinguish between resource types.
* **New Event Handlers**: Added handlers for Pod Add and Delete events.
* **Unit Tests**: Expanded scanner_test.go to include tests for syncPod, Pod/PVC/PV interactions, scheduling gate removal, and the new event handlers.

**Manual Verification**:
An example deployment that would interact with this controller:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deployment
spec:
  replicas: 2
  selector:
    matchLabels:
      app: my-app
  template:
    metadata:
      labels:
        app: my-app
        # Patched by mutating webhook:
        gke-gcsfuse/profile-managed: "true"
      annotations:
        gke-gcsfuse/volumes: "true"
    spec:
      serviceAccountName: my-ksa
      containers:
      - name: my-container
        image: nginx:latest
        volumeMounts:
        - name: my-gcs-volume
          mountPath: "/data/gcs"
        resources:
          limits:
            cpu: 100m
            memory: 100Mi
          requests:
            cpu: 10m
            memory: 80Mi
      volumes:
      - name: my-gcs-volume
        persistentVolumeClaim:
          claimName: my-pvc
      schedulingGates:
        # Patched by mutating webhook:
        - name: "gke-gcsfuse/bucket-scan-pending"
```

The controller will process this Pod, check the PV bound to my-pvc, trigger a scan if needed, and then remove the gate. Here are some test case examples:

1. PV is being scanned:
```
Name:            my-pv
Labels:          <none>
Annotations:     pv.kubernetes.io/bound-by-controller: yes
Finalizers:      [kubernetes.io/pv-protection]
StorageClass:    gcsfusecsi-inference
Status:          Bound
Claim:           default/my-pvc
Reclaim Policy:  Retain
Access Modes:    RWX
VolumeMode:      Filesystem
Capacity:        10Gi
Node Affinity:   <none>
Message:         
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            gcsfuse.csi.storage.gke.io
    FSType:            
    VolumeHandle:      test-bucket
    ReadOnly:          false
    VolumeAttributes:  <none>
Events:
  Type    Reason                       Age   From                 Message
  ----    ------                       ----  ----                 -------
  Normal  ScanOperationStartSucceeded  4s    gke-gcsfuse-scanner  Started bucket scan for PV my-pv, bucket test-bucket, directory '/my-folder/', with timeout 2m0s
```
Pods are schedule gated while they wait:
```
default               my-deployment-68f5ff8675-5zbpr                                   0/2     SchedulingGated  
default               my-deployment-68f5ff8675-7z2qd                                   0/2     SchedulingGated   
```
PV is finished scanning (notice that it was only scanned once, even though there were 2 Pods):
```
Warning  ScanOperationTimedOut        2s    gke-gcsfuse-scanner  Bucket scan timed out after 2m0s for bucket test-bucket, directory '/my-folder/' (2m0.002581084s). Updating with partial results. Consider increasing timeout if bucket size is big
```
Pod scheduling gates are removed:
```
default               my-deployment-68f5ff8675-5zbpr                                   0/2     Init:0/1   
default               my-deployment-68f5ff8675-7z2qd                                   0/2     Init:0/1   
```
A subsequent Pod attempts to use the PV shortly after (this one's gate is immediately removed, since the scan was already peformed):
```
default               my-deployment-68f5ff8675-5zbpr                                   0/2     Init:0/1   
default               my-deployment-68f5ff8675-7z2qd                                   0/2     Init:0/1  
default               my-deployment-68f5ff8675-fqrlz                                   0/2     Init:0/1   
```
A future Pod attempts to use the PV after the scanner's TTL expires and is gated:
```
default               my-deployment-68f5ff8675-5zbpr                                   0/2     Init:0/1  
default               my-deployment-68f5ff8675-7z2qd                                   0/2     Init:0/1  
default               my-deployment-68f5ff8675-fqrlz                                   0/2     Init:0/1   
default               my-deployment-68f5ff8675-98rqw                                    0/2     SchedulingGated  
```
The scanner starts a second operation:
```
  Warning  ScanOperationTimedOut        5m37s                  gke-gcsfuse-scanner  Bucket scan timed out after 10s for bucket test-bucket, directory '/my-folder/' (10.002400573s). Updating with partial results. Consider increasing timeout if bucket size is big
  Normal   ScanOperationStartSucceeded  5m10s (x2 over 5m47s)  gke-gcsfuse-scanner  Started bucket scan for PV my-pv, bucket test-bucket, directory '/my-folder/', with timeout 10s
  Warning  ScanOperationTimedOut        5m                     gke-gcsfuse-scanner  Bucket scan timed out after 10s for bucket test-bucket, directory '/my-folder/' (10.000924826s). Updating with partial results. Consider increasing timeout if bucket size is big
```
The scheduling gate is also removed from the new Pod:
```
default               my-deployment-68f5ff8675-5zbpr                                   0/2     Init:0/1  
default               my-deployment-68f5ff8675-7z2qd                                   0/2     Init:0/1  
default               my-deployment-68f5ff8675-fqrlz                                   0/2     Init:0/1   
default               my-deployment-68f5ff8675-98rqw                                    0/2     Init:0/1
```

Fixes #

* This PR builds upon the Bucket Scanner skeleton. The core PV scanning logic (syncPV and defaultScanBucket) remains largely the same (still simulated bucket scan).
* The actual GCS bucket interaction (e.g., using Dataflux or GCS client) is still not yet implemented in defaultScanBucket.
* Leader election is not yet implemented.
* A mutating webhook will be used to inject the `gke-gcsfuse/profile-managed` label and the schedulingGates into user Pods. That webhook is not part of this PR, but is being developed here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/955. The TODOs in the example YAML reflect this.
* The new test file scanner_test.go contains detailed unit tests for the Pod reconciliation logic.

```release-note
NONE
```

### DIFF
--- a/pkg/scanner/scanner.go
+++ b/pkg/scanner/scanner.go
@@ -27,13 +27,13 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
-	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
@@ -44,7 +44,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	corev1listers "k8s.io/client-go/listers/core/v1"
+	typedv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	v1listers "k8s.io/client-go/listers/core/v1"
 	storagelisters "k8s.io/client-go/listers/storage/v1"
 )
 
@@ -76,6 +77,17 @@ const (
 	// Bucket scan status values
 	scanCompleted = "completed"
 	scanTimeout   = "timeout"
+
+	// Key prefixes for workqueue
+	pvPrefix  = "pv/"
+	podPrefix = "pod/"
+
+	// Scheduling Gate
+	schedulingGateName = "gke-gcsfuse/bucket-scan-pending"
+
+	// Label for filtering Pods
+	profileManagedLabelKey   = "gke-gcsfuse/profile-managed"
+	profileManagedLabelValue = "true"
 )
 
 var (
@@ -90,6 +102,9 @@ var (
 
 	// scanBucket is a function to scan the bucket, can be overridden in tests.
 	scanBucket = defaultScanBucket
+
+	// Fake error to signal worker to re-queue the Pod without emitting an error log.
+	errRequeuePod = errors.New("requeuing pod")
 )
 
 // stringPtr returns a pointer to the passed string.
@@ -122,19 +137,34 @@ type bucketInfo struct {
 // Scanner is the main controller structure.
 type Scanner struct {
 	kubeClient    kubernetes.Interface
-	pvLister      corev1listers.PersistentVolumeLister
+	pvLister      v1listers.PersistentVolumeLister
+	pvcLister     v1listers.PersistentVolumeClaimLister
 	scLister      storagelisters.StorageClassLister
+	podLister     v1listers.PodLister
 	pvSynced      cache.InformerSynced
+	pvcSynced     cache.InformerSynced
 	scSynced      cache.InformerSynced
+	podSynced     cache.InformerSynced
 	factory       informers.SharedInformerFactory
+	podFactory    informers.SharedInformerFactory
 	queue         workqueue.TypedRateLimitingInterface[string]
 	eventRecorder record.EventRecorder
+
+	// pvMutex protects trackedPVs.
+	pvMutex sync.RWMutex
 	// Set of PV names being tracked/processed. This is
 	// used to avoid reprocessing if multiple user Pods
 	// trigger a PV addition to the queue.
 	trackedPVs map[string]struct{}
-	// Mutex to protect trackedPVs.
-	pvMutex sync.RWMutex
+
+	// scanMutex protects lastSuccessfulScan.
+	scanMutex sync.RWMutex
+	// Map to track the last successful scan time for each PV in this instance.
+	// This is required because there exists a non-zero time window where a PV
+	// patch may not be reflected in the informer's cache after syncPV processes
+	// and removes the key from trackedPVs, potentially resulting in multiple
+	// scans not seeing the PV's annotationLastUpdatedTime.
+	lastSuccessfulScan map[string]time.Time
 }
 
 // buildConfig creates a Kubernetes rest.Config for the client.
@@ -153,6 +183,35 @@ func buildConfig(kubeconfigPath string) (*rest.Config, error) {
 	}
 	klog.Info("Using In-Cluster Kubeconfig")
 	return cfg, nil
+}
+
+// trimPodObject is a transform function for the Pod informer to reduce memory usage.
+// It keeps only the fields relevant to the scanner logic.
+func trimPodObject(obj any) (any, error) {
+	if accessor, err := meta.Accessor(obj); err == nil {
+		accessor.SetManagedFields(nil)
+	} else {
+		klog.Warningf("Failed to get meta accessor for object: %v", err)
+	}
+
+	podObj, ok := obj.(*v1.Pod)
+	if !ok {
+		return obj, nil
+	}
+
+	// Create a new Pod object with only the fields we need.
+	trimmedPod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podObj.ObjectMeta.Name,      // Needed for workqueue key.
+			Namespace: podObj.ObjectMeta.Namespace, // Needed for workqueue key.
+		},
+		Spec: v1.PodSpec{
+			Volumes:         podObj.Spec.Volumes,         // Needed to check PVs requiring scans.
+			SchedulingGates: podObj.Spec.SchedulingGates, // Needed to remove scheduling gates.
+		},
+	}
+
+	return trimmedPod, nil
 }
 
 // NewScanner creates a new Scanner instance.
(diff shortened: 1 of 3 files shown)


pkg/scanner/scanner.go
```diff
@@ -249,24 +335,181 @@ func (s *Scanner) processNextWorkItem(ctx context.Context) bool {
...
+		klog.V(6).Infof("Successfully synced %s %q", syncType, itemKey)
+	} else if strings.Contains(err.Error(), requeuePodMessage) {
+		// Specific requeue signal for pods waiting on PV scans or PVC binds
+		klog.Infof("Requeuing %s %q: %v", syncType, itemKey, err)
+		s.queue.AddRateLimited(key)
 	} else {
-		klog.Errorf("Error syncing PV %q: %v", key, err)
+		klog.Errorf("Error syncing %s %q: %v", syncType, itemKey, err)
 		s.queue.AddRateLimited(key)
 	}
```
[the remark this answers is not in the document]
uriel-guzman: Resolved.

pkg/scanner/scanner.go
```diff
@@ -249,24 +335,181 @@ func (s *Scanner) processNextWorkItem(ctx context.Context) bool {
...
 	return true
 }
 
+// syncPod is the core reconciliation function for a Pod. It checks if any of the
+// Pod's PVs require scanning and removes the Pod's scheduling gate:
+//  1. If any PV requires scanning, both the PVs and the Pod are enqueued to be
+//     handled later. This eventually results in the PV being scanned by syncPV,
+//     and the Pod's scheduling gate being removed by syncPod (step #2).
+//  2. If none of the PVs require scanning, the Pod's scheduling gate is removed,
+//     allowing the Kubernetes scheduler to find a Node for the Pod.
```
saikat-royc, line 392: also put a comment about the error

pkg/scanner/scanner.go
```diff
@@ -174,25 +233,48 @@ func NewScanner(config *ScannerConfig) (*Scanner, error) {
...
 	}
+
+	// Factory for PV, PVC and SC informers
 	factory := informers.NewSharedInformerFactory(kubeClient, config.ResyncPeriod)
 	pvInformer := factory.Core().V1().PersistentVolumes()
+	pvcInformer := factory.Core().V1().PersistentVolumeClaims()
 	scInformer := factory.Storage().V1().StorageClasses()
 
+	// Factory for Pod informer with label selector and transform
+	podLabelSelector := fmt.Sprintf("%s=%s", profileManagedLabelKey, profileManagedLabelValue)
```
saikat-royc, line 245: would the user put this label? or the webhook? best to capture a comment

pkg/scanner/scanner.go
```diff
@@ -249,24 +335,181 @@ func (s *Scanner) processNextWorkItem(ctx context.Context) bool {
...
+	klog.Infof("Syncing Pod: %s/%s", pod.Namespace, pod.Name)
+
+	var anyPVRelevant bool
+	var needsRequeue bool
+
+	for _, vol := range pod.Spec.Volumes {
+		if vol.PersistentVolumeClaim == nil {
+			continue
+		}
+
```
saikat-royc, line 423: can you clarify the behavior? if we encounter error due to unbounded PVC, would the pod be schedule gated ? 
the webhook may face the same issue and, of not knowing where a PVC is the one which intends to use the storage profile? Would the webhook deny admission of the pod, if PVC or SC is missing?


another point is even an unrelated PVC (somehting which is not even using gcsfuse) can block it (i.e keep it in schedule gated state), if the PVC get fails for any reason. Something we can call out explicity in the comment

uriel-guzman: > if we encounter error due to unbounded PVC, would the pod be schedule gated?

In L430, if the PV is unbounded to the PVC (i.e. `pvName` is ""), the Pod is re-queued until it is bound.

> the webhook may face the same issue and, of not knowing where a PVC is the one which intends to use the storage profile? Would the webhook deny admission of the pod, if PVC or SC is missing?

This is a trickier one, but should be handled separately in the webhook PR: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/955/files#diff-86e3c6139b8e3b43a302a9ef4638929d20d2797c3ba5787b757a58a207ea0a24R248. An edge case may exist if the Pod is created before the PVC. For this PR, this is handled by re-queuing the Pod if the PVC doesn't exist yet (L421). 

Since the webhook can't re-try like `syncPod` can, I think the best solution may be to have the customer explicitly be required to add the `gke-gcsfuse/profile-managed` label to the Pod spec. While a minor incovenience, I believe it's the only way to make the mutation deterministic. However, regardless of this, I think the `syncPod` function would remain the same since it only watches Pods with that label.

> another point is even an unrelated PVC (somehting which is not even using gcsfuse) can block it (i.e keep it in schedule gated state), if the PVC get fails for any reason. Something we can call out explicity in the comment

Yes, I think this is going to be unavoidable though, since we don't know if the PVC are relevant or not until they are successfully GET from the API. I'll add this comment.

pkg/scanner/scanner.go
uriel-guzman: I think we'll need to resolve the discussion in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/957#discussion_r2331700662 first before I add this comment. I'm inclining now more towards the user putting it, in order to resolve the race condition.

pkg/scanner/scanner.go
uriel-guzman: Acknowledged, I'll add it in the next push once the webhook discussion is resolved.

pkg/scanner/scanner.go
saikat-royc: As discussed offline, it is reasonable tradeoff (and shared responsibility) to ask the user to add the necessary label.

pkg/scanner/scanner.go
```diff
@@ -249,24 +335,181 @@ func (s *Scanner) processNextWorkItem(ctx context.Context) bool {
...
+				needsRequeue = true
+				continue
+			}
+			klog.Errorf("Failed to get PVC %s/%s for Pod %s: %v", namespace, pvcName, key, err)
+			return fmt.Errorf("failed to get PVC %s/%s: %w", namespace, pvcName, err) // API server error, retry with backoff
+		}
+
+		pvName := pvc.Spec.VolumeName
+		if pvName == "" {
+			klog.Infof("PVC %s/%s for Pod %s is not bound to a PV yet, will recheck Pod later", namespace, pvcName, key)
```
saikat-royc: ... is not bound to a PV yet, requeue pod to evaluate later

pkg/scanner/scanner.go
```diff
@@ -495,20 +748,43 @@ func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume) (*bucketInfo, error)
...
-			if elapsed < *currentScanTTL {
-				klog.Infof("PV %s: Skipping scan, only %s elapsed since last scan, which is less than TTL %s", pv.Name, elapsed.Round(time.Second), currentScanTTL)
-				return nil, nil
+
+	var lastScanTime time.Time
+	found := false
+	source := ""
+
+	s.scanMutex.RLock()
+	if lastTime, ok := s.lastSuccessfulScan[pv.Name]; ok {
```
saikat-royc, line 765: on a crash restart lastSuccessfulScan would be empty? So it would retrigger the scan?

alternately should we think about rebuilding the lastSuccessfulScan map, once the pv informer cache is sycned on a restart?

uriel-guzman: If the controller crashes `lastSuccessfulScan` will be empty, but in L765, it would verify the annotation as fallback. Therefore, if the scan is still within the TTL, the scan will not be restarted during controller crash unnecessarily because the annotation would have the last updated time.

pkg/scanner/scanner.go
uriel-guzman: Resolved.

pkg/scanner/scanner.go
uriel-guzman: Resolving, since `syncPod` will remain the same. The discussion of this conversation will be reflected in the webhook PR.

pkg/scanner/scanner.go
uriel-guzman: Done.

pkg/scanner/scanner.go
uriel-guzman: Added comment.

pkg/scanner/scanner.go
```diff
@@ -249,24 +336,185 @@ func (s *Scanner) processNextWorkItem(ctx context.Context) bool {
...
-		klog.V(6).Infof("Successfully synced PV %q", key)
+		klog.V(6).Infof("Successfully synced %s %q", syncType, itemKey)
+	} else if errors.Is(err, errRequeuePod) {
+		// Specific requeue signal for pods waiting on PV scans or PVC binds
+		klog.Infof("Requeuing %s %q: %v", syncType, itemKey, err)
+		s.queue.AddRateLimited(key)
 	} else {
-		klog.Errorf("Error syncing PV %q: %v", key, err)
+		klog.Errorf("Error syncing %s %q: %v", syncType, itemKey, err)
 		s.queue.AddRateLimited(key)
```
saikat-royc, line 381: is this queue.AddRateLimited needed?

pkg/scanner/scanner.go
saikat-royc: ack

pkg/scanner/scanner.go
uriel-guzman: This is for re-trying to sync the Pod with exponential back-off in the case of an error. For example: L448 (Kubernetes API server error should be re-tried).

saikat-royc: ok add it as a comment

uriel-guzman: Done.

saikat-royc (approved):

PR #964, opened 2025-09-08
ZB Test Flake Fix: Disable compatibility test that does not support zb

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test
> /kind feature
> /kind flake

This pr disables compatibility tests that run using an older gcsfuse sidecar image when running e2e tests with zonal bucket enabled. This is needed because zonal buckets is only supported for gcsfuse 2.10+ but the image uses 2.5. This test is a compatibility test meaning updating is not an option and thus the test should not be applicable to zb.

Fixes #

```release-note

```

### DIFF
--- a/cmd/sidecar_mounter/gcsfuse_binary
+++ b/cmd/sidecar_mounter/gcsfuse_binary
@@ -1,1 +1,1 @@
-gs://gke-release-staging/gcsfuse/v3.2.0-gke.0/gcsfuse_bin
+gs://gke-release-staging/gcsfuse/v3.3.0-gke.1/gcsfuse_bin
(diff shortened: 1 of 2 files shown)


test/e2e/testsuites/volumes.go
```diff
@@ -346,16 +346,33 @@ func (t *gcsFuseCSIVolumesTestSuite) DefineTests(driver storageframework.TestDri
 		}
 	}
 
+	isZBEnabled := false
+	if driver, ok := driver.(*specs.GCSFuseCSITestDriver); ok {
+		isZBEnabled = driver.EnableZB
+	}
+
 	ginkgo.It("should store data using custom sidecar container image", func() {
+		if isZBEnabled {
```
saikat-royc, line 355: do we have a plan to enable these test cases when the custom sidecar would eventually support ZB? if so , let's add a todo

chrisThePattyEater: no, this sidecar image should never change as it is a compatibility test for an olderversion of the sidecar, so it will never support zb

saikat-royc (approved):

PR #966, opened 2025-09-10
Dataflux

uriel-guzman:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


examples/gcsfuse-profiles/README.md
```diff
@@ -0,0 +1,60 @@
...
+limitations under the License.
+-->
+
+# GCSFuse Profiles Example
+
+## Grant the GCSFuse CSI Controller access to your bucket:
+```bash
+GCS_PROJECT=your-bucket-project-id
+BUCKET_NAME=your-bucket-name
+PROJECT_NUMBER=$(gcloud projects describe $(gcloud config get project) --format="value(projectNumber)")
```
amacaskill: Nit: you can swap these then use PROJECT_ID within PROJECT_NUMBER

cmd/csi_driver/main.go
```diff
@@ -51,6 +51,9 @@ var (
 	identityProvider             = flag.String("identity-provider", "", "The Identity Provider to authenticate with GCS API.")
 	enableProfiling              = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
 	enableBucketScanner          = flag.Bool("enable-bucket-scanner", false, "enable the bucket scanner feature")
+	datafluxParallelism          = flag.Int("dataflux-parallelism", 0, "number of go routines for Dataflux lister. Defaults to 0 (10X number of available vCPUs)")
```
amacaskill, line 54: Why do we default to 0? "Defaults to 0 (10X number of available vCPUs)" Seems like 10x number of available vCPUs should be higher than 0?

pkg/scanner/scanner.go
```diff
@@ -565,7 +615,7 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
 		klog.V(6).Infof("PV %q is no longer relevant, skipping sync", key)
 		return nil // Remove irrelevant PV from queue
 	}
-	klog.Infof("PV %q is relevant, bucket: %s, dir: %s", key, bucketInfo.name, bucketInfo.dir)
+	klog.Infof("PV %q is relevant, bucket: %s, dir: %s, onlyDirSpecified: %v", key, bucketInfo.name, bucketInfo.dir, bucketInfo.onlyDirSpecified)
```
amacaskill: onlyDirSpecified: %v -> onlyDirSpecified: %t

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
 
 	// The scan has been successful and complete results have been used.
 	duration := timeNow().Sub(syncStartTime)
-	s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonScanOperationSucceeded, "Bucket scan completed successfully for bucket %s, directory '%s' (%v)", bucketInfo.name, bucketInfo.dir, duration)
-	if patchErr := s.updatePVScanResult(ctx, pv, info, scanCompleted); patchErr != nil {
+	s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonScanOperationSucceeded, "Bucket scan completed successfully for bucket %s, directory '%s' (%v): %d objects, %d bytes", bucketInfo.name, bucketInfo.dir, duration.Round(time.Second), bucketInfo.numObjects, bucketInfo.totalSizeBytes)
```
amacaskill: '%s' -> %q

%q will put quotes around the string

Fix here and everyone else please

pkg/scanner/scanner.go
amacaskill: here and throughout code where you use bool in infof

cmd/csi_driver/main.go
uriel-guzman: `0` in the Dataflux package automatically assigns the parallelism to be 10X the number of the available CPU: https://github.com/googleapis/google-cloud-go/blob/storage/v1.56.1/storage/dataflux/fast_list.go#L75. This is the default behavior.

pkg/scanner/scanner.go
```diff
@@ -300,6 +334,22 @@ func (s *Scanner) Run(ctx context.Context) {
...
+	defer gcsClient.Close()
+
+	metricClient, err := monitoring.NewMetricClient(ctx)
+	if err != nil {
+		klog.Errorf("Failed to create metric client: %v", err)
+		return
+	}
+	s.metricClient = metricClient
+	defer metricClient.Close()
+
```
amacaskill, line 352: Why don't we initialize these clients in NewScanner? Similar to other clients like kubeClient?

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if info.onlyDirSpecified {
+		klog.Infof("onlyDirSpecified is true for bucket %s, dir '%s'. Scanning with Dataflux.", info.name, info.dir)
+		dfErr := scanBucketWithDataflux(ctx, s.gcsClient, info, scanTimeout, s.datafluxConfig)
+		if dfErr != nil {
+			klog.Errorf("Dataflux scan failed for bucket %s, dir '%s': %v", info.name, info.dir, dfErr)
+			// No fallback, as metrics are not applicable for a specific directory.
+		}
+		return dfErr
+	} else {
```
amacaskill: see go/go-style/decisions#indent-error-flow. 
Simplify the if/else nesting where possible. .Since your if ends with return of error,  remove your else statement here and change this to: 

```
if info.onlyDirSpecified {
		klog.Infof("onlyDirSpecified is true for bucket %s, dir '%s'. Scanning with Dataflux.", info.name, info.dir)
		dfErr := scanBucketWithDataflux(ctx, s.gcsClient, info, scanTimeout, s.datafluxConfig)
		if dfErr != nil {
			klog.Errorf("Dataflux scan failed for bucket %s, dir '%s': %v", info.name, info.dir, dfErr)
			// No fallback, as metrics are not applicable for a specific directory.
		}
		return dfErr
}
klog.Infof("onlyDirSpecified is false for bucket %s. Attempting scan with GCS Bucket Metrics first.", info.name)
mErr := scanBucketWithMetrics(ctx, s.metricClient, info)
if mErr == nil {
	klog.Infof("Successfully scanned bucket %s using GCS Bucket Metrics: %d objects, %d bytes", info.name, info.numObjects, info.totalSizeBytes)
	return nil
}
...

```

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
 	}
 	return nil // Remove since this is a complete and successful scan.
 }
 
-// defaultScanBucket simulates scanning a GCS bucket to gather metadata.
+func (s *Scanner) scanBucket(ctx context.Context, info *bucketInfo, scanTimeout time.Duration, pv *v1.PersistentVolume) error {
+	return s.scanBucketImpl(s, ctx, info, scanTimeout, pv)
+}
+
+func defaultGetBucketAttrs(ctx context.Context, gcsClient *storage.Client, bucketName string) (*storage.BucketAttrs, error) {
```
amacaskill: You have several functions violating https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#getters 

defaultGetBucketAttrs  -> defaultBucketAttrs 

getBucketAttrs  -> bucketAttrs 

getOnlyDirValue -> onlyDirValue

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
-func defaultScanBucket(ctx context.Context, info *bucketInfo, scanTimeout time.Duration) (*bucketInfo, error) {
-	klog.Infof("Simulating scan for bucket: %s, dir: %s, timeout: %s", info.name, info.dir, scanTimeout)
+//
+// If the `only-dir` mount option is specified, the bucket will be scanned using the
+// GCS Dataflux client library for the specified directory. Otherwise, Google Cloud Metrics
+// will be used and fallback to the GCS Dataflux client library in the case of any errors or
+// unavailable metrics.
+//
+// Optionally, GCS Dataflux client scanning on the entire bucket can be forced by specifying `only-dir=/`
+func defaultScanBucket(s *Scanner, ctx context.Context, info *bucketInfo, scanTimeout time.Duration, pv *v1.PersistentVolume) error {
```
amacaskill: nit:  I would find this easier to trace if you replace info with bucket so calls like bucket.isHNSEnabled is more clear , or you can do something like bucketI for bucket info.

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	bucketAttrs, err := getBucketAttrs(ctx, s.gcsClient, info.name)
+	if err != nil {
+		return err
+	}
+	info.isHNSEnabled = bucketAttrs.HierarchicalNamespace != nil && bucketAttrs.HierarchicalNamespace.Enabled
+	info.projectNumber = fmt.Sprint(bucketAttrs.ProjectNumber)
+	klog.Infof("Bucket %s HNS enabled: %v", info.name, info.isHNSEnabled)
+
+	if info.onlyDirSpecified {
+		klog.Infof("onlyDirSpecified is true for bucket %s, dir '%s'. Scanning with Dataflux.", info.name, info.dir)
```
amacaskill: This exposes impl details. Should we just say something like `"only-dir" is set for bucket`? Same comment elsewhere

pkg/scanner/scanner.go
uriel-guzman: These are the reasons why I thought it was more aligned with `Run` instead of `NewScanner`:

1. `NewScanner`  doesn't have a context, `Run` does. The clients require a context. We could add a context in `NewScanner` too, but I think it's simpler like this.
2. `NewScanner` is synchronous. It's mostly used to create new data structures. `Run` is asynchronous and keeps the controller running. I think it would make more sense to close the clients when the controller shuts down. If we add the clients in `NewScanner` we need to close them somewhere else, be it the end of `Run` or after this line: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/gcs_fuse_driver.go#L220. I'd rather keep these clients contained in their package and avoid exposing them to the driver package if not required.

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return dfErr
+	} else {
+		klog.Infof("onlyDirSpecified is false for bucket %s. Attempting scan with GCS Bucket Metrics first.", info.name)
+		mErr := scanBucketWithMetrics(ctx, s.metricClient, info)
+		if mErr == nil {
+			klog.Infof("Successfully scanned bucket %s using GCS Bucket Metrics: %d objects, %d bytes", info.name, info.numObjects, info.totalSizeBytes)
+			return nil
+		}
+
+		s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationWarning, "Failed to scan bucket %s using GCS Bucket Metrics, falling back to Dataflux: %v", info.name, mErr)
```
amacaskill: Nit: do you think customers / TSE  will get confused when they see "Failed to scan bucket". I know this happens when things are working as intended, but we have confused customers in the past with errors saying things failed when they didn't. Should we use a word like "Unable to scan bucket" to indicate its not really a "failure"

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
-	defer ticker.Stop()
+	// Optimize resource consumption by filtering only relevant fields.
+	dfInput.Query.SetAttrSelection([]string{"Name", "Size"})
+
+	// Only scan for objects under this directory, if defined.
+	if info.dir != "" && info.dir != "/" {
+		// Ensure that the directory name ends with "/" to avoid picking up
+		// files with prefixes of other directories, since GCS "nested"
+		// objects are just file names with "/" in the names.
+		dfInput.Query.Prefix = strings.Trim(info.dir, "/") + "/"
```
amacaskill: Please make sure you have e2e tests for this case in future PR

pkg/scanner/scanner.go
```diff
@@ -600,48 +650,212 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return fmt.Errorf("failed to fetch total bytes metric: %w", err)
+	}
+
+	info.numObjects = objectCount
+	info.totalSizeBytes = totalBytes
+	return nil
+}
+
+// fetchMetricValue retrieves the latest value for a given metric type from Cloud Monitoring.
+func fetchMetricValue(ctx context.Context, metricClient *monitoring.MetricClient, projectNumber string, info *bucketInfo, metricType string) (int64, error) {
```
amacaskill: I'm curious of performance of this. Whats the max number of times we can call this per day? Is there a way to cut down on number of calls / AlignmentPeriod if we detect it was called recently?

examples/gcsfuse-profiles/README.md
uriel-guzman: Resolved.

pkg/scanner/scanner.go
uriel-guzman: Fixed.

pkg/scanner/scanner.go
uriel-guzman: Fixed. Note that `%s` is used in hundreds of lines of this file. So there will be several substitutions reflected by this change now.

pkg/scanner/scanner.go
uriel-guzman: Fixed.

pkg/scanner/scanner.go
uriel-guzman: Fixed.

pkg/scanner/scanner.go
uriel-guzman: Resolved.

pkg/scanner/scanner.go
amacaskill: Ok, we can always make changes like that in follow up PRs if its going to be that big of change. Let's leave the change in since you already made it, but in the future something to consider

pkg/scanner/scanner.go
uriel-guzman: Fixed.

pkg/scanner/scanner.go
uriel-guzman: Added TODO here.

pkg/scanner/scanner.go
amacaskill: All the Nits are optional, you can choose to reject them if you want

pkg/scanner/scanner.go
uriel-guzman: This is only being called once per scan, which by default is once per week and can be configured by the customer (if they want, but they normally shouldn't). The scan is limited by the `bucketScanTTL` parameter, so it will not be called more than once per that time window.

pkg/scanner/scanner.go
amacaskill: Thanks for explaining that. What happens if clients fail to be created in Run()? Run doesn't return any error, so how do we detect what's happening ?

pkg/scanner/scanner.go
uriel-guzman: Fixed. Note that this was used all over the place, so there's going to be several hundreds of lines for this change as well, since it's a big refactoring operation.

pkg/scanner/scanner.go
uriel-guzman: I already updated the PR to change `%s` to `%q`. So I think we can leave this in now.

pkg/scanner/scanner.go
uriel-guzman: The errors are logged as `klog.Errorf` inside `Run`, which is the main function for the scanner. If we returned the errors in `Run`, we would still `klog.Errorf` them somewhere else, so I think it would be the same outcome.

uriel-guzman:
> What did you change between the tests: `if only-dir not specified and bucket metrics fail` and `only-dir not specified and bucket metrics succeed`. The output looks the same?
> 
> Also did you test `only-dir=/` ? is that covered by only-dir tests in description

Yeah, that was not right. Updated that test (it shouldn't show any warning) and added the `only-dir=/` test as #6.

pkg/scanner/scanner.go
amacaskill: Yea the only difference is we hit the error during driver initialization rather than later during Run, but I think there's pros and cons to both.

pkg/scanner/scanner.go
amacaskill: Do we have a check in place somewhere/test that will fail if buckets start being scanned more than once per week for some reason (or more than bucketScanTTL)?

uriel-guzman: Yes, the basic E2E gcsfuse storage profiles test will have a test case where a Pod is created soon after the first scan, before the `bucketScanTTL`, and will explicitly verify the timestamp to make sure the test is not triggered. Subsequently, another Pod will be created after the `bucketScanTTL` expires and verify that the scan is triggered.

Additionally, there are already 2 unit tests that explicitly verify this. See https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/a453b1824f646b6425e4ec938cd1dd148920fd4d/pkg/scanner/scanner_test.go#L411

PR #969, opened 2025-09-15
Filter out deprecated flag error

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Recently, we've seen a surge in SLO notifications triggered by users including deprecated flags in their configurations. After connecting with the gcsfuse team, we've confirmed that these are false positives; the fuse process is successfully mounting the volume. This pull request addresses the issue by downgrading the severity of these alerts. This will prevent them from impacting our SLOs and allow us to focus on identifying and responding to genuine errors.

I manually tested and logged to verify that this regex if statement is correctly entered for the mount option error.

Fixes #

```release-note
NONE
```


pkg/csi_driver/utils.go
```diff
@@ -457,6 +457,10 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 			return codes.Unavailable, fmt.Errorf("benign grpc error: %s", errMsgStr)
 		}
 
+		re := regexp.MustCompile("Flag .*? has been deprecated, This flag has been deprecated .*? in favor of .*?.")
+		if re.MatchString(errMsgStr) {
+			return codes.Unavailable, fmt.Errorf("benign error: %s", errMsgStr)
+		}
```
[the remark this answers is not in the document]
amacaskill: +1 please move regex compilation to https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/3e89c39c5fc2bc110d18deabfc0c8b8ae56cd8fa/pkg/csi_driver/utils.go#L76-L82.

Also, you are only matching regex for one of the flags. We want to match all the flags deprecated here: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cfg/config.go

Seems like the common phrase will be "Flag --stat-cache-ttl has been deprecated" since gcsfuse uses https://github.com/spf13/pflag. Please add a regex to match that

pkg/csi_driver/utils.go
```diff
@@ -457,6 +458,9 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 			return codes.Unavailable, fmt.Errorf("benign grpc error: %s", errMsgStr)
 		}
 
+		if deprecatedFlagPatterns.MatchString(errMsgStr) {
```
amacaskill, line 462: I know you were just  following pattern of everyone else, but it actually would be good to have a unit test for this. Can you please move everything starting from

 	
```
if err == nil && len(errMsg) > 0 {
```

into a helper function parseCodeFromGCSFuseError or something similar. Have it return an error code + the wrapped error. Then you can unit test it to be sure the exact string we saw in SLO will be fixed with this.

pkg/csi_driver/utils.go
```diff
@@ -414,6 +415,14 @@ func checkGcsFuseErr(isInitContainer bool, pod *corev1.Pod, targetPath string) (
 	if err != nil && !os.IsNotExist(err) {
 		return code, fmt.Errorf("failed to open error file %q: %w", emptyDirBasePath+"/error", err)
 	}
+
+	return extractErrorFromGcsFuseErrorFile(errMsg, err)
+}
+
+func extractErrorFromGcsFuseErrorFile(errMsg []byte, err error) (codes.Code, error) {
+	if err != nil {
```
amacaskill: Shouldn't you also add `&& !os.IsNotExist(err)` here? The purpose of adding the nil check here as well is to have same behavior + have a failing unit test that will fail if someone changes how extractErrorFromGcsFuseErrorFile is called within checkGcsFuseErr. So this should be 

err != nil && !os.IsNotExist(err) and there should be a unit test for that case (when os.IsNotExist(err) is true, then we should return codes.OK, nil

pkg/csi_driver/utils_test.go
```diff
@@ -18,17 +18,93 @@ limitations under the License.
...
+				expectedCode: codes.Unavailable,
+			},
+			{
+				name:         "deprecated flag error",
+				errorMessage: []byte("Flag --stat-cache-ttl has been deprecated, This flag has been deprecated (starting v2.0) in favor of metadata-cache-ttl-secs."),
+				expectedCode: codes.Unavailable,
+			},
+			{
+				name:         "deprecated flag error 2",
+				errorMessage: []byte("%sFlag --max-retry-duration has been deprecated, This is currently unused."),
```
amacaskill: []byte("%sFlag -> []byte("Flag

pkg/csi_driver/utils.go
```diff
@@ -78,6 +78,7 @@ var (
 	managedSidecarRegexAR    = regexp.MustCompile(managedSidecarPatternAR)
 	managedSidecarPatternGCR = `^(gke|staging-gke|master-gke)\.gcr\.io/gcs-fuse-csi-driver-sidecar-mounter:v\d+\.\d+\.\d+-gke\.\d+.*`
 	managedSidecarRegexGCR   = regexp.MustCompile(managedSidecarPatternGCR)
+	deprecatedFlagPatterns   = regexp.MustCompile(`Flag .*? has been deprecated`)
```
amacaskill: Please add comment saying this regex should match the flags marked as deprecated here: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/cfg/config.go

amacaskill (approved):

PR #971, opened 2025-09-16
Bucket scanner controller - Leader election and graceful shutdown

uriel-guzman:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


vendor/github.com/kubernetes-csi/csi-lib-utils/leaderelection/leader_election.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// sanitizeName sanitizes the provided string so it can be consumed by leader election library
+func sanitizeName(name string) string {
+	re := regexp.MustCompile("[^a-zA-Z0-9-]")
+	name = re.ReplaceAllString(name, "-")
+	if name[len(name)-1] == '-' {
+		// name must not end with '-'
+		name = name + "X"
+	}
```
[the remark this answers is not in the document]
uriel-guzman: This is on a vendor package...

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -215,9 +215,7 @@ func (driver *GCSDriver) Run(ctx context.Context, endpoint string) {
 	s := NewNonBlockingGRPCServer()
 	s.Start(endpoint, driver.ids, driver.cs, driver.ns)
 	if driver.config.RunController && driver.config.FeatureOptions.FeatureScanner.Enabled {
-		// Start the scanner on controller driver.
-		// TODO(urielguzman): Enable leader election.
-		go driver.cs.(*controllerServer).scanner.Run(ctx)
+		driver.cs.(*controllerServer).scanner.Start(ctx)
 	}
```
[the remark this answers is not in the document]
uriel-guzman: The gRPC server doesn't expose a context and the PR is intended to only handle graceful shutdown of the scanner, not to improve the driver. So I think we can skip this.

pkg/scanner/scanner.go
```diff
@@ -371,6 +385,59 @@ func (s *Scanner) Run(ctx context.Context) {
...
+	kubeconfig.QPS = float32(s.config.KubeAPIQPS)
+	kubeconfig.Burst = s.config.KubeAPIBurst
+	kubeconfig.ContentType = runtime.ContentTypeProtobuf
+
+	leClient, err := kubernetes.NewForConfig(kubeconfig)
+	if err != nil {
+		klog.Fatalf("Failed to create leaderelection client: %v", err)
+	}
+	// Name of config map with leader election lock
+	lockName := "gke-gcsfuse-scanner-leader"
```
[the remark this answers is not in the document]
uriel-guzman: Updated PR description.

pkg/scanner/scanner.go
```diff
@@ -371,6 +398,52 @@ func (s *Scanner) Run(ctx context.Context) {
...
+			OnStartedLeading: func(ctx context.Context) {
+				klog.Infof("Scanner controller %q started leading", s.id)
+				s.run(ctx)
+			},
+			OnStoppedLeading: func() {
+				klog.Fatalf("%q is no longer the leader", s.id)
+			},
+		},
+	})
+}
```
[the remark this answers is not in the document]
uriel-guzman: Fixed PR description.

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -209,15 +209,14 @@ func (driver *GCSDriver) ValidateControllerServiceRequest(c csi.ControllerServic
...
-		// Start the scanner on controller driver.
-		// TODO(urielguzman): Enable leader election.
-		go driver.cs.(*controllerServer).scanner.Run(ctx)
+		go driver.cs.(*controllerServer).scanner.Start(context.TODO())
 	}
+
+	s := NewNonBlockingGRPCServer()
+	s.Start(endpoint, driver.ids, driver.cs, driver.ns)
 	s.Wait()
 }
```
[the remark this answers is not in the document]
uriel-guzman: We're not handling graceful shutdown improvements for the CSI driver in this PR. Only of the scanner.

pkg/scanner/scanner.go
```diff
@@ -371,6 +406,53 @@ func (s *Scanner) Run(ctx context.Context) {
...
+func (s *Scanner) runWithLeaderElection(ctx context.Context, cancel context.CancelFunc) {
+	rl, err := resourcelock.New(
+		resourcelock.LeasesResourceLock,
+		s.config.LeaderElectionNamespace,
+		leaseName,
+		nil,
+		s.kubeClient.CoordinationV1(),
+		resourcelock.ResourceLockConfig{
+			Identity: s.id,
+		})
```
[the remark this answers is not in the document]
uriel-guzman: Event recorder is optional.

deploy/overlays/profiles/controller.yaml
```diff
@@ -29,6 +29,8 @@ spec:
             - "--enable-bucket-scanner=true"
             - "--dataflux-batch-size=25000"
             - "--dataflux-parallelism=0"
+            - "--leader-election=true"
```
amacaskill, line 32: pdcsi, volume populator, and filestorecsi all have leader election liveliness probe. Do we need that here as well? 
- https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/b96005c77e71b8a95c98e94430dc557a29df3ba8/deploy/kubernetes/base/controller/controller.yaml#L49C1-L50C1)
- https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/d285089d75a8e18fbe7ced0d9eb5100bb4b7c75f/deploy/kubernetes/base/controller/controller.yaml#L53

cmd/csi_driver/main.go
```diff
@@ -40,31 +42,40 @@ import (
...
+
+	// Bucket scanner flags.
+	enableBucketScanner          = flag.Bool("enable-bucket-scanner", false, "Enable the bucket scanner feature.")
+	datafluxParallelism          = flag.Int("dataflux-parallelism", 0, "Number of go routines for Dataflux lister. Defaults to 0 (10X number of available vCPUs).")
+	datafluxBatchSize            = flag.Int("dataflux-batch-size", 25000, "Batch size for Dataflux lister. Defaults to 25000.")
+	datafluxSkipDirectoryObjects = flag.Bool("dataflux-skip-directory-objects", false, "Set to true to skip Dataflux listing objects that include files with names ending in '/'.")
+
+	// Leader election flags.
+	leaderElection              = flag.Bool("leader-election", false, "Enables leader election for stateful driver.")
+	leaderElectionNamespace     = flag.String("leader-election-namespace", "gcs-fuse-csi-driver", "The namespace where the leader election resource exists. Defaults to 'gcs-fuse-csi-driver', but is typically overridden in deployments to use the pod's namespace.")
```
amacaskill: nit: I don't think we need a default here. This is always set to metadata.namespace (correct?) which will differ for managed and non managed. 'gcs-fuse-csi-driver' may be non managed namespace, but might not be managed. It would be unexpected if leader-election-namespace isn't set so should we really just be setting "gcs-fuse-csi-driver" here?

amacaskill:
I like the manual test you did ensuring only 1 leader is acting at a given time to scan buckets. Do we plan to implement any automated test to ensure leader election is working properly? I don't see any tests in other components, but just curious if this failure scenario is covered by any SLO or test?

pkg/scanner/scanner.go
```diff
@@ -31,14 +32,16 @@ import (
...
-	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/leaderelection"
```
amacaskill, line 43: Was there a reason we needed to use lower level client-go leaderelection package to implement leader election rather than [ github.com/kubernetes-csi/csi-lib-utils/leaderelection](https://pkg.go.dev/github.com/kubernetes-csi/csi-lib-utils/leaderelection), which is a higher-level wrapper around the client-go leader election library, specifically tailored for use in CSI (Container Storage Interface) sidecar containers?

Seems like implementation could be simpler (less prone to mistakes), if we use the package for CSI drivers, unless there is some missing functionality in github.com/kubernetes-csi/csi-lib-utils/leaderelection?

uriel-guzman: Yeah, I initially used that library, but the wrapper doesn't have the `OnStoppedLeading` [method](https://github.com/uriel-guzman/gcs-fuse-csi-driver/blob/e6664d34857e1e91d8e2a5f70a8b2c98325af3ef/pkg/scanner/scanner.go#L438), which is critical for cancelling the context when leader election fails.

If we don't do this, then a leader election failure would immediately exit, completely bypassing graceful shutdown by [context cancellation](https://github.com/uriel-guzman/gcs-fuse-csi-driver/blob/e6664d34857e1e91d8e2a5f70a8b2c98325af3ef/pkg/scanner/scanner.go#L388) and ignoring all of the [defer methods](https://github.com/uriel-guzman/gcs-fuse-csi-driver/blob/e6664d34857e1e91d8e2a5f70a8b2c98325af3ef/pkg/scanner/scanner.go#L370) that perform cleanup.

Here's an [example](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/b96005c77e71b8a95c98e94430dc557a29df3ba8/pkg/csi_driver/gcfs_driver.go#L438) of Filestore CSI Driver using the wrapper which I was using as reference. It calls `le.Run`, which internally calls the `OnStoppedLeading` with a [fatal exit call](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/leaderelection/leader_election.go#L199), which is not overridable.

deploy/overlays/profiles/controller.yaml
uriel-guzman: The overlay is using a strategic patch merge, so it's adding on top of https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/deploy/base/controller/controller.yaml#L35, which is already using a liveness probe, checking `/healthz` on port `29633`, which is exposed by the gcs-fuse-csi-driver: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/deploy/base/controller/controller.yaml#L101.

cmd/csi_driver/main.go
uriel-guzman: Yeah, I think this is optional default, since ideally it should always be set by the deployment spec.

uriel-guzman: Removed the default.

uriel-guzman:
> I like the manual test you did ensuring only 1 leader is acting at a given time to scan buckets. Do we plan to implement any automated test to ensure leader election is working properly? I don't see any tests in other components, but just curious if this failure scenario is covered by any SLO or test?

We could tackle this in a number of ways, but I have never seen a leader election E2E test either. I think the nuance here is that the non-managed driver the controller is always a single replica on a worker node, so it actually only makes sense in KCP, where I think the controller is replicated. 

One potential idea is to create a specific test where we create 3 replicas of the controller and test exactly what I did in the PR description (verify one scan, kill leader, verify another leader takes over and finishes the scan), but I think we could discuss this in our test plan and add that during the test implementation if we think it would be valuable.

As for SLOs, I don't think any existing SLOs can cover leader election logic. What would the measurement be here? Alert if more than one scanner go routines are coexisting at once? Uptime is already covered.

amacaskill:
> Alert if more than one scanner go routines are coexisting at once?

Yes, that's what I was thinking. Thanks for the detailed response. I was mostly just curious. I'm ok with no e2e since we are just modeling other components, but if adding a test in the future is easy , we should do it

pkg/scanner/scanner.go
amacaskill: Thanks for explanation! I linked this in PR description because this is great context. Thanks!

amacaskill (approved):

PR #936, opened 2025-08-14
Fix version comparison format while using version package

Sneha-at:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>

> /kind bug

**What this PR does / why we need it**:
Currently we compare GCS Fuse version to check if a particular test should run at that version. This comparison logic used different version formats i.e X.Y.Z and X.Y.Z-gke.0 which resulted in incorrect comparisons in some cases. This PR updates the format for all hardcoded version checks and logs if version format is different that expected. Logging is added for debugging purposes, we don't have to fail on version format mismatch.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #975, opened 2025-09-18
Add support and documentation for running GCSFuse CSI Driver on OSS K8s

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR adds support and documentation for running GCSFuse CSI Driver on OSS K8s, providing instructions for 2 different options. Option 1: (workload identity federation + skipCSIBucketAccessCheck. Option 2: Link KSA to GSA. See details in oss-k8s-e2e.md.

1. Fixed issues in Cloud Build install script (cloudbuild-install.yaml) which caused audience mismatch errors during STS token exchange. Typos that were fixed: 
     1. Replace use of identity pool with the identity provider for spec/tokenRequests/0/audience 
     2. Replace specifying IDENTITY_PROVIDER as full URI starting with: `https://iam.googleapis.com/` rather than `//iam.googleapis.com/`
2. Update cloudbuild-build-image.yaml to support building GCSFuse from source at a particular release (instead of only master branch). This is used in e2e k8s OSS deployment documentation. A follow up PR will improve this by letting customers install from published images in gke-release.
3. Add support for detecting if IDENTITY_PROVIDER is GKE provider in format (`https://container.googleapis.com/v1/projects/{project_id}/locations/{location}/clusters/{cluster_name}`) and changing the audience in various steps in STS token exchange to be the correct format based on if it is a GKE cluster or  non-GKE cluster. 

Testing steps: 
- Tested docs/oss-k8s-e2e-setup.md e2e as described in the guide for both hostnetwork / non-hostnetwork, with and without private sidecar image. 
- Tested Ran e2e tests against GKE cluster with my custom image as well: 

```
gcloud container clusters create test-gcsfuse-8 \
    --zone=us-central1-a \
    --workload-pool=amacaskill-joonix.svc.id.goog --machine-type=n2-standard-8
gcloud container clusters get-credentials  test-gcsfuse-8 --zone us-central1-a
# build and push image to registry

export REGION='us-central1'
export PROJECT_ID=amacaskill-joonix
export REGISTRY="$REGION-docker.pkg.dev/$PROJECT_ID/csi-dev"
export LATEST_TAG=$(curl -s "https://api.github.com/repos/GoogleCloudPlatform/gcs-fuse-csi-driver/releases/latest" | jq -r .tag_name)
export GCSFUSE_TAG=$(curl -sL "https://raw.githubusercontent.com/GoogleCloudPlatform/gcs-fuse-csi-driver/$LATEST_TAG/cmd/sidecar_mounter/gcsfuse_binary" | cut -d'/' -f5 | cut -d'-' -f1)
echo "latest GCSFuse CSI driver release is $LATEST_TAG, which uses GCSFuse version $GCSFUSE_TAG"
latest GCSFuse CSI driver release is v1.18.1, which uses GCSFuse version v3.3.0
gcloud builds submit . --config=cloudbuild-build-image.yaml --substitutions=_REGISTRY=$REGISTRY,_BUILD_GCSFUSE_FROM_SOURCE=true,_STAGINGVERSION=final-oss-not-private,_GCSFUSE_TAG=$GCSFUSE_TAG
# install non-managed driver on GKE cluster
make install REGISTRY=$REGISTRY STAGINGVERSION=final-oss-not-private PROJECT=amacaskill-joonix

#Confirm that the driver is up and running
kubectl get CSIDriver,Deployment,DaemonSet,Pods -n gcs-fuse-csi-driver


# run tests on that cluster, run 
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false STAGINGVERSION=final-oss-not-private REGISTRY=$REGISTRY
```

Fixes #

```release-note
Add support and documentation for running GCSFuse CSI Driver on OSS K8s
```

### DIFF
--- a/cloudbuild-build-image.yaml
+++ b/cloudbuild-build-image.yaml
@@ -22,6 +22,9 @@ substitutions:
   _STAGINGVERSION: 'v999.999.999'
   # (Optional) Set to 'true' to build gcsfuse from its source repository.
   _BUILD_GCSFUSE_FROM_SOURCE: 'false'
+  # (Optional) Set to a gcsfuse tag (https://github.com/GoogleCloudPlatform/gcsfuse/tags) to build gcsfuse from its source repository,
+  # at that branch or tag. Defaults to 'master' branch. If no branch exists with that name, it will try to find the tag with that name.
+  _GCSFUSE_TAG: 'master'
   # (Optional) Set to 'true' to also build for arm64. Note, this is not yet supported with cloud build.
   # TODO(amacaskill): Add support for building arm64 images with cloud build.
   _BUILD_ARM: 'false'
@@ -85,8 +88,16 @@ steps:
         exit 0
       fi
 
-      curl https://raw.githubusercontent.com/GoogleCloudPlatform/gcsfuse/master/tools/package_gcsfuse_docker/Dockerfile -o Dockerfile.gcsfuse
-      GCSFUSE_VERSION="0.0.1-gcsfuse-git-master-$(git ls-remote https://github.com/GoogleCloudPlatform/gcsfuse.git HEAD | cut -c1-7)"
+      curl -L https://raw.githubusercontent.com/GoogleCloudPlatform/gcsfuse/${_GCSFUSE_TAG}/tools/package_gcsfuse_docker/Dockerfile -o Dockerfile.gcsfuse
+      
+      if [ "${_GCSFUSE_TAG}" = "master" ]; then
+        GCSFUSE_VERSION="0.0.1-gcsfuse-git-master-$(git ls-remote https://github.com/GoogleCloudPlatform/gcsfuse.git HEAD | cut -c1-7)"
+      else
+        # Remove the leading 'v' so we can pass this to docker buildx build build-arg GCSFUSE_VERSION
+        GCSFUSE_VERSION=$(echo "${_GCSFUSE_TAG}" | sed 's/^v//')
+      fi
+      
+      echo "Using GCSFuse version: $${GCSFUSE_VERSION}"
       echo -n "$${GCSFUSE_VERSION}" > /workspace/gcsfuse_version.txt
 
       # Build for AMD64 and save to a tar file
@@ -94,7 +105,7 @@ steps:
         --file Dockerfile.gcsfuse \
         --tag gcsfuse-release-amd \
         --build-arg GCSFUSE_VERSION=$${GCSFUSE_VERSION} \
-        --build-arg BRANCH_NAME=master \
+        --build-arg BRANCH_NAME=${_GCSFUSE_TAG} \
         --build-arg ARCHITECTURE=amd64 \
         --platform=linux/amd64 \
         --output type=docker,dest=/workspace/gcsfuse-amd.tar .
@@ -105,7 +116,7 @@ steps:
           --file Dockerfile.gcsfuse \
           --tag gcsfuse-release-arm \
           --build-arg GCSFUSE_VERSION=$${GCSFUSE_VERSION} \
-          --build-arg BRANCH_NAME=master \
+          --build-arg BRANCH_NAME=${_GCSFUSE_TAG} \
           --build-arg ARCHITECTURE=arm64 \
           --platform=linux/arm64 \
           --output type=docker,dest=/workspace/gcsfuse-arm.tar .
@@ -132,7 +143,7 @@ steps:
         docker run --rm -v $(pwd)/bin/linux/amd64:/release gcsfuse-release-amd cp /gcsfuse_$${GCSFUSE_VERSION}_amd64/usr/bin/gcsfuse /release
       else
         # Download pre-built binary
-        gsutil "${_GCSFUSE_BINARY_GCS_PATH}/linux/amd64/gcsfuse" ./bin/linux/amd64/gcsfuse
+        gsutil cp "${_GCSFUSE_BINARY_GCS_PATH}/linux/amd64/gcsfuse" ./bin/linux/amd64/gcsfuse
       fi
       chmod 0555 ./bin/linux/amd64/gcsfuse
 
@@ -146,7 +157,7 @@ steps:
           docker run --rm -v $(pwd)/bin/linux/arm64:/release gcsfuse-release-arm cp /gcsfuse_$${GCSFUSE_VERSION}_arm64/usr/bin/gcsfuse /release
         else
           # Download pre-built ARM binary
-          gsutil "${_GCSFUSE_BINARY_GCS_PATH}/linux/arm64/gcsfuse" ./bin/linux/arm64/gcsfuse
+          gsutil cp "${_GCSFUSE_BINARY_GCS_PATH}/linux/arm64/gcsfuse" ./bin/linux/arm64/gcsfuse
         fi
         chmod 0555 ./bin/linux/arm64/gcsfuse
       fi
@@ -218,4 +229,3 @@ steps:
         --build-arg STAGINGVERSION=${_STAGINGVERSION} \
         --build-arg TARGETPLATFORM=$$PLATFORM \
         --file ./cmd/metadata_prefetch/Dockerfile .
-
--- a/docs/oss-k8s-e2e-setup.md
+++ b/docs/oss-k8s-e2e-setup.md
@@ -1,0 +1,1064 @@
+<!--
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Cloud Storage FUSE CSI Driver Installation Guide for Self-Built K8s
+
+The following guide is an e2e working setup for installing the Cloud Storage FUSE CSI Driver on a self-built k8s cluster. Since self-built k8s setup can vary, there is no guarantee the following guide will work for other self-built k8s clusters, with different setup steps.
+
+## VM Creation
+
+### Create VPC and Subnet
+
+```bash
+# Replace the env variables with your own values. Region and Zone are the locations for the VMs.
+export PROJECT_ID=amacaskill-joonix
+export REGION="us-central1" 
+export ZONE="$REGION-b"
+
+gcloud config set project $PROJECT_ID
+
+gcloud compute networks create k8s-vpc \
+    --subnet-mode=custom \
+    --bgp-routing-mode=global \
+    --mtu=1500
+
+gcloud compute networks subnets create $REGION-sub-1 \
+    --network=k8s-vpc \
+    --range=10.170.0.0/16 \
+    --enable-private-ip-google-access \
+    --region=$REGION
+```
+
+## Create Master
+
+```bash
+export PROJECT_NUMBER=$(gcloud projects describe $PROJECT_ID --format="value(projectNumber)")
+
+gcloud compute instances create kub-m \
+  --project=$PROJECT_ID \
+  --zone=$ZONE \
+  --machine-type=n2-standard-2 \
+  --network-interface=network-tier=PREMIUM,subnet=us-central1-sub-1 \
+  --metadata=enable-oslogin=true \
+  --can-ip-forward \
+  --maintenance-policy=MIGRATE \
+  --create-disk=auto-delete=yes,boot=yes,device-name=kub-m,image=projects/rocky-linux-cloud/global/images/rocky-linux-9-v20250611,mode=rw,size=100,type=pd-balanced \
+  --no-shielded-secure-boot \
+  --shielded-vtpm \
+  --shielded-integrity-monitoring \
+  --reservation-affinity=any \
+  --tags=k8s-vpc \
+  --service-account=$PROJECT_NUMBER-compute@developer.gserviceaccount.com \
+  --scopes=https://www.googleapis.com/auth/cloud-platform
+```
+
+## Create Node-1
+
+```bash 
+gcloud compute instances create kub-n-1 \
+  --project=$PROJECT_ID \
+  --zone=$ZONE \
+  --machine-type=n2-standard-4 \
+  --network-interface=network-tier=PREMIUM,subnet=us-central1-sub-1 \
+  --metadata=enable-oslogin=true \
+  --can-ip-forward \
+  --create-disk=auto-delete=yes,boot=yes,device-name=kub-n-1,image=projects/rocky-linux-cloud/global/images/rocky-linux-9-v20250611,mode=rw,size=100,type=pd-balanced \
+  --no-shielded-secure-boot \
+  --maintenance-policy=MIGRATE \ 
+  --shielded-vtpm \
+  --shielded-integrity-monitoring \
+  --tags=k8s-vpc \
+  --reservation-affinity=any \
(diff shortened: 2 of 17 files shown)


pkg/cloud_provider/auth/token_sources.go
```diff
@@ -126,13 +136,20 @@ func (ts *GCPTokenSource) FetchIdentityBindingToken(ctx context.Context, k8sSATo
...
 			"identitynamespace:%s:%s",
 			ts.meta.GetIdentityPool(),
 			ts.meta.GetIdentityProvider(),
 		)
 	}
+
+	customAudience := ts.meta.GetCustomAudience()
+	if customAudience != "" {
+		audience = customAudience
+	}
```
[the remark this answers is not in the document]
amacaskill: This was for testing and has been removed

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -88,13 +88,23 @@ func (ts *GCPTokenSource) Token() (*oauth2.Token, error) {
...
 		tokenMap := make(map[string]*authenticationv1.TokenRequestStatus)
 		if err := json.Unmarshal([]byte(ts.k8sSAToken), &tokenMap); err != nil {
 			return "", fmt.Errorf("failed to unmarshal TokenRequestStatus: %w", err)
 		}
-		if trs, ok := tokenMap[ts.meta.GetIdentityPool()]; ok {
+
+		//  For self-managed clusters using a custom audience, the token is keyed
+		// by that audience (the full identity provider URI), not the identity pool.
+		audienceKey := ts.meta.GetIdentityPool()
+		if customAudience := ts.meta.GetCustomAudience(); customAudience != "" {
```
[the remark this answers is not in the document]
amacaskill: This was for testing hnd has been removed

docs/oss-k8s-e2e-setup.md
```diff
@@ -0,0 +1,921 @@
...
+exit
+sudo su -
+
+
+# This has to be run inside of the master vm (any kubectl commands do)
+kubectl get --raw /.well-known/openid-configuration | jq -r .issuer
+
+kubectl get --raw /openid/v1/jwks > cluster-jwks.json
+
+# move cluster-jwks.json from the master vm to cloudtop so I can run gcloud command.
```
saikat-royc: ... from the master vm to the dev machine in order to run the below gcloud commands

docs/oss-k8s-e2e-setup.md
```diff
@@ -0,0 +1,921 @@
...
+   --issuer-uri="https://kubernetes.default.svc.cluster.local" \
+   --attribute-mapping="google.subject=assertion.sub" \
+   --display-name=$WORKLOAD_IDENTITY_POOL_PROVIDER \
+   --jwk-json-path="cluster-jwks.json"
+
+gcloud compute ssh kub-m --zone=$ZONE --project=$PROJECT_ID
+sudo su -
+
+# now create resources used to test workload identity
+kubectl create serviceaccount k8s-on-gce-ksa --namespace default
```
saikat-royc: should we also make this namespace configurable?

docs/oss-k8s-e2e-setup.md
```diff
@@ -0,0 +1,921 @@
...
+exit
+```
+
+## Deploy a pod
+
+1. **Create a Cloud Storage Bucket** following instructions [here](https://cloud.google.com/storage/docs/creating-buckets).
+
+2. **Deploy the pod.**
+
+You must specify the `skipCSIBucketAccessCheck: true` `VolumeAttribute` when deploying Pods that use GCSFuse Volumes on OSS K8s.
```
saikat-royc: you can put a small explanation why this needs to be set to true for this setup

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -88,13 +88,26 @@ func (ts *GCPTokenSource) Token() (*oauth2.Token, error) {
...
+
+		// If identity provider is a GKE identity provider, we use the identity pool as the audience key.
+		// Otherwise, we use the identity provider directly.
+		if util.IsGKEIdentityProvider(identityProvider) {
+			audienceKey = ts.meta.GetIdentityPool()
+		} else {
+			audienceKey = identityProvider
+		}
+
+		if trs, ok := tokenMap[audienceKey]; ok {
```
saikat-royc: in L114, could not find token for the audience key ...

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -88,13 +88,26 @@ func (ts *GCPTokenSource) Token() (*oauth2.Token, error) {
...
+
+		var audienceKey string
+		identityProvider := ts.meta.GetIdentityProvider()
+
+		// If identity provider is a GKE identity provider, we use the identity pool as the audience key.
+		// Otherwise, we use the identity provider directly.
+		if util.IsGKEIdentityProvider(identityProvider) {
+			audienceKey = ts.meta.GetIdentityPool()
+		} else {
+			audienceKey = identityProvider
```
saikat-royc: what about L125 which uses audience as identity pool? should we make that logic consistent as well? the audience key is only used in tokenMap tracking

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -81,12 +81,15 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	if mc.HostNetworkKSAOptIn {
 		if mc.TokenServerIdentityProvider != "" {
 			klog.V(4).Infof("Pod has hostNetwork enabled and token server feature is supported and opted in. Starting Token Server on %s/%s", mc.TempDir, TokenFileName)
+			if !util.IsGKEIdentityProvider(mc.TokenServerIdentityProvider) {
+				klog.Warningf("Unsupported configuration: The HostNetwork Pod KSA feature is enabled with a non-GKE identity provider %q. To use this feature, remove the identityProvider from your VolumeAttributes.", mc.TokenServerIdentityProvider)
```
saikat-royc: " ... remove the identityProvider from your VolumeAttributes"

"." is not needed in logs

cloudbuild-build-image.yaml
```diff
@@ -85,16 +88,24 @@ steps:
         exit 0
       fi
 
-      curl https://raw.githubusercontent.com/GoogleCloudPlatform/gcsfuse/master/tools/package_gcsfuse_docker/Dockerfile -o Dockerfile.gcsfuse
-      GCSFUSE_VERSION="0.0.1-gcsfuse-git-master-$(git ls-remote https://github.com/GoogleCloudPlatform/gcsfuse.git HEAD | cut -c1-7)"
+      curl -L https://raw.githubusercontent.com/GoogleCloudPlatform/gcsfuse/${_GCSFUSE_TAG}/tools/package_gcsfuse_docker/Dockerfile -o Dockerfile.gcsfuse
+      
+      if [ "${_GCSFUSE_TAG}" = "master" ]; then
+        GCSFUSE_VERSION="0.0.1-gcsfuse-git-master-$(git ls-remote https://github.com/GoogleCloudPlatform/gcsfuse.git HEAD | cut -c1-7)"
```
Sneha-at, line 94: Is this just some random version?

pkg/cloud_provider/auth/token_sources.go
```diff
@@ -88,13 +88,26 @@ func (ts *GCPTokenSource) Token() (*oauth2.Token, error) {
 }
 
 // fetch Kubernetes Service Account token by calling Kubernetes API.
+// Skip this code path by using skipCSIBucketAccessCheck: "true".
```
Sneha-at: This will still be triggered if using Host Network with sidecar bucket access check logic

pkg/sidecar_mounter/sidecar_mounter.go
amacaskill: why not? Workload will error if they go through this path, so I thought having a log explaining how to fix the issue is useful. They can see this log in node logs

pkg/cloud_provider/auth/token_sources.go
amacaskill: Yea. fetchK8sSAToken doesn't work at all for oss k8s right now (which is why we always need skipCSIBucketAccessCheck right now), so that's why I missed updating L125 , but you are right, I should have changed that. Will change

pkg/cloud_provider/auth/token_sources.go
amacaskill: fixed

docs/oss-k8s-e2e-setup.md
amacaskill: thanks! fixed

docs/oss-k8s-e2e-setup.md
amacaskill: yes :)

pkg/sidecar_mounter/sidecar_mounter.go
siyanshen: What is the expected experience for non-GKE identity provider? If the user remove their identityProvider from VolumeAttribute they will be redirected to the below "Please set identityProvider in VolumeAttributes", no?

pkg/cloud_provider/auth/token_sources.go
amacaskill: fixed

cloudbuild-build-image.yaml
amacaskill: Yes because  https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/936 broke BUILD_GCSFUSE_FROM_SOURCE. I fixed it in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/980 by changing it to comply with the version check that was added. It is 0.0.1 + commit hash of master

docs/oss-k8s-e2e-setup.md
amacaskill: Fixed

pkg/sidecar_mounter/sidecar_mounter.go
amacaskill: If customer doesn't set identityProvider, [shouldPopulateIdentityProvider](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/2dd5c1af6a3c3a2d889bc85ac205bb4d5246075b/pkg/csi_driver/node.go#L143 ) returns false, so 

[optInHW is set as false](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/2dd5c1af6a3c3a2d889bc85ac205bb4d5246075b/pkg/csi_driver/node.go#L150), so we skip this code entirely, and just fall back to using node default sa (hostnetwork workaround) as explained in docs/oss-k8s-e2e-setup.md. 

Hostnetwork isn't supported without the workaround in oss k8s right now. I think I can actually remove this warning since we should never get to this point. I can also remove  set `hostNetworkPodKSA: "true"` in the hostnetwork oss setup steps since that isn't necessary either since we use the workaround. Will check first

pkg/util/util.go
```diff
@@ -45,12 +44,16 @@ const (
 	TokenServerIdentityProviderConst    = "token-server-identity-provider"
 	OptInHnw                            = "hnw-ksa"
 	EnableCloudProfilerForSidecarConst  = "enable-cloud-profiler-for-sidecar"
+	SidecarContainerTmpVolumeName       = "gke-gcsfuse-tmp"
```
siyanshen, line 47: We should leave a note that this needs to be consistent with https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/2dd5c1af6a3c3a2d889bc85ac205bb4d5246075b/pkg/webhook/sidecar_spec.go#L31

amacaskill: What do you mean? I removed 	

```SidecarContainerTmpVolumeName          = "gke-gcsfuse-tmp"``` from [gcs-fuse-csi-driver/pkg/webhook/sidecar_spec.go](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/2dd5c1af6a3c3a2d889bc85ac205bb4d5246075b/pkg/webhook/sidecar_spec.go#L31) ,and now it imports from this file. I had to do that because of cycle of dependencies, so util is now where we import this from everywhere.

File: docs/oss-k8s-e2e-setup.md (hunk 5 above)
Comment (amacaskill) [reply docs/oss-k8s-e2e-setup.md]: this isn't needed, you only need skipCSIBucketAccessCheck if you don't annotate sa to link GSA to KSA (which is a required step in the setup)

File: pkg/sidecar_mounter/sidecar_mounter.go (hunk 8 above)
Comment (amacaskill) [reply pkg/sidecar_mounter/sidecar_mounter.go]: yea this isn't a case we can hit so removing

You only go into `if mc.HostNetworkKSAOptIn` if both `hostNetworkPodKSA: "true"` and `identityProvider` in VolumeAttributes are set for OSS k8s. hostnetwork isn't supported right now except for workaround.

File: docs/oss-k8s-e2e-setup.md
Comment (amacaskill) [inline docs/oss-k8s-e2e-setup.md:1]: @amacaskill Fix the env variables so they are set in and outside of vm

File: pkg/cloud_provider/auth/token_sources.go (hunk 10 above)
Comment (amacaskill) [reply pkg/cloud_provider/auth/token_sources.go]: Added the comment.

**Review by Sneha-at — APPROVED** [review]

**Comment by Sneha-at** [comment]
LGTM but will be good to have another approval who has more context in OSS K8s support

File: docs/oss-k8s-e2e-setup.md [hunk 12]
```diff
@@ -0,0 +1,941 @@
...
+
+# IMPORTANT: Replace namespace with the namespace you want to use.
+export NAMESPACE=gcsfuse-test-ns
+kubectl create namespace $NAMESPACE
+kubectl create serviceaccount k8s-on-gce-ksa --namespace $NAMESPACE
+
+# This is required because the driver uses this annotation to link the Kubernetes Service Account with a GCP SA in fetchGCPSAToken. This is so the driver can configure the KSA to impersonate IAM service accounts, which configures GKE to exchange the federated access token for an access token from the IAM Service Account Credentials API.
+kubectl annotate serviceaccount k8s-on-gce-ksa \
+    --namespace $NAMESPACE \
+iam.gke.io/gcp-service-account=$GSA@$PROJECT_ID.iam.gserviceaccount.com
```
amacaskill, line 757: @amacaskill Add both Workload Identity Federation for GKE + skipCSIBucketAccessCheck as well as non WorkloadIdentity for GKE method + annotation method. Both are valid and customers might want to use different approaches for different reasons.

pkg/util/util.go
```diff
@@ -45,12 +44,16 @@ const (
...
 
 var (
-	volumeIDRegEx          = regexp.MustCompile(`:.*$`)
-	targetPathRegexp       = regexp.MustCompile(`/var/lib/kubelet/pods/(.*)/volumes/kubernetes\.io~csi/(.*)/mount`)
-	emptyReplacementRegexp = regexp.MustCompile(`kubernetes\.io~csi/(.*)/mount`)
+	volumeIDRegEx            = regexp.MustCompile(`:.*$`)
+	targetPathRegexp         = regexp.MustCompile(`/var/lib/kubelet/pods/(.*)/volumes/kubernetes\.io~csi/(.*)/mount`)
+	emptyReplacementRegexp   = regexp.MustCompile(`kubernetes\.io~csi/(.*)/mount`)
+	gkeIdentityProviderRegex = regexp.MustCompile(
+		`^https://container\.googleapis\.com/v1/projects/[^/]+/locations/[^/]+/clusters/[^/]+$`,
```
saikat-royc: this regex wont work in non prod GKE . we can revisit this in a followup PR as needed

amacaskill: Added all env since we also run eve tests in staging2, and test env is run also. Added sandbox as well.

PR #602, opened 2025-05-12
Add minimum sidecar version for intelligent defaults and enable E2E tests

hungnguyen243:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

**What this PR does / why we need it**:
We have minimum sidecar version that supports the feature now (1.15.1-gke.0), so we can update the version filter logic to use this version. The PR also enables the existing E2E tests for this feature with a 1.33 cluster version filter.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add minimum sidecar version for intelligent defaults on high-performance machine types and enable corresponding E2E tests
```

PR #976, opened 2025-09-18
add version filter for bucket access check and cloud profiler

Sneha-at:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Recent [PR](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/605) broke the testgrid as it passes some mount options to managed sidecar which gets registered as unknown flags to custom sidecar. This PR adds version filters for Sidecar bucket access check and cloud profiler 

***Testing***
```
 make e2e-test E2E_TEST_FOCUS="volumes.*" E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false STAGINGVERSION=$STAGINGVERSION  REGISTRY=$REGISTRY
....

[ReportAfterSuite] PASSED [0.058 seconds]
autogenerated by Ginkgo

Ran 56 of 442 Specs in 121.216 seconds
SUCCESS! -- 56 Passed | 0 Failed | 0 Pending | 386 Skipped
```

Fixes #

This fixes only the `volume.*` focused tests i.e. the custom sidecar failing tests from 
image
I will send a followup CL to fix the failed Mount tests.

```release-note
NONE
```


amacaskill:
Can you please fix the What type of PR is this? .You need to add a space "Uncomment only one  /kind <> line, hit enter to put that in a new line, and remove leading whitespaces from that line:"

pkg/csi_driver/utils.go
```diff
@@ -70,6 +70,8 @@ const (
 	VolumeContextKeyEphemeral              = "csi.storage.k8s.io/ephemeral"
 	VolumeContextKeyBucketName             = "bucketName"
 	tokenServerSidecarMinVersion           = "v1.17.2-gke.0" // #nosec G101
+	sidecarBucketAccessCheckMinVersion     = "v1.18.2-gke.0"
+	sidecarCloudProfilerMinVersion         = "v1.18.2-gke.0"
```
amacaskill: Until you actually know the version you will use, set it to something like v1.99 or something we know we won't hit because otherwise, if we bump that version before you cut release for your feature, then they will automatically be re-enabled and fail again. 

This is what we did for machine type defaulting: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/602/files

pkg/csi_driver/node.go
```diff
@@ -148,7 +150,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		fuseMountOptions = joinMountOptions(fuseMountOptions, []string{util.OptInHnw + "=true", util.TokenServerIdentityProviderConst + "=" + identityProvider})
 	}
 
-	if s.driver.config.EnableSidecarBucketAccessCheck {
+	if s.driver.config.EnableSidecarBucketAccessCheck && isSidecarVersionSupportedForSidecarBucketAccessCheck(gcsFuseSidecarImage) {
```
amacaskill: Why do you need a function for this. We can just call 

isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, utils.sidecarBucketAccessCheckMinVersion). Same for cloudProfiler

pkg/csi_driver/node.go
```diff
@@ -210,7 +212,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	// Only pass mountOptions flags for defaulting if sidecar container is managed and satisifies min version requirement
-	if s.shouldPassDefaultingFlags(pod) {
+	if s.shouldPassDefaultingFlags(gcsFuseSidecarImage) {
```
amacaskill: Same thing here. Why do we have a function to call another function, isSidecarVersionSupportedForGivenFeature especially if shouldPassDefaultingFlags  is only called once.

pkg/csi_driver/utils_test.go
```diff
@@ -141,101 +141,117 @@ func TestJoinMountOptions(t *testing.T) {
...
+			imageName                  string
+			expectedSupported          bool
+			minFeatureVersionSupported string
 		}{
 			{
-				name:              "should return true for supported sidecar version",
-				imageName:         "us-central1-artifactregistry.gcr.io/gke-release/gke-release/gcs-fuse-csi-driver-sidecar-mounter:v1.18.3-gke.2@sha256:abcd",
-				expectedSupported: true,
+				name:                       "should return true for supported sidecar version (managed driver image) for flag defaulting",
+				imageName:                  "us-central1-artifactregistry.gcr.io/gke-release/gke-release/gcs-fuse-csi-driver-sidecar-mounter:v1.99.0-gke.2@sha256:abcd",
```
amacaskill, line 156: This should be the correct artifact registry format: https://cloud.google.com/artifact-registry/docs/docker/names#containers, and I don't really think we should be returning true when it doesn't match this format. 

IS there a case where this can be -artifactregistry.gcr.io ? instead of -docker.pkg.dev

pkg/csi_driver/node.go
```diff
@@ -105,6 +105,21 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+	// Check if the sidecar container was injected into the Pod
+	pod, err := s.k8sClients.GetPod(vc[VolumeContextKeyPodNamespace], vc[VolumeContextKeyPodName])
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "failed to get pod: %v", err)
+	}
+	gcsFuseSidecarImage := gcsFuseSidecarContainerImage(pod)
+	enableSidecarBucketAccessCheckForSidecarVersion := s.driver.config.EnableSidecarBucketAccessCheck && gcsFuseSidecarImage != "" && isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarBucketAccessCheckMinVersion)
+
+	if optInHostnetworkKSA && enableSidecarBucketAccessCheckForSidecarVersion {
+		// Sidecar bucket access check feature performs this vaildation in sidecar, for host network pods this doing a bucket access check in both node and sidecar will
```
amacaskill: This comment has some typos change to: 

"Sidecar bucket access check feature performs this validation in sidecar. For host network pods, doing a bucket access check in both node and sidecar will lead to increased STS quota so we force skip the bucket access check on node. WI caches token for sidecar so it wouldn't result in quota increase."

pkg/csi_driver/node.go
```diff
@@ -105,6 +105,21 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+		// lead to increased STS quota so force skipping the bucket access check on node. WI caches token for sidecar so it wouldn't result in quota increase.
+		klog.V(6).Infof("Skipping bucket access check `--skipCSIBucketAccessCheck` for Host Network pods as %s is enabled", util.EnableSidecarBucketAccessCheckConst)
+		skipBucketAccessCheck = true
```
amacaskill: If the customer has skipBucketAccessCheck set to false explicitly, does this override their option? Should we do that, or should we only override if it is unset? Is there a reason someone would still want this as false?

should we say "Skipping bucket access check by forcing `--skipCSIBucketAccessCheck=true` for Host Network pods as %s is enabled" ?  I could see how "Skipping bucket access check then saying the flag could be a little confusing

pkg/csi_driver/node.go
```diff
@@ -384,15 +394,18 @@ func (s *nodeServer) shouldPopulateIdentityProvider(pod *corev1.Pod, optInHnwKSA
 	return tokenVolumeInjected && (sidecarVersionSupported || userInput)
 }
 
-func (s *nodeServer) shouldPassDefaultingFlags(pod *corev1.Pod) bool {
-	var sidecarVersionSupported bool
+func (s *nodeServer) shouldPassDefaultingFlags(gcsFuseSidecarImage string) bool {
```
amacaskill: Seems like you removed the use of this function. If its no longer needed, please remove the function

amacaskill (approved):

PR #977, opened 2025-09-22
add some logging to debug version check issue for custom sidecars

Sneha-at:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This adds some logs to test failing prow jobs for custom sidecars

Fixes #

This doesn't actually fix the issue, it just adds more logging to help with root cause

```release-note
NONE
```


pkg/csi_driver/utils.go
```diff
@@ -519,6 +519,7 @@ func getSidecarContainerStatus(isInitContainer bool, pod *corev1.Pod) (*corev1.C
 func isSidecarVersionSupportedForGivenFeature(imageName string, sidecarMinSupportedVersion string) bool {
 	// If the image is from our non-managed testgrid, just assume the sidecar version is supported
 	// since it's built off latest code in main
+	klog.V(4).Infof("Checking version support for image %s and minSupportedVersion %s", imageName, sidecarMinSupportedVersion)
```
amacaskill: Since this is a customer facing log, I feel like this should be more understandable to the customer.  change minSupportedVersion to readable text minimum supported version, and explain checking version support for sidecar flags or something like that so they don't get confused by this log. Disregard if you will revert this after you root cause

PR #981, opened 2025-09-26
Create AnywhereCache during bucket scan operation

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR introduces the following changes, these changes are currently only applicable while using the 'profiles' overlay:

* Add AnywhereCache creation capabilities to the GCSFuse bucket scanner in the syncPV method in scanner.go. During the bucket scan workflow we now attempt to create an anywhere cache for every zone that exists for the region a cluster belongs to. 
* Pass cluster region through make file to the controller deployment through ScannerConfig struct
* Method for determining all available zones for a region (uses google.golang.org/api/compute/v1)
* Adds go dependencies on google.golang.org/api/compute/v1 (for determining zones for a region) and cloud.google.com/go/storage/control/apiv2 (for creating anywherecache)
* Update to docs to include storage.anywhereCaches.create, storage.anywhereCaches.get, storage.anywhereCaches.list to storage profile tuner as these are required by anywhere cache api
* Update to docs to include "roles/compute.viewer" (required one the workload project by compute api while retrieving zones for cluster region)
* Add unit tests for the following new methods in Scanner.go : createAnywhereCache, getAnywhereCacheTtlFromPV, getAnywhereCacheAdmissionPolicyFromPV.

The workflow is as follows:
* Check if AnywhereCache is enabled for the storage class used by a PV
* Determine the zones for the clusters region
* Parse user input for anywherecacheAdmissionPolicy and anywherecacheTtl
* Build the request body for each zone - https://pkg.go.dev/cloud.google.com/go/storage@v1.56.2/control/apiv2/controlpb#CreateAnywhereCacheMetadata
* Perform ac creation for every zone
* Log errors to user using event recorder and retry on any error that is not codes.AlreadyExists

Anywhere cache creation returns a long running operation object, for now we assume that if no error occurs than the anywhere cache was successfully created. The plan is post aga we will perform operations to confirm the cache was created post request being sent.

Testing

Success events from pv (first attempt - AC does not exist):
```
Events:
  Type    Reason                                     Age   From                 Message
  ----    ------                                     ----  ----                 -------
  Normal  ScanOperationStartSucceeded                19s   gke-gcsfuse-scanner  Started bucket scan for PV "my-pv1", bucket "test-bucket-fuech", directory "data/gcs", with timeout 2m0s
  Normal  ScanOperationSucceeded                     19s   gke-gcsfuse-scanner  Bucket scan completed successfully for bucket "test-bucket-fuech", directory "data/gcs" (0s): 0 objects, 0 bytes
  Normal  ScanOperationAnywhereCacheCreateSucceeded  18s   gke-gcsfuse-scanner  Submitted anywhere cache requests for for bucket: test-bucket-fuech
```
https://screenshot.googleplex.com/C3bFX3X8SGp6dBW

Success events from pv (second attempt - AC does exist):
```
Events:
  Type    Reason                                     Age   From                 Message
  ----    ------                                     ----  ----                 -------
  Normal  ScanOperationStartSucceeded                46s   gke-gcsfuse-scanner  Started bucket scan for PV "my-pv1", bucket "test-bucket-fuech", directory "data/gcs", with timeout 2m0s
  Normal  ScanOperationSucceeded                     46s   gke-gcsfuse-scanner  Bucket scan completed successfully for bucket "test-bucket-fuech", directory "data/gcs" (0s): 0 objects, 0 bytes
  Normal  ScanOperationAnywhereCacheCreateSucceeded  45s   gke-gcsfuse-scanner  Submitted anywhere cache requests for for bucket: test-bucket-fuech
```

Failed event invalid ttl
```
Events:
  Type     Reason                                 Age   From                 Message
  ----     ------                                 ----  ----                 -------
  Normal   ScanOperationStartSucceeded            31s   gke-gcsfuse-scanner  Started bucket scan for PV "my-pv1", bucket "test-bucket-fuech", directory "data/gcs", with timeout 2m0s
  Normal   ScanOperationSucceeded                 31s   gke-gcsfuse-scanner  Bucket scan completed successfully for bucket "test-bucket-fuech", directory "data/gcs" (0s): 0 objects, 0 bytes
  Warning  ScanOperationAnywhereCacheCreateError  30s   gke-gcsfuse-scanner  rpc error: code = InvalidArgument desc = failed to get anywhere cache admission policy for PV "my-pv1": invalid anywhere cache admission policy provided provided: invalid value to disable on purpose, valid values are "admit-on-first-miss" or "admit-on-second-miss"
```

Failed multiple api failure
```
Events:
  Type     Reason                                 Age   From                 Message
  ----     ------                                 ----  ----                 -------
  Normal   ScanOperationStartSucceeded            12s   gke-gcsfuse-scanner  Started bucket scan for PV "my-pv1", bucket "test-bucket-fuech", directory "data/gcs", with timeout 2m0s
  Normal   ScanOperationSucceeded                 12s   gke-gcsfuse-scanner  Bucket scan completed successfully for bucket "test-bucket-fuech", directory "data/gcs" (0s): 0 objects, 0 bytes
  Warning  ScanOperationAnywhereCacheCreateError  11s   gke-gcsfuse-scanner  errors occurred while creating anywhere caches: us-central1-a:[test error us-central1-a]
us-central1-b:[test error us-central1-b]
us-central1-c:[test error us-central1-c]
us-central1-f:[test error us-central1-f]
```

Failed event regions
```
Events:
  Type     Reason                                 Age   From                 Message
  ----     ------                                 ----  ----                 -------
  Normal   ScanOperationStartSucceeded            17s   gke-gcsfuse-scanner  Started bucket scan for PV "my-pv1", bucket "test-bucket-fuech", directory "data/gcs", with timeout 2m0s
  Normal   ScanOperationSucceeded                 17s   gke-gcsfuse-scanner  Bucket scan completed successfully for bucket "test-bucket-fuech", directory "data/gcs" (0s): 0 objects, 0 bytes
  Warning  ScanOperationAnywhereCacheCreateError  16s   gke-gcsfuse-scanner  failed to get zones for region: failed to get loction us-central1: googleapi: Error 403: Required 'compute.regions.get' permission for 'projects/25941571321/regions/us-central1', forbidden
```

Fails and retries
```
Events:
  Type     Reason                                 Age               From                 Message
  ----     ------                                 ----              ----                 -------
  Normal   ScanOperationStartSucceeded            13s               gke-gcsfuse-scanner  Started bucket scan for PV "my-pv1", bucket "mr_chris_test", directory "data/gcs", with timeout 2m0s
  Normal   ScanOperationSucceeded                 12s               gke-gcsfuse-scanner  Bucket scan completed successfully for bucket "mr_chris_test", directory "data/gcs" (0s): 0 objects, 0 bytes
  Warning  ScanOperationAnywhereCacheCreateError  5s (x4 over 12s)  gke-gcsfuse-scanner  failed to get zones for region: failed to get location us-central1: googleapi: Error 403: Required 'compute.regions.get' permission for 'projects/25941571321/regions/us-central1', forbidden
```

Sample of what it looks like with no anywherecache - https://screenshot.googleplex.com/6SxfXSCTMU48Gkd

Fixes #

```release-note
NONE
```


pkg/profiles/scanner.go
```diff
@@ -775,9 +816,179 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return false, err
+	}
+
+	requeue := false // keep track if at least one creation failed with a retryable error
+	for _, zone := range zones {
+		acName := fmt.Sprintf("%s-%s", pv.Name, zone)
+		req.AnywhereCache.Zone = zone
+
+		klog.Infof("Attempting to create anywhere cache: %s", acName)
+		_, err := storageControlClient.CreateAnywhereCache(ctx, req)
```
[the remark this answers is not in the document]
chrisThePattyEater: acked

pkg/profiles/scanner.go
```diff
@@ -775,9 +822,188 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+func BuildCreateAnywhereCacheRequest(pv *v1.PersistentVolume, zone string, anywhereCacheTtl *durationpb.Duration, anywhereCacheAdmissionPolicy string) (*controlpb.CreateAnywhereCacheRequest, error) {
+	return &controlpb.CreateAnywhereCacheRequest{
+		// projects/_/buckets/... is the required format for the api
+		Parent: fmt.Sprintf(`projects/_/buckets/%s`, pv.Spec.CSI.VolumeHandle),
+		AnywhereCache: &controlpb.AnywhereCache{
+			Ttl:             anywhereCacheTtl,
+			Zone:            zone,
+			AdmissionPolicy: anywhereCacheAdmissionPolicy,
+		},
+	}, nil
```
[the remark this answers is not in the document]
chrisThePattyEater: this has been tested to create ac, see pr description. Also this documentation does not say id is a required field - https://cloud.google.com/storage/docs/reference/rpc/google.storage.control.v2#google.storage.control.v2.CreateAnywhereCacheRequest

pkg/profiles/scanner.go
```diff
@@ -775,9 +822,188 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+			shouldRequeueThisZone := isRetryableError(err)
+			requeue = requeue || shouldRequeueThisZone
+			s.recordACCreateErrorEvent(pv, acName, err, shouldRequeueThisZone)
+		} else {
+			// TODO(fuechr): validate request was successful, for now we assume the creation was successful if no error.
+			s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonAnywhereCacheCreateSucceeded, "Created Anywhere Cache %s for pv: %s", acName, pv.Name)
+		}
+	}
+	return requeue, nil
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is intended if there is no retryable err we do not need further action

pkg/profiles/scanner.go
```diff
@@ -775,9 +826,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+func BuildCreateAnywhereCacheRequest(pv *v1.PersistentVolume, zone string, anywhereCacheTtl *durationpb.Duration, anywhereCacheAdmissionPolicy string) (*controlpb.CreateAnywhereCacheRequest, error) {
+	return &controlpb.CreateAnywhereCacheRequest{
+		// projects/_/buckets/... is the required format for the api
+		Parent: fmt.Sprintf(`projects/_/buckets/%s`, pv.Spec.CSI.VolumeHandle),
+		AnywhereCache: &controlpb.AnywhereCache{
+			Ttl:             anywhereCacheTtl,
+			Zone:            zone,
+			AdmissionPolicy: anywhereCacheAdmissionPolicy,
+		},
+	}, nil
```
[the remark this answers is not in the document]
chrisThePattyEater: this has been tested and does return already exist for caches that already exists

pkg/profiles/scanner.go
```diff
@@ -775,9 +826,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+
+		if err != nil {
+			shouldRequeueThisZone := isRetryableError(err)
+			requeue = requeue || shouldRequeueThisZone
+			s.recordACCreateErrorEvent(pv, acName, err, shouldRequeueThisZone)
+		} else {
+			// TODO(fuechr): validate request was successful, for now we assume the creation was successful if no error.
+			s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonAnywhereCacheCreateSucceeded, "Created Anywhere Cache %s for pv: %s", acName, pv.Name)
+		}
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: only have 4 per region

pkg/profiles/scanner.go
```diff
@@ -775,9 +822,187 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	return requeue, nil
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is intended, if error occurs we log through error event and continue

cmd/csi_driver/main.go
```diff
@@ -68,6 +68,7 @@ var (
 	datafluxParallelism           = flag.Int("dataflux-parallelism", 0, "Number of go routines for Dataflux lister. Defaults to 0 (10X number of available vCPUs).")
 	datafluxBatchSize             = flag.Int("dataflux-batch-size", 25000, "Batch size for Dataflux lister. Defaults to 25000.")
 	datafluxSkipDirectoryObjects  = flag.Bool("dataflux-skip-directory-objects", false, "Set to true to skip Dataflux listing objects that include files with names ending in '/'.")
+	clusterRegion                 = flag.String("cluster-region", "", "The region in which the cluster is deployed. If not set, the region will be inferred from the cluster location. (e.g. us-central1)")
```
uriel-guzman: What if it's zonal? In that case, then location is more generic.

uriel-guzman: > If not set, the region will be inferred from the cluster location. (e.g. us-central1)

This is kind of confusing, because there's no code that actually determines the cluster's location inside the golang binary, as it *must* be passed down from the deployment. I'd just delete this sentence.

deploy/overlays/profiles/cluster_region_patch_csi_controller.json
```diff
@@ -0,0 +1 @@
+[]
```
uriel-guzman: Please remove this.

deploy/overlays/profiles/controller.yaml
```diff
@@ -31,6 +31,7 @@ spec:
             - "--dataflux-parallelism=0"
             - "--leader-election=true"
             - "--leader-election-namespace=$(CLOUDSTORAGECSI_NAMESPACE)"
+            - "--cluster-region=$(CLUSTER_REGION)"
```
uriel-guzman: cluster-location, CLUSTER_LOCATION, here and everywhere else

deploy/overlays/profiles/controller.yaml
```diff
@@ -46,3 +47,5 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: CLUSTER_REGION
+              value: ""
```
uriel-guzman: Fix

deploy/overlays/stable/cluster_region_patch_csi_controller.json
```diff
@@ -0,0 +1 @@
+[]
```
uriel-guzman: Why do we need this file at all? Plesae remove.

deploy/overlays/stable/kustomization.yaml
```diff
@@ -52,4 +52,10 @@ patches:
     group: apps
     kind: DaemonSet
     name: gcsfusecsi-node
+    version: v1
```
uriel-guzman, line 55: delete

deploy/overlays/profiles/kustomization.yaml
```diff
@@ -65,3 +65,9 @@ patches:
     kind: DaemonSet
     name: gcsfusecsi-node
     version: v1
+- path: cluster_region_patch_csi_controller.json
```
uriel-guzman: I don't understand why this is required.

examples/gcsfuse-profiles/README.md
```diff
@@ -29,7 +29,11 @@ gcloud iam roles create gke.gcsfuse.profileTuner \
   --project=${GCS_PROJECT} \
   --title="GCSFuse CSI Profile Tuner" \
   --description="Allows scanning GCS buckets for objects and retrieving bucket metadata." \
-  --permissions="storage.objects.list,storage.buckets.get"
+  --permissions="storage.objects.list,storage.buckets.get,\
+                storage.anywhereCaches.create,\
+                storage.anywhereCaches.get,\
+                storage.anywhereCaches.list
+                "
   # TODO(urielguzman): Add permissions for Anywhere Cache and update description.
```
uriel-guzman: Remove TODO?

examples/gcsfuse-profiles/README.md
```diff
@@ -29,7 +29,11 @@ gcloud iam roles create gke.gcsfuse.profileTuner \
   --project=${GCS_PROJECT} \
   --title="GCSFuse CSI Profile Tuner" \
   --description="Allows scanning GCS buckets for objects and retrieving bucket metadata." \
```
uriel-guzman: Update description?

examples/gcsfuse-profiles/README.md
```diff
@@ -54,6 +58,14 @@ gcloud storage buckets add-iam-policy-binding gs://${BUCKET_NAME} \
     --role "roles/storage.objectUser"
 ```
 
+## Grant access needed to determine zones for anywherecaches:
+```bash
+gcloud projects add-iam-policy-binding ${PROJECT_ID} \
+	    --member "principal://iam.googleapis.com/projects/${PROJECT_NUMBER}/locations/global/workloadIdentityPools/${PROJECT_ID}.svc.id.goog/subject/ns/gcs-fuse-csi-driver/sa/gcs-fuse-csi-controller-sa" \
```
uriel-guzman: I don't think we need a new section for this. Just add this as step #4 in line 50. Creating more sections are more things customers needs to copy / paste.

pkg/profiles/scanner.go
```diff
@@ -74,12 +80,14 @@ const (
...
-	reasonScanOperationWarning        = "ScanOperationWarning"
-	reasonScanOperationSucceeded      = "ScanOperationSucceeded"
-	reasonScanOperationTimedOut       = "ScanOperationTimedOut"
+	reasonScanOperationStartError      = "ScanOperationStartError"
+	reasonScanOperationStartSucceeded  = "ScanOperationStartSucceeded"
+	reasonScanOperationFailed          = "ScanOperationFailed"
+	reasonScanOperationWarning         = "ScanOperationWarning"
+	reasonScanOperationSucceeded       = "ScanOperationSucceeded"
+	reasonScanOperationTimedOut        = "ScanOperationTimedOut"
+	reasonAnywhereCacheCreateError     = "AnywhereCacheCreateError"
```
uriel-guzman: These names break the style of the scanner event reasons. Prefix with "ScanOperation" to keep consistency.

pkg/profiles/scanner.go
```diff
@@ -148,6 +175,7 @@ type ScannerConfig struct {
 	LeaderElectionLeaseDuration time.Duration
 	LeaderElectionRenewDeadline time.Duration
 	LeaderElectionRetryPeriod   time.Duration
+	ClusterRegion               string
```
uriel-guzman: ClusterLocation

deploy/overlays/profiles/controller.yaml
```diff
@@ -46,3 +47,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: CLUSTER_LOCATION
+              valueFrom:
+                configMapKeyRef:
+                  name: gcsfusecsi-profiles-config
+                  key: cluster-location
```
uriel-guzman: Fix this. It needs an empty line at the end of the YAML.

deploy/overlays/profiles/kustomization.yaml
```diff
@@ -64,4 +63,4 @@ patches:
     group: apps
     kind: DaemonSet
     name: gcsfusecsi-node
-    version: v1
+    version: v1
```
uriel-guzman: Fix newline issue. Also, why was the version param moved? Let's add it back to where it was.

deploy/overlays/stable/kustomization.yaml
```diff
@@ -52,4 +52,10 @@ patches:
...
+- path: cluster_region_patch_csi_controller.json
```
uriel-guzman: Delete.

uriel-guzman: Fix newline.

examples/gcsfuse-profiles/README.md
```diff
@@ -28,9 +28,12 @@ PROJECT_NUMBER=$(gcloud projects describe ${PROJECT_ID} --format="value(projectN
...
   --project=${GCS_PROJECT} \
   --title="GCSFuse CSI Profile Tuner" \
-  --description="Allows scanning GCS buckets for objects and retrieving bucket metadata." \
-  --permissions="storage.objects.list,storage.buckets.get"
-  # TODO(urielguzman): Add permissions for Anywhere Cache and update description.
+  --description="Allows scanning GCS buckets for objects and retrieving bucket metadata and the creation of Anywhere Caches." \
+  --permissions="storage.objects.list,storage.buckets.get,\
+                storage.anywhereCaches.create,\
+                storage.anywhereCaches.get,\
+                storage.anywhereCaches.list
```
uriel-guzman: Why is `"` in a newline? This looks like a bug. Seems like the command will be broken.

pkg/profiles/scanner.go
```diff
@@ -74,12 +80,14 @@ const (
...
-	reasonScanOperationSucceeded      = "ScanOperationSucceeded"
-	reasonScanOperationTimedOut       = "ScanOperationTimedOut"
+	reasonScanOperationStartError               = "ScanOperationStartError"
+	reasonScanOperationStartSucceeded           = "ScanOperationStartSucceeded"
+	reasonScanOperationFailed                   = "ScanOperationFailed"
+	reasonScanOperationWarning                  = "ScanOperationWarning"
+	reasonScanOperationSucceeded                = "ScanOperationSucceeded"
+	reasonScanOperationTimedOut                 = "ScanOperationTimedOut"
+	reasonScanOperationAnywhereCacheCreateError = "ScanOperationAnywhereCacheCreateError"
+	reasonAnywhereCacheCreateSucceeded          = "AnywhereCacheCreateSucceeded"
```
uriel-guzman: You didn't fix AnywhereCacheCreateSucceeded

pkg/profiles/scanner.go
```diff
@@ -117,6 +129,21 @@ var (
 	bucketAttrs            = defaultBucketAttrs
 	scanBucketWithMetrics  = defaultScanBucketWithMetrics
 	scanBucketWithDataflux = defaultScanBucketWithDataflux
+
+	// Non-Retryable GCS Error codes
+	nonRetryCodes = map[codes.Code]bool{
+		codes.InvalidArgument:    true,
+		codes.PermissionDenied:   true,
```
uriel-guzman: I don't think all of these are non-retryable. For example, the permission denied can be fixed by just fixing the permission with whatever needed it. It can also be error once because of IAM propagation latency and succeed later. So let's not be overly restrictive. I'd just retry on InvalidArgument. But to simplify, let's just always retry, since it's kind of hard to know when to retry and when to not retry. We can optimize that later.

pkg/profiles/scanner.go
```diff
@@ -378,6 +415,30 @@ func (s *Scanner) run(ctx context.Context) {
 		}
 	}()
 
+	computeService, err := compute.NewService(context.Background(), option.WithScopes(compute.ComputeReadonlyScope))
```
uriel-guzman: Why create `context.Background()` when you already have a `ctx`?

pkg/profiles/scanner.go
```diff
@@ -378,6 +415,30 @@ func (s *Scanner) run(ctx context.Context) {
...
+	if err != nil {
+		klog.Errorf("Failed to create compute service: %v", err)
+		return
+	}
+	s.computeService = computeService
+	defer func() {
+		if err := metricClient.Close(); err != nil {
```
uriel-guzman: Why is this closing the metric client? This seems like a copy / paste error.

pkg/profiles/scanner.go
```diff
@@ -378,6 +415,30 @@ func (s *Scanner) run(ctx context.Context) {
...
+			klog.Errorf("Failed to close compute service: %v", err)
+		}
+	}()
+
+	storageControlClient, err := control.NewStorageControlClient(context.Background())
```
uriel-guzman: Again, don't use `context.Background()` if there's a `ctx`.

pkg/profiles/scanner.go
```diff
@@ -378,6 +415,30 @@ func (s *Scanner) run(ctx context.Context) {
...
+	if err != nil {
+		klog.Errorf("Failed to create storage control client: %v", err)
+		return
+	}
+	s.storageControlClient = storageControlClient
+	defer func() {
+		if err := metricClient.Close(); err != nil {
```
uriel-guzman: This is also closing metric client?

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	acNameString := ""
+	if acName != "" {
+		acNameString = fmt.Sprintf(" Requested anywherecache: %s.", acName)
+	}
+	s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, "AnywhereCacheCreationFailed %s. The GCS bucket used by pv %s failed to enable anywherecache due to: %v%s", shouldRetryString, pv.Name, err, acNameString)
+}
+
+// isAnywhereCacheEnabled returns the value of 'enableAnywhereCache' or error if invalid value is present.
+func isAnywhereCacheEnabled(storageClassLister storagelisters.StorageClassLister, pv *v1.PersistentVolume) (bool, error) {
+	// get the storage class from sc informer
```
uriel-guzman: User proper capitalization and punctuation in the comments to be consistent withh the rest of this file. Here and everywhere else.

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+	acNameString := ""
+	if acName != "" {
+		acNameString = fmt.Sprintf(" Requested anywherecache: %s.", acName)
+	}
+	s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, "AnywhereCacheCreationFailed %s. The GCS bucket used by pv %s failed to enable anywherecache due to: %v%s", shouldRetryString, pv.Name, err, acNameString)
+}
+
+// isAnywhereCacheEnabled returns the value of 'enableAnywhereCache' or error if invalid value is present.
+func isAnywhereCacheEnabled(storageClassLister storagelisters.StorageClassLister, pv *v1.PersistentVolume) (bool, error) {
```
uriel-guzman: I don't think `isAnywhereCacheEnabled` described this function. Instead, this should be called `shouldAnywhereCacheBeCreated`

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
 	if patchErr := s.updatePVScanResult(ctx, pv, bucketI, scanCompleted); patchErr != nil {
 		return fmt.Errorf("failed to patch PV %q results, err: %w", pv.Name, patchErr)
 	}
+
+	// Check if the pv uses a anywhere cache enabled storage class.
+	isAnywhereCacheEnabled, err := isAnywhereCacheEnabled(s.scLister, pv)
```
uriel-guzman: `checkPVRelevance` is already getting the StorageClass: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L1047

Instead of calling it twice, you should move the GET to syncPV and pass it both to `checkPVRelevance` and to your `shouldAnywhereCacheBeCreated` functions: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L585

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+// isAnywhereCacheEnabled returns the value of 'enableAnywhereCache' or error if invalid value is present.
+func isAnywhereCacheEnabled(storageClassLister storagelisters.StorageClassLister, pv *v1.PersistentVolume) (bool, error) {
+	// get the storage class from sc informer
+	sc, err := storageClassLister.Get(pv.Spec.StorageClassName)
+	if err != nil {
+		// storage class not available
+		return false, err
+	}
+	// retrieve 'enableAnywhereCache' from list of parameters and parse
+	enableAnywhereCache, ok := sc.Parameters["enableAnywhereCache"]
```
uriel-guzman: Make `enableAnywhereCache` a constant and put it together with the rest of the parameters in the top of this file.

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if ok {
+		return wh.ParseBool(enableAnywhereCache)
+	}
+	// found the sc but didn't have the parameter, not enabling anywhere cache
+	return false, nil
```
uriel-guzman: At this point, all this function does is return true and you won't get any errors since you will be passing the StorageClass in your next revision, so this function doesn't add anything new. Just check the map directly from syncPV.

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+}
+
+// getAnywhereCacheTtlFromPV returns the value of 'anywhereCacheTTL', defaults to 1h for no value or error if invalid value is present.
+func getAnywhereCacheTtlFromPV(pv *v1.PersistentVolume) (*durationpb.Duration, error) {
```
uriel-guzman: getAnywhereCacheTTLFromPV

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if pv.Spec.CSI == nil || pv.Spec.CSI.VolumeAttributes == nil {
+		klog.Infof("no volume attributes provided, defaulting the anywhere cache ttl to 1h for pv: %s", pv.Name)
+		return hourDuration, nil
+	}
+
+	ttl, ok := pv.Spec.CSI.VolumeAttributes["anywhereCacheTTL"]
```
uriel-guzman: Make constant and add to top of file.

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+
+	if !ok {
+		klog.Infof("no ttl volume attribute provided, defaulting the anywhere cache ttl to 1h for pv: %s", pv.Name)
+		return hourDuration, nil
+	}
+
+	goDuration, err := time.ParseDuration(ttl)
```
uriel-guzman: What does goDuration mean?

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	return durationpb.New(goDuration), nil
+}
+
+// getAnywhereCacheAdmissionPolicyFromPV returns the value of 'anywhereCacheAdmissionPolicy', defaulting to 'admit-on-first-miss' if no value is provided.
+func getAnywhereCacheAdmissionPolicyFromPV(pv *v1.PersistentVolume) (string, error) {
+	if pv.Spec.CSI == nil || pv.Spec.CSI.VolumeAttributes == nil {
+		klog.Infof("no volume attributes provided, defaulting the anywhere cache admission policy to admit-on-first-miss for pv: %s", pv.Name)
+		return admitOnFirstMiss, nil
+	}
+	admissionPolicy, ok := pv.Spec.CSI.VolumeAttributes["anywhereCacheAdmissionPolicy"]
```
uriel-guzman: Make const and add to top of file.

pkg/profiles/scanner.go
```diff
@@ -763,9 +824,171 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if !ok {
+		klog.Infof("no admission policy volume attribute provided, defaulting the anywhere cache admission policy to admit-on-first-miss for pv: %s", pv.Name)
+		return admitOnFirstMiss, nil
+	}
+	if admissionPolicy == admitOnFirstMiss || admissionPolicy == admitOnSecondMiss {
+		return admissionPolicy, nil
+	}
+	return "", fmt.Errorf("invalid admission anywhereCacheAdmissionPolicy provided: %s, valid values are admit-on-first-miss or admit-on-second-miss", admissionPolicy)
```
uriel-guzman: Nit: This can be simplified as:

switch admissionPolicy {
  case admitOnFirstMiss, admitOnSecundMiss:
    return admissionPolicy
  default:
   // return the error
}

pkg/profiles/scanner.go
```diff
@@ -99,6 +107,12 @@ const (
 	// Cloud Monitoring metrics
 	objectCountMetric = "storage.googleapis.com/storage/object_count"
 	totalBytesMetric  = "storage.googleapis.com/storage/v2/total_bytes"
+
+	// Anywhere Cache constants
+	admitOnFirstMiss    = "admit-on-first-miss"
+	admitOnSecondMiss   = "admit-on-second-miss"
+	enableAnywhereCache = "enableAnywhereCache"
```
uriel-guzman: This grouping is not the same. Please add the `enableAnywhereCache` and `anywhereCacheTTL` grouped with `volumeAttributeScanTimeoutKey` and `volumeAttributeScanTTLKey` and change their names accordingly for consistency (they are volume attributes).

pkg/profiles/scanner.go
```diff
@@ -117,6 +131,21 @@ var (
 	bucketAttrs            = defaultBucketAttrs
 	scanBucketWithMetrics  = defaultScanBucketWithMetrics
 	scanBucketWithDataflux = defaultScanBucketWithDataflux
+
+	// TODO(fuechr): Determine optimized list of retry codes
+	nonRetryCodes = map[codes.Code]bool{
```
uriel-guzman: Please remove these nonRetryCodes and logic associated with it. As we already discussed in the comment I had made before this. Let's re-try everything and figure out the error codes later when we do the SLO / metrics implementation.

pkg/profiles/scanner.go
```diff
@@ -378,6 +417,25 @@ func (s *Scanner) run(ctx context.Context) {
 		}
 	}()
 
+	computeService, err := compute.NewService(ctx, option.WithScopes(compute.ComputeReadonlyScope))
+	if err != nil {
```
uriel-guzman, line 414: No defer statement to close the compute service?

chrisThePattyEater: no it doesn't keep a long standing connection so there is nothing to close

pkg/profiles/scanner.go
```diff
@@ -583,7 +641,14 @@ func (s *Scanner) syncPod(ctx context.Context, key string) error {
 			return fmt.Errorf("failed to get PV %q: %w", pvName, err) // API server error, retry with backoff
 		}
 
-		bucketI, err := s.checkPVRelevance(pv)
+		sc, err := s.getStorageClass(pv)
+		if err != nil {
+			klog.Errorf("Failed to get Profiles enabled StorageClass for PV %q: %v", key, err)
+			s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationStartError, "Failed to get Profiles enabled StorageClass for PV %q: %v", key, err)
```
uriel-guzman: Remove event recorder.

pkg/profiles/scanner.go
```diff
@@ -705,9 +770,15 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
 		return err
 	}
 
+	sc, err := s.getStorageClass(pv)
+	if err != nil {
+		klog.Errorf("Failed to get Profiles enabled StorageClass for PV %q: %v", key, err)
+		s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationStartError, "Failed to get Profiles enabled StorageClass for PV %q: %v", key, err)
```
uriel-guzman: Remove this event recorder.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
 	if patchErr := s.updatePVScanResult(ctx, pv, bucketI, scanCompleted); patchErr != nil {
 		return fmt.Errorf("failed to patch PV %q results, err: %w", pv.Name, patchErr)
 	}
+
+	// Check if the pv uses a anywhere cache enabled storage class.
+	isAnywhereCacheEnabled := false
+	enableAnywhereCache, ok := sc.Parameters["enableAnywhereCache"]
```
uriel-guzman: Use the constant from the top of the file instead of this hardcoded value.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if !ok {
+		return nil
+	}
+	isAnywhereCacheEnabled, err = strconv.ParseBool(enableAnywhereCache)
+	if err != nil {
+		s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, "Will not enable anywhere cache, error occured while checking its enablement for sc %q: %v", pv.Spec.StorageClassName, err)
+		klog.Errorf("will not enable anywhere cache, error occured while checking its enablement for sc %q: %v", pv.Spec.StorageClassName, err)
```
uriel-guzman: Return an InvalidArg error.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+
+	if isAnywhereCacheEnabled {
```
uriel-guzman: Return nil if this is false.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
 	return nil // Remove since this is a complete and successful scan.
 }
 
+// createAnywhereCache does prework and makes request to create anywherecache.
+// Prework: get zones, create clients, and sets up the request for anywhere cache.
+// Create request: sends a create anywhere cache for each zone of the cluster (api is impodent).
+// If any retryable error occurs we requeue the pv to try again later.
+func (s *Scanner) createAnywhereCache(ctx context.Context, pv *v1.PersistentVolume) bool {
+	anywhereCacheTtl, err := getAnywhereCacheTTLFromPV(pv)
```
uriel-guzman: anywhereCacheTTL

pkg/profiles/scanner.go
uriel-guzman: Please use the consts you declared above the file.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+
+	// Check if the pv uses a anywhere cache enabled storage class.
+	isAnywhereCacheEnabled := false
+	enableAnywhereCache, ok := sc.Parameters["enableAnywhereCache"]
+	if !ok {
+		return nil
+	}
+	isAnywhereCacheEnabled, err = strconv.ParseBool(enableAnywhereCache)
+	if err != nil {
+		s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, "Will not enable anywhere cache, error occured while checking its enablement for sc %q: %v", pv.Spec.StorageClassName, err)
```
uriel-guzman: We can just log to the event recorder: InvalidArgument: "failed to enable anywhere cache for PV %q: %v" . The message is a bit verbose, and reasonScanOperationAnywhereCacheCreateError already makes it clear that anywhere cache enablmenet flow failed.

This if statement is also not returning anything. It should return the error.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
 		return fmt.Errorf("failed to patch PV %q results, err: %w", pv.Name, patchErr)
 	}
+
+	// Check if the pv uses a anywhere cache enabled storage class.
+	isAnywhereCacheEnabled := false
+	enableAnywhereCache, ok := sc.Parameters["enableAnywhereCache"]
+	if !ok {
+		return nil
+	}
+	isAnywhereCacheEnabled, err = strconv.ParseBool(enableAnywhereCache)
```
uriel-guzman: `isAnywhereCacheEnabled` is a misnomber. It doesn't actually say whether anywhere cache is enabled or not. Instead, should be called "shouldEnableAnywhereCache"

pkg/profiles/scanner.go
uriel-guzman: if `!shouldEnableAnywhereCache` {
   return ...
}

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if err != nil {
+		s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, "Will not enable anywhere cache, error occured while checking its enablement for sc %q: %v", pv.Spec.StorageClassName, err)
+		klog.Errorf("will not enable anywhere cache, error occured while checking its enablement for sc %q: %v", pv.Spec.StorageClassName, err)
+	}
+
+	if isAnywhereCacheEnabled {
+		shouldRequeue := s.createAnywhereCache(ctx, pv)
```
uriel-guzman: shouldRequeue is a bit redundant and not consistent with the scanner's logic. Instead, make the `createAnywhereCache` return an error. Whether or not it will be re-queue'd is already handled in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L507 so there's no need for this nested and complex logic.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+	acNameString := ""
+	if acName != "" {
+		acNameString = fmt.Sprintf(" Requested anywherecache: %s.", acName)
+	}
+	s.eventRecorder.Eventf(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, "AnywhereCacheCreationFailed %s. The GCS bucket used by pv %s failed to enable anywherecache due to: %v%s", shouldRetryString, pv.Name, err, acNameString)
+}
+
+// getAnywhereCacheTTLFromPV returns the value of 'anywhereCacheTTL', defaults to 1h for no value or error if invalid value is present.
+func getAnywhereCacheTTLFromPV(pv *v1.PersistentVolume) (*durationpb.Duration, error) {
```
uriel-guzman, line 912: This is a duration, correct? No need for this function. Just re-use https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L658

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		AnywhereCache: &controlpb.AnywhereCache{
+			Ttl:             anywhereCacheTtl,
+			Zone:            zone,
+			AdmissionPolicy: anywhereCacheAdmissionPolicy,
+		},
+	}, nil
+}
+
+// recordACCreateErrorEvent records an event for an anywhere cache creation error.
+func (s *Scanner) recordACCreateErrorEvent(pv *v1.PersistentVolume, acName string, err error, willRetry bool) {
```
uriel-guzman: I don't think this helper function adds readability. Just use the event recorder directly.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return durationpb.New(ttlAsDuration), nil
+}
+
+// getAnywhereCacheAdmissionPolicyFromPV returns the value of 'anywhereCacheAdmissionPolicy', defaulting to 'admit-on-first-miss' if no value is provided.
+func getAnywhereCacheAdmissionPolicyFromPV(pv *v1.PersistentVolume) (string, error) {
+	if pv.Spec.CSI == nil || pv.Spec.CSI.VolumeAttributes == nil {
```
uriel-guzman: By this point, you have already called `checkPVRelevance` which validates this. So this logic is not needed.

pkg/profiles/scanner.go
```diff
@@ -763,9 +834,163 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		klog.Infof("no volume attributes provided, defaulting the anywhere cache admission policy to admit-on-first-miss for pv: %s", pv.Name)
+		return admitOnFirstMiss, nil
+	}
+
+	admissionPolicy, ok := pv.Spec.CSI.VolumeAttributes["anywhereCacheAdmissionPolicy"]
```
uriel-guzman: Please use consts.

pkg/profiles/scanner.go
chrisThePattyEater: No this is a go time.duration, the api requires a duratiopb.Duration

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return err
+	}
+
+	if !shouldEnableAnywhereCache {
+		return nil
+	}
+
+	err = s.createAnywhereCache(ctx, pv)
+	if err != nil {
+		s.eventRecorder.Event(pv, v1.EventTypeNormal, reasonScanOperationAnywhereCacheCreateSucceeded, err.Error())
```
uriel-guzman: This event recorder message says anywhere cache succeeded but it actually failed...

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return err
+	}
+	s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonScanOperationAnywhereCacheCreateSucceeded, "Created Anywhere Cache for bucket: %s", bucketI.name)
+	return nil
+}
+
+// createAnywhereCache does prework and makes request to create anywherecache.
+// Prework: get zones, create clients, and sets up the request for anywhere cache.
+// Create request: sends a create anywhere cache for each zone of the cluster (api is impodent).
```
uriel-guzman: idempotent?

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+// If any retryable error occurs we requeue the pv to try again later.
+func (s *Scanner) createAnywhereCache(ctx context.Context, pv *v1.PersistentVolume) error {
+	anywhereCacheTTL, err := getAnywhereCacheTTLFromPV(pv)
+	if err != nil {
+		err = fmt.Errorf("failed to enable anywhere cache for PV %q: %v", pv.Name, err)
+		klog.Error(err)
```
uriel-guzman: you can simplify this by just returning the error and logging it in the caller function (syncPV)

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+			Zone:            zone,
+			AdmissionPolicy: anywhereCacheAdmissionPolicy,
+		},
+	}, nil
+}
+
+// getAnywhereCacheTTLFromPV returns the value of 'anywhereCacheTTL', defaults to 1h for no value or error if invalid value is present.
+func getAnywhereCacheTTLFromPV(pv *v1.PersistentVolume) (*durationpb.Duration, error) {
+	if pv.Spec.CSI == nil || pv.Spec.CSI.VolumeAttributes == nil {
+		klog.Infof("no volume attributes provided, defaulting the anywhere cache ttl to 1h for pv: %s", pv.Name)
```
uriel-guzman: You don't need these checks anymore since you already called checkPVRelevance before this?

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	return nil
+}
+
+// createAnywhereCache does prework and makes request to create anywherecache.
+// Prework: get zones, create clients, and sets up the request for anywhere cache.
+// Create request: sends a create anywhere cache for each zone of the cluster (api is impodent).
+// If any retryable error occurs we requeue the pv to try again later.
+func (s *Scanner) createAnywhereCache(ctx context.Context, pv *v1.PersistentVolume) error {
+	anywhereCacheTTL, err := getAnywhereCacheTTLFromPV(pv)
+	if err != nil {
```
uriel-guzman, line 868: getAnywhereCacheTTLFromPV can only fail because of parsing, so return InvalidArgument here

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+
+	return durationpb.New(ttlAsDuration), nil
+}
+
+// getAnywhereCacheAdmissionPolicyFromPV returns the value of 'anywhereCacheAdmissionPolicy', defaulting to 'admit-on-first-miss' if no value is provided.
+func getAnywhereCacheAdmissionPolicyFromPV(pv *v1.PersistentVolume) (string, error) {
+	admissionPolicy, ok := pv.Spec.CSI.VolumeAttributes[volumeAttributeAnywhereCacheAdmissionPolicy]
+	if !ok {
+		klog.Infof("no admission policy volume attribute provided, defaulting the anywhere cache admission policy to admit-on-first-miss for pv: %s", pv.Name)
```
uriel-guzman: klog.Infof("no admission policy volume attribute provided, defaulting the anywhere cache admission policy to %q for pv: %s", adminOnFirstMiss,pv.Name)

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return admitOnFirstMiss, nil
+	}
+
+	switch admissionPolicy {
+	case admitOnFirstMiss, admitOnSecondMiss:
+		return admissionPolicy, nil
+	default:
+		return "", fmt.Errorf("invalid admission anywhereCacheAdmissionPolicy provided: %s, valid values are admit-on-first-miss or admit-on-second-miss", admissionPolicy)
```
uriel-guzman: fmt.Errorf("invalid anywhere cache admission policy provided provided: %s, valid values are %q or %q", admissionPolicy, admitOnfirstMiss, admitOnsecondMiss)

Let's avoid exposing implementation details and hardcoding stuff as a general practice

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if err != nil {
+		err = fmt.Errorf("failed to enable anywhere cache for PV %q: %v", pv.Name, err)
+		klog.Error(err)
+		return err
+	}
+	anywhereCacheAdmissionPolicy, err := getAnywhereCacheAdmissionPolicyFromPV(pv)
+	if err != nil {
+		err = fmt.Errorf("failed to enable anywhere cache for PV %q: %v", pv.Name, err)
+		klog.Error(err)
+		return err
```
uriel-guzman: getAnywhereCacheAdmissionPolicyFromPV can also only return parsing errors, so return invalid argument here?

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+
+	zones, err := utilGetZonesForClusterLocation(ctx, s.computeService, s.config.ClusterLocation)
+	if err != nil {
+		err = fmt.Errorf("failed to get zones for region: %v", err)
+		klog.Error(err)
+		return err
+	}
+	if s.storageControlClient == nil {
+		err = fmt.Errorf("failed to create storage client")
```
uriel-guzman: It did't fail to create anything. This error should probably just be "storage control client cannot be nil"

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return err
+	}
+
+	errorCreatingAnywhereCache := false
+	for _, zone := range zones {
+		acName := fmt.Sprintf("%s/%s", pv.Spec.CSI.VolumeHandle, zone)
+		req, err := buildCreateAnywhereCacheRequest(pv, zone, anywhereCacheTTL, anywhereCacheAdmissionPolicy)
+		if err != nil {
+			errorCreatingAnywhereCache = true
+			klog.Error(status.Errorf(codes.InvalidArgument, "failed to build anywhere cache request for anywhere cache %q with error: %v", acName, err))
```
uriel-guzman: failed to build anywhere cache for anywhere cache?

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,139 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+			continue
```
uriel-guzman: Why continue if there's an error? It's likely the same error will happen to the other caches too. It's better to return the error early as soon as possible to avoid wasting resources so the scanner can re-try on next sync pv event.

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+
+// createAnywhereCache does prework and makes request to create anywherecache.
+// Prework: get zones, create clients, and sets up the request for anywhere cache.
+// Create request: sends a create anywhere cache for each zone of the cluster (api is idempotent).
+// If any retryable error occurs we requeue the pv to try again later.
+func (s *Scanner) createAnywhereCache(ctx context.Context, pv *v1.PersistentVolume) error {
+	anywhereCacheTTL, err := getAnywhereCacheTTLFromPV(pv)
+	if err != nil {
+		err = status.Errorf(codes.InvalidArgument, "failed to enable anywhere cache for PV %q: %v", pv.Name, err)
+		return err
```
uriel-guzman: return status.Errorf(codes.InvalidArgument, "failed to get anywhere cache ttl for PV %q: %v", pv.Name, err)

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+	anywhereCacheAdmissionPolicy, err := getAnywhereCacheAdmissionPolicyFromPV(pv)
+	if err != nil {
+		err = status.Errorf(codes.InvalidArgument, "failed to enable anywhere cache for PV %q: %v", pv.Name, err)
+		return err
```
uriel-guzman: return status.Errorf(codes.InvalidArgument, "failed to get anywhere cache admission policy for PV %q: %v", pv.Name, err)

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+
+	zones, err := utilGetZonesForClusterLocation(ctx, s.computeService, s.config.ClusterLocation)
+	if err != nil {
+		err = fmt.Errorf("failed to get zones for region: %v", err)
+		return err
```
uriel-guzman: return fmt.Errorf("failed to get zones for region: %v", err)

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+	if s.storageControlClient == nil {
+		err = fmt.Errorf("storage control client should not be nil")
+		return err
```
uriel-guzman: return fmt.Errorf("storage control client should not be nil")

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+
+	errorCreatingAnywhereCache := false
+	for _, zone := range zones {
+		acName := fmt.Sprintf("%s/%s", pv.Spec.CSI.VolumeHandle, zone)
+		req, err := buildCreateAnywhereCacheRequest(pv, zone, anywhereCacheTTL, anywhereCacheAdmissionPolicy)
+		if err != nil {
+			errorCreatingAnywhereCache = true
+			err = status.Errorf(codes.InvalidArgument, "failed to build anywhere cache creation request for %q with error: %v", acName, err)
+			return err
```
uriel-guzman: return status.Errorf(codes.InvalidArgument, "failed to build anywhere cache creation request for %q with error: %v", acName, err)

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if err != nil {
+		err = fmt.Errorf("failed to get zones for region: %v", err)
+		return err
+	}
+	if s.storageControlClient == nil {
+		err = fmt.Errorf("storage control client should not be nil")
+		return err
+	}
+
+	errorCreatingAnywhereCache := false
```
uriel-guzman: I don't understand the purpose of this variable. If an error happens, it immediately returns and nothing is done with it. Delete?

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,131 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		}
+
+		_, err = s.storageControlClient.CreateAnywhereCache(ctx, req)
+		if err != nil {
+			if status.Code(err) == codes.AlreadyExists {
+				// AC already exists for this bucket/zone no further action needed.
+				continue
+			}
+			errorCreatingAnywhereCache = true
+			klog.Error(fmt.Errorf("anywhere cache api failed for anywhere cache %q with error: %v", acName, err))
```
uriel-guzman: Seems like you're just logging these errors to the control plane which customer can't see. As discussed offline, please return a joined error and make sure the scanner reports this to the customer and they can see which zones failed and why.

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,123 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		}
+	}
+	if len(errs) <= 0 {
+		return nil
+	}
+	return fmt.Errorf("errors occurred while creating anywhere caches: %w", errors.Join(errs...))
+}
+
+// BuildCreateAnywhereCacheRequest builds the request to create an anywhere cache based on the pv ().
+func BuildCreateAnywhereCacheRequest(pv *v1.PersistentVolume, zone string, anywhereCacheTtl *durationpb.Duration, anywhereCacheAdmissionPolicy string) (*controlpb.CreateAnywhereCacheRequest, error) {
```
uriel-guzman: Why is this even returning an error at all? It just returns nil.

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,123 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+	if s.storageControlClient == nil {
+		return fmt.Errorf("storage control client should not be nil")
+	}
+
+	var errs []error
+	for _, zone := range zones {
+		req, err := buildCreateAnywhereCacheRequest(pv, zone, anywhereCacheTTL, anywhereCacheAdmissionPolicy)
+		if err != nil {
+			return status.Errorf(codes.InvalidArgument, "failed to build anywhere cache creation request for %s/%s with error: %v", pv.Spec.CSI.VolumeHandle, zone, err)
```
uriel-guzman: I don't even think this is possible. Your buildCreateAnywhereCacheRequest doesn't return any errors. Why InvalidArgument? I don't think this logic makes sense.

Remove the error return value from your helper function buildCreateAnywhereCacheRequest and don't return an error in this line, since it's impossible.

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,123 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	if len(errs) <= 0 {
+		return nil
+	}
+	return fmt.Errorf("errors occurred while creating anywhere caches: %w", errors.Join(errs...))
+}
+
+// BuildCreateAnywhereCacheRequest builds the request to create an anywhere cache based on the pv ().
+func BuildCreateAnywhereCacheRequest(pv *v1.PersistentVolume, zone string, anywhereCacheTtl *durationpb.Duration, anywhereCacheAdmissionPolicy string) (*controlpb.CreateAnywhereCacheRequest, error) {
+	return &controlpb.CreateAnywhereCacheRequest{
+		// projects/_/buckets/... is the required format for the api
```
uriel-guzman: `projects/{project}/buckets/{bucket}` to be more precise.

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,123 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+
+	if !shouldEnableAnywhereCache {
+		return nil
+	}
+
+	err = s.createAnywhereCache(ctx, pv)
+	if err != nil {
+		klog.Error(err)
+		s.eventRecorder.Event(pv, v1.EventTypeNormal, reasonScanOperationAnywhereCacheCreateError, err.Error())
```
uriel-guzman: Why would an error be EventTypeNormal? That's a warning.

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,123 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+		return err
+	}
+	s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonScanOperationAnywhereCacheCreateSucceeded, "Created Anywhere Cache for bucket: %s", bucketI.name)
```
uriel-guzman: I don't think we can yet declare success. Remember these are long-running operations which may actually fail after you log this event.

Maybe you can instead log "Created anywhere cache requests for bucket %s in zones %+v". This just meaans that the requests succeeded. 

Unfortunately we can't know if the AnyC actually succeeded creation unless we check later, which is out of scope. So just change the message wording and I think we're good here.

chrisThePattyEater: Should this be "Submitted..." instead of "Created..."

pkg/util/util.go
```diff
@@ -256,3 +291,11 @@ func ParseBool(str string) (bool, error) {
 		return false, fmt.Errorf("could not parse string to bool: the acceptable values for %q are 'True', 'true', 'false' or 'False'", str)
 	}
 }
+
+func GetProjectNumber(ctx context.Context) (string, error) {
```
uriel-guzman: Cool function. This works?

chrisThePattyEater: yes its been tested

pkg/util/util.go
```diff
@@ -239,6 +240,40 @@ func ParseVolumeID(bucketHandle string) string {
...
+func GetZonesForALocation(ctx context.Context, projectNumber string, computeService *compute.Service, location string) ([]string, error) {
+	var zones []string
+	var err error
+
+	if projectNumber == "" {
+		return zones, fmt.Errorf("no project number provided")
+	}
+
+	// Handle the zonal cluster case.
+	if parts := strings.Split(location, "-"); len(parts) == 2 {
```
uriel-guzman: This seems like a fragile way of validating this. Shall we use compute service's `compute.Regions.Get` and `compute.service.Zones` get intead?
https://cloud.google.com/compute/docs/reference/rest/v1/regions/get

pkg/util/util.go
```diff
@@ -239,6 +240,40 @@ func ParseVolumeID(bucketHandle string) string {
...
+		zones = append(zones, location)
+		return zones, nil
+	}
+	klog.Info("Getting zones for region: ", location, " in project: ", projectNumber)
```
uriel-guzman: Move log to L250 so it shows up both for regions and zones. Otherwise, it just shows up for regions.

deploy/overlays/profiles/controller.yaml
```diff
@@ -46,3 +48,8 @@ spec:
...
```
uriel-guzman, line 55: Please also modify the serving storage class with the anywhere cache param (not commented out): https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/deploy/overlays/profiles/storageclass.yaml#L40

Also, add your new anywhere cache ttl and admission policy params with some example values (commented out).

deploy/overlays/profiles/storageclass.yaml
```diff
@@ -39,6 +39,11 @@ mountOptions:
   - read_ahead_kb=1024
 parameters:
   workloadType: "serving"
+  enableAnywhereCache: "true"
+  # Optional: The longest time that a chunk of data will remain in the anywhere cache from the last read. The default is "1h". 
+  # anywhereCacheTTL: "1h"
+  # Optional: Policy for whether the anywhere cache ingests data on the first miss or on the second miss. The default is "admit-on-first-miss".
+  # anywhereCacheAdmissionPolicy: "admit-on-second-miss"
```
uriel-guzman: Change default to "admin-on-first-miss". In the comment, mention that "admin-on-second-miss" is also an option.

pkg/util/util.go
```diff
@@ -239,6 +245,45 @@ func ParseVolumeID(bucketHandle string) string {
...
+		return zones, nil
+	}
+	if status.Code(err) != codes.NotFound {
+		return zones, fmt.Errorf("failed to get loction %s: %v", location, err)
+	}
+
+	// No region found for the location, checking if its a zone.
+	_, err = computeService.Zones.Get(projectNumber, location).Context(ctx).Do()
+	if err != nil {
+		return zones, fmt.Errorf("failed to get loction %s: %v", location, err)
```
uriel-guzman: Also check for NotFound here. If NotFound, you should return something that says that the location isn't neither a region or a zone. If it's something else, then just return it directly (it's probably an intenal error).

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,115 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+	}
+	if s.storageControlClient == nil {
+		return fmt.Errorf("storage control client should not be nil")
+	}
+
+	var errs []error
+	for _, zone := range zones {
+		req := &controlpb.CreateAnywhereCacheRequest{
+			// projects/{project}/buckets/{bucket} is the required format for the api
+			Parent: fmt.Sprintf(`projects/_/buckets/%s`, pv.Spec.CSI.VolumeHandle),
```
uriel-guzman: Don't use VolumeHandle directly! Use parseVolumeID (which you implemented): https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L1077

uriel-guzman: Does this work? You already have project number for bucket here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L800. Maybe use that to populate that filed? (if it works, then fine, but I still think it's safer to populate it with the requested param).

examples/gcsfuse-profiles/README.md
```diff
@@ -28,9 +28,11 @@ PROJECT_NUMBER=$(gcloud projects describe ${PROJECT_ID} --format="value(projectN
 gcloud iam roles create gke.gcsfuse.profileTuner \
   --project=${GCS_PROJECT} \
   --title="GCSFuse CSI Profile Tuner" \
-  --description="Allows scanning GCS buckets for objects and retrieving bucket metadata." \
-  --permissions="storage.objects.list,storage.buckets.get"
-  # TODO(urielguzman): Add permissions for Anywhere Cache and update description.
+  --description="Allows scanning GCS buckets for objects and retrieving bucket metadata and the creation of Anywhere Caches." \
```
uriel-guzman, line 31: This formatting doesn't seem right? It seems like `storage.anywhereCaches.create` will be parsed as `                     storage.anywhereCaches.create` and not work. Same for all other anywhere cache permissions.

pkg/util/util.go
```diff
@@ -239,6 +245,47 @@ func ParseVolumeID(bucketHandle string) string {
...
+	regionObj, err := computeService.Regions.Get(projectNumber, location).Context(ctx).Do()
+	if err == nil {
+		for _, zoneURL := range regionObj.Zones {
+			zoneName := path.Base(zoneURL)
+			zones = append(zones, zoneName)
+		}
+		return zones, nil
+	}
+	if status.Code(err) != codes.NotFound {
+		return zones, fmt.Errorf("failed to get loction %s: %v", location, err)
```
uriel-guzman: location

pkg/profiles/scanner.go
```diff
@@ -763,7 +825,117 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
+
+	if !shouldEnableAnywhereCache {
+		return nil
+	}
+
+	err = s.createAnywhereCache(ctx, pv)
+	if err != nil {
+		klog.Error(err)
+		s.eventRecorder.Event(pv, v1.EventTypeWarning, reasonScanOperationAnywhereCacheCreateError, err.Error())
+		return err
```
uriel-guzman, line 856: What happens if:

1. Scan succeeds and patches PV with timestamp.
2. Anywhere Cache creation fails
3. syncPV re-synced
4. syncPV skipped because there's already a scan in L772 and L780.

I think this will lead to a bug where AnyC failures are never re-tried.

The simplest solution I can think of is making the `checkPVRelevance` function return the following:

1. bucketInfo (already does this)
2. isRelevant (new: this should be 'true' if the PV is relevant [is actually a gcsfuse volume and has the `workloadType` parameter)
3. alreadyScanned (new: this should be 'true' if the PV has already been scanned within the current TTL [check the `checkPVRelevance` to see what I'm talking about)
4. error (already does this)

Then, do the following:
1. If isRelevant == false, then return nil to syncPV (we don't care about this PV, at all).
2. If alreadyScanned == true, skip the scanning logic and go straight to the AnyC logic.

Unfortunately we can't merge this yet, because this is a kind-of-serious bug that needs to be addressed.

uriel-guzman (changes requested):
request changes

examples/gcsfuse-profiles/README.md
chrisThePattyEater: Issue seems to be the spacing between the comma and forward slash - added a space to correct

pkg/profiles/scanner.go
chrisThePattyEater: I implemented this with one caveat
> Made it so we only return bucket info without an error if pv is relevant, meaning if bucketi is not null it is relevant, this removes the need for the extra isRelevant variable

pkg/profiles/scanner.go
```diff
@@ -705,65 +762,185 @@ func (s *Scanner) syncPV(ctx context.Context, key string) error {
...
 	}
 
+	klog.Infof("PV %q is relevant, bucket: %q, dir: %q, onlyDirSpecified: %t", key, bucketI.name, bucketI.dir, bucketI.onlyDirSpecified)
+
+	// ----- At this stage, the PV has been considered eligible for a scan. -----
+
 	if bucketI.isOverride {
 		// Bypass the scanner if the override mode is set.
-		return s.bypassScanForOverride(ctx, pv, key, bucketI)
+		err = s.bypassScanForOverride(ctx, pv, key, bucketI)
```
uriel-guzman, line 790: Why was this changed? I think we should leave `return s.bypassScanForOverride(ctx, pv, key, bucketI)`

examples/gcsfuse-profiles/README.md
```diff
@@ -28,9 +28,11 @@ PROJECT_NUMBER=$(gcloud projects describe ${PROJECT_ID} --format="value(projectN
...
+  --permissions="storage.objects.list,storage.buckets.get, \
+                storage.anywhereCaches.create \
+                storage.anywhereCaches.get, \
+                storage.anywhereCaches.list"
```
uriel-guzman: This still seems like it's broken.

chrisThePattyEater: ack

pkg/profiles/scanner.go
```diff
@@ -1026,52 +1203,35 @@ func (s *Scanner) updatePVScanResult(ctx context.Context, pv *v1.PersistentVolum
...
+// checkPVRelevance determines if a PersistentVolume is relevant for gcsfuse profiles and whether there is a scan pending.
 // A PV is relevant if it uses the gcsfuse CSI driver and its StorageClass
 // has a workloadType parameter set to serving, training, or checkpointing.
-// The PV is relevant if the current time - last scan time > scan TTL, or if the status is "override".
+// The PV is pending a scan if the current time - last scan time >= scan TTL and the status is not "override".
 // This function returns a bucketInfo with the bucket name and the directory if
 // relevant, otherwise, it will return nil and any error.
-func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume) (*bucketInfo, error) {
+func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume, sc *storagev1.StorageClass) (*bucketInfo, bool, error) {
+	isPendingScan := false
```
uriel-guzman: Nit: This variable is not actually being used for anything. Consider just returning `true` or `false` directly to minimize the risk of some other future code unintentionally modifying the variable and returning the wrong value. Returning directly also improves readability in this case because otherwise we have to check what was the previous value to understand what it's returning on every line.

pkg/profiles/scanner.go
```diff
@@ -1144,10 +1304,12 @@ func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume) (*bucketInfo, error)
 		profilesutil.AnnotationTotalSize,
 		profilesutil.AnnotationHNSEnabled,
 	}); len(annotationsUsed) > 0 {
-		return nil, status.Errorf(codes.InvalidArgument, "scanner annotations for PV %q found in non-override mode: %+v", pv.Name, annotationsUsed)
+		return bucketI, isPendingScan, status.Errorf(codes.InvalidArgument, "scanner annotations for PV %q found in non-override mode: %+v", pv.Name, annotationsUsed)
```
uriel-guzman: Nit: Make the bucketI be returned as `nil` here, since this breaks the pattern where `nil` is returned everywhere else in this function in the case of an error.

PR #982, opened 2025-09-27
Release GCSFuse 3.4.0 and disallow profile flag

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Release GCSFuse v3.4.0 and dis-allow unsupported 'profile' flag which is included in 3.4

Fixes #

Testing setup
```
gcloud container clusters create gcsfuse-cluster     --workload-pool=fuechr-gke-dev.svc.id.goog --machine-type=n2-standard-8 --zone us-central1-c

make build-image-and-push-multi-arch REGISTRY=gcr.io/fuechr-gke-dev STAGINGVERSION=3.4

make install REGISTRY=gcr.io/fuechr-gke-dev STAGINGVERSION=3.4 PROJECT=fuechr-gke-dev

make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false E2E_TEST_SKIP="should.succeed.in.performance.test" E2E_TEST_FOCUS="mount.*" REGISTRY=gcr.io/fuechr-gke-dev STAGINGVERSION=3.4 

```

```release-note
Bumping GCSFuse binary version to v3.4.0
```

### DIFF
--- a/pkg/sidecar_mounter/sidecar_mounter_config.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_config.go
@@ -92,6 +92,7 @@ var disallowedFlags = map[string]bool{
 	"cache-dir":                            true,
 	"experimental-local-file-cache":        true,
 	"prometheus-port":                      true,
+	"profile":                              true,
 }
 
 var boolFlags = map[string]bool{
--- a/pkg/sidecar_mounter/sidecar_mounter_config_test.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_config_test.go
@@ -145,6 +145,18 @@ func TestPrepareMountArgs(t *testing.T) {
 			expectedConfigMapArgs: defaultConfigFileFlagMap,
 		},
 		{
+			name: "temporary until we support profile flag - no valid options provided, should return default maps",
+			mc: &MountConfig{
+				BucketName: "test-bucket",
+				BufferDir:  "test-buffer-dir",
+				CacheDir:   "test-cache-dir",
+				ConfigFile: "test-config-file",
+				Options:    []string{"profile=aiml-training", "profile:aiml-training"},
+			},
+			expectedArgs:          defaultFlagMap,
+			expectedConfigMapArgs: defaultConfigFileFlagMap,
+		},
+		{
 			name: "should return valid args with custom app-name",
 			mc: &MountConfig{
 				BucketName: "test-bucket",
--- a/test/e2e/specs/specs.go
+++ b/test/e2e/specs/specs.go
@@ -1242,3 +1242,18 @@ func (t *TestPod) VerifyDefaultingFlagsArePassed(namespace string, expectedMachi
 	gomega.Expect(stdout).To(gomega.ContainSubstring(expectedMachineTypeFlagString),
 		"Should find MachineType flag string %q in stdout", expectedMachineTypeFlagString)
 }
+
+func (t *TestPod) VerifyProfileFlagsAreNotPassed(namespace string) {
+	stdout, stderr, err := e2ekubectl.RunKubectlWithFullOutput(namespace, "logs", t.pod.Name, "-c", "gke-gcsfuse-sidecar")
+	framework.ExpectNoError(err,
+		"Error accessing logs from pod %v, but failed with error message %q\nstdout: %s\nstderr: %s",
+		t.pod.Name, err, stdout, stderr)
+
+	// handles profile=aiml-training case
+	gomega.Expect(stdout).To(gomega.Not(gomega.ContainSubstring("--profile aiml-training")),
+		"Should not find profile flag string in stdout")
+
+	// handles profile:aiml-training case
+	gomega.Expect(stdout).To(gomega.Not(gomega.MatchRegexp(`map\[.*profile:aiml-training.*\]`)),
+		"Should NOT find 'profile:aiml-training' within the gcsfuse config file content map, but it was found.")
+}
--- a/test/e2e/testsuites/mount.go
+++ b/test/e2e/testsuites/mount.go
@@ -219,6 +219,27 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
 		tPod.Cleanup(ctx)
 	}
 
+	testGcsfuseProfilesFlagFilters := func(configPrefix ...string) {
+		init(configPrefix...)
+		defer cleanup()
+
+		ginkgo.By("Configuring test pod")
+		tPod := specs.NewTestPod(f.ClientSet, f.Namespace)
+		profileMO := "profile=aiml-training,profile:aiml-training"
+		tPod.SetupVolume(l.volumeResource, volumeName, mountPath, false, profileMO)
+
+		tPod.Create(ctx)
+
+		ginkgo.By("Checking pod is running")
+		tPod.WaitForRunning(ctx)
+
+		ginkgo.By("Checking pod is does not have --profile=aiml-training passed to gcsfuse")
+		tPod.VerifyProfileFlagsAreNotPassed(f.Namespace.Name)
+
+		ginkgo.By("Deleting pod")
+		tPod.Cleanup(ctx)
+	}
+
 	ginkgo.It("should pass --machine-type and --disable-autoconfig=false from driver to gcsfuse", func() {
 		testDefaultingFlags(specs.DisableAutoconfig)
 	})
@@ -250,5 +271,9 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
 			e2eskipper.Skipf("skip for volume type %v", storageframework.DynamicPV)
 		}
 		testCaseLongMountOptions()
+	})
+
+	ginkgo.It("should not pass profile as a user-specified mountOption to gcsfuse when profile is specified", func() {
+		testGcsfuseProfilesFlagFilters()
 	})
 }
--- a/cmd/sidecar_mounter/gcsfuse_binary
+++ b/cmd/sidecar_mounter/gcsfuse_binary
@@ -1,1 +1,1 @@
-gs://gke-release-staging/gcsfuse/v3.3.0-gke.1/gcsfuse_bin
+gs://gke-release-staging/gcsfuse/v3.4.0-gke.0/gcsfuse_bin


pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -172,12 +173,22 @@ func NewMountConfig(sp string, flagMapFromDriver map[string]string) *MountConfig
 }
 
 func mergeFlags(mountConfigFlagMap map[string]string, driverFlagMap map[string]string) {
```
amacaskill, line 175: Why are these changes needed? The logic for removing disallowed flags already exists below, so I don't think we need these changes

amacaskill:
please add e2e tests similar to "mount should pass --machine-type and --disable-autoconfig=false from driver to gcsfuse". These should be in e2e mount folder.

pkg/sidecar_mounter/sidecar_mounter_config.go
Sneha-at, line 243: Since we are adding this flag in disallaowed flags, this would skip the `enable-gcsfuse-profile` flag? So it never reaches the switch case.

amacaskill:
> Can you please update release note to correct version? Also, testing setup in CL description is cut off. Can you please update that also?

@chrisThePattyEater Can you please fix this?

deploy/base/node/node.yaml
```diff
@@ -56,6 +56,7 @@ spec:
             - --metrics-endpoint=:9920
             - --max-metric-collectors=10
             - --enable-cloud-profiler-for-driver=false
+            - --enable-gcsfuse-profiles=false
```
amacaskill: Default is false for both flags. Please remove both of these lines as this isn't needed

```
            - --enable-cloud-profiler-for-driver=false
            - --enable-gcsfuse-profiles=false
```

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -92,6 +92,8 @@ var disallowedFlags = map[string]bool{
 	"cache-dir":                            true,
 	"experimental-local-file-cache":        true,
 	"prometheus-port":                      true,
+	"profile":                              true,
+	"enable-gcsfuse-profile":               true,
```
amacaskill: Remove

```

	"profile":                              true,
	"enable-gcsfuse-profile":               true,

```

test/e2e/testsuites/mount.go
```diff
@@ -223,8 +222,12 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
 		testDefaultingFlags(specs.DisableAutoconfig)
 	})
 
-	ginkgo.It("should pass --disable-autoconfig=true as a user-specified mountOption to gcsfuse", func() {
```
amacaskill, line 226: Why did we delete this test case? Please keep this test case as is, and add your own test cases below. We want each test case to only test 1 thing, so when it fails, it is clear why it is failing. So please keep test cases for disable-autoconfig and machine-type separate from --profile and --enable-gcsfuse-profiles

pkg/sidecar_mounter/sidecar_mounter_config.go
amacaskill: +1, remove profile and enable-gcsfuse-profiles from disallowed flags map initialization. enable-gcsfuse-profiles isn't a disallowedFlag in any case, and for profile, whether or not it is disallowed depends on the value of enable-gcsfuse-profiles. behavior we want: If `enable-gcsfuse-profiles` is false,  add add `profile` to disallowedFlags.

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -250,6 +252,10 @@ func (mc *MountConfig) prepareMountArgs() {
 		}
 
 		switch flag {
+		case util.EnableGcsfuseProfiles:
+			flagMap["profile"] = util.TrueStr
```
amacaskill: Adding profile disallowedFlags from beginning means it gets removed from flagmap, then only adding back to flag map if 	case util.EnableGcsfuseProfiles is more complicated and hard to follow.

Please just implement If !util.EnableGcsfuseProfiles then add profile to disallowedFlags.

test/e2e/testsuites/mount.go
amacaskill: We should have 2 test cases: 
1. "should pass --profile as a user-specified mountOption to gcsfuse when enable-gcsfuse-profiles=true"
1. "should not pass --profile as a user-specified mountOption to gcsfuse when enable-gcsfuse-profiles=false"

cmd/csi_driver/main.go
uriel-guzman, line 53: There's already an `--enable-bucket-scanner` flag which does exactly the same thing. I think this should be renamed to `--enable-gcsfuse-profiles` instead of having two flags that enable the same exact feature.

uriel-guzman (dismissed):

cmd/csi_driver/main.go
```diff
@@ -178,6 +179,7 @@ func main() {
 		DisableAutoconfig:              *disableAutoconfig,
 		WINodeLabelCheck:               *wiNodeLabelCheck,
 		EnableSidecarBucketAccessCheck: *enableSidecarBucketAccessCheck,
+		EnableGcsfuseProfiles:          *enableGcsfuseProfiles,
```
uriel-guzman: There is already a feature config for gcsfuse profiles that has an `Enabled` field: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/cmd/csi_driver/main.go#L151. There is no need to duplicate this field here.

deploy/base/node/node.yaml
```diff
@@ -55,7 +55,6 @@ spec:
             - --identity-pool=$(IDENTITY_POOL)
             - --metrics-endpoint=:9920
             - --max-metric-collectors=10
-            - --enable-cloud-profiler-for-driver=false
```
uriel-guzman, line 58: Why is this deleted?

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -59,6 +59,7 @@ type GCSDriverConfig struct {
 	MetricsManager                 metrics.Manager
 	DisableAutoconfig              bool
 	EnableSidecarBucketAccessCheck bool
+	EnableGcsfuseProfiles          bool
```
uriel-guzman: Use `FeatureOptions.FeatureScanner.Enabled` instead of duplicating the flag here.

pkg/csi_driver/node.go
```diff
@@ -150,6 +150,10 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		fuseMountOptions = joinMountOptions(fuseMountOptions, []string{util.OptInHnw + "=true", util.TokenServerIdentityProviderConst + "=" + identityProvider})
 	}
 
+	if s.driver.config.EnableGcsfuseProfiles {
```
uriel-guzman: Check `s.driver.config.FeatureOptions.FeatureScanner.Enabled`.

deploy/base/node/node.yaml
chrisThePattyEater: This should default to false, there is no need to explicitly set it as false

chrisThePattyEater: Change was requested by @amacaskill

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -250,6 +252,13 @@ func (mc *MountConfig) prepareMountArgs() {
 		}
 
 		switch flag {
+		case util.EnableGcsfuseProfiles:
+			klog.Info("BLAHHHHH"+value)
```
uriel-guzman: Remove debugging message?

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -250,6 +252,13 @@ func (mc *MountConfig) prepareMountArgs() {
...
+			if value != util.TrueStr {
+				disallowedFlags["profile"] = true
+			}
```
uriel-guzman: This logic is a bit confusing. There's already a `disallowedFlags` map in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/sidecar_mounter/sidecar_mounter_config.go#L79. Why not simply add `profiles` to that map, and bypass it if `s.driver.config.FeatureOptions.FeatureScanner.Enabled` is `true` during the check at line 243?

cmd/csi_driver/main.go
uriel-guzman: You'll also need to update the name of the flag in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/deploy/overlays/profiles/controller.yaml#L29

cmd/csi_driver/main.go
amacaskill: enable-gcsfuse-profiles isn't related to enabling the bucket scanner. It is an internal flag to allow GCSFuse team to pass --profile mountOption on GKE workloads before we we officially launch profiles with GCSFuse and GKE profiles (at which point, we will  remove profiles from disallowedFlags). This was discussed as launch plan while you were OOO

cmd/csi_driver/main.go
amacaskill: Resolving as this isn't needed since this isn't for the bucket scanner  / GKE GCSFuse profile feature (that Uriel is implementing).

cmd/csi_driver/main.go
```diff
@@ -50,7 +50,7 @@ var (
 	identityPool                   = flag.String("identity-pool", "", "The Identity Pool to authenticate with GCS API.")
 	identityProvider               = flag.String("identity-provider", "", "The Identity Provider to authenticate with GCS API.")
 	enableProfiling                = flag.Bool("enable-profiling", false, "enable the golang pprof at port 6060")
-	enableBucketScanner            = flag.Bool("enable-bucket-scanner", false, "enable the bucket scanner feature")
+	enableGcsfuseProfiles          = flag.Bool("enable-gcsfuse-profiles", false, "enable the gcsfuse profiles feature")
```
amacaskill: @uriel-guzman @chrisThePattyEater  We don't need to change the name of enable-bucket-scanner in this PR also. Let's leave `enable-bucket-scanner` as is, and that will be done in a follow up PR. 

Please only add"enable-gcsfuse-profiles-internal" flag, and leave existing flags as is

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -38,8 +38,9 @@ import (
 const DefaultName = "gcsfuse.csi.storage.gke.io"
 
 type FeatureScanner struct {
-	Enabled bool
-	Config  *scanner.ScannerConfig
+	Enabled         bool
+	EnabledInternal bool
```
amacaskill: Please remove EnabledInternal from FeatureScanner. It is not related to FeatureScanner in any way. Please put it in GCSDriverConfig. I think in the future, we can move feature specific flags to another type, but let's just add to GCSDriverConfig for now

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -201,14 +201,18 @@ func (mc *MountConfig) prepareMountArgs() {
 	}
 
 	invalidArgs := []string{}
-
+	// conditionally disallow flags, this loop ensures that order of args does not matter.
+	for _, arg := range mc.Options {
+		if arg == "enable-gcsfuse-profiles-internal=false" {
+			disallowedFlags["profile"] = true
+		}
+	}
```
amacaskill: The way you implemented this, you only pass this iff `enable-gcsfuse-profiles-internal=true`  so we will never hit this case


```
if s.driver.config.FeatureOptions.FeatureScanner.EnabledInternal {
	fuseMountOptions = joinMountOptions(fuseMountOptions, []string{fmt.Sprintf("%s=%v", util.EnableGcsfuseProfilesInternal, util.TrueStr)})
}
```

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -92,6 +92,7 @@ var disallowedFlags = map[string]bool{
 	"cache-dir":                            true,
 	"experimental-local-file-cache":        true,
 	"prometheus-port":                      true,
+	"profile":                              false,
```
amacaskill: This should be true  not false

pkg/sidecar_mounter/sidecar_mounter_config_test.go
```diff
@@ -144,6 +144,18 @@ func TestPrepareMountArgs(t *testing.T) {
...
 			expectedConfigMapArgs: defaultConfigFileFlagMap,
 		},
+		{
+			name: "temporary until we support profile flag - should return valid args with error correctly",
+			mc: &MountConfig{
+				BucketName: "test-bucket",
+				BufferDir:  "test-buffer-dir",
+				CacheDir:   "test-cache-dir",
+				ConfigFile: "test-config-file",
+				Options:    []string{"profile=true", "profile:true"},
```
amacaskill: nit: Profile is a profile (e.g. aiml-training, not a boolean)

Also I don't understand : "should return valid args with error correctly" what does error correctly mean? I think we can saw should return valid args with profile removed when specified, or something

pkg/sidecar_mounter/sidecar_mounter_config.go
chrisThePattyEater: yeah it was for testing purposes, I pushed it on accident

pkg/sidecar_mounter/sidecar_mounter_config.go
chrisThePattyEater: to be continued in post release 3.4 pr

test/e2e/testsuites/mount.go
```diff
@@ -251,4 +271,8 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
 		}
 		testCaseLongMountOptions()
 	})
+
+	ginkgo.It("should not pass --profile=aiml-training as a user-specified mountOption to gcsfuse when --profile is specified", func() {
+		testGcsfuseProfilesFlagFilters(specs.GCSFuseProfile)
+	})
```
amacaskill, line 278: I don't think we need this specs.GCSFuseProfile yet. Features in specs is so we can test a bunch of different test cases and trigger different features (skipCSIBUcketAccessCheck, metdataPrefetch, etc), for each test case.

For your case, we just need to do something like  testCaseLongMountOptions(). Within the testcase itself, testGcsfuseProfilesFlagFilters, you can set the similar [to this](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/506fca991485e59882b38aa8d247d918bc3a4bac/test/e2e/testsuites/mount.go#L183-L185C3). . Additionally, is it valid to pass  a mount option as both mount flag and config file format? I feel like that could cause some weird behavior, so it might be better to have separate case for --profile=aiml-training and profile:aiml-training

test/e2e/testsuites/mount.go
```diff
@@ -251,4 +271,8 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
 		}
 		testCaseLongMountOptions()
 	})
+
+	ginkgo.It("should not pass --profile=aiml-training as a user-specified mountOption to gcsfuse when --profile is specified", func() {
```
amacaskill: nit: "should not pass --profile=aiml-training as a user-specified mountOption to gcsfuse when --profile is specified" -> "should not pass profile as a user-specified mountOption to gcsfuse when profile is specified" 

Or if you want to add test case as mount flag format vs config file, you could say  "should not pass profile as a user-specified mountOption to gcsfuse when profile is specified in X format"

test/e2e/specs/specs.go
```diff
@@ -1242,3 +1243,13 @@ func (t *TestPod) VerifyDefaultingFlagsArePassed(namespace string, expectedMachi
...
 }
+
+func (t *TestPod) VerifyProfileFlagsAreNotPassed(namespace string) {
+	stdout, stderr, err := e2ekubectl.RunKubectlWithFullOutput(namespace, "logs", t.pod.Name, "-c", "gke-gcsfuse-sidecar")
+	framework.ExpectNoError(err,
+		"Error accessing logs from pod %v, but failed with error message %q\nstdout: %s\nstderr: %s",
+		t.pod.Name, err, stdout, stderr)
+
+	gomega.Expect(stdout).To(gomega.Not(gomega.ContainSubstring("profile")),
+		"Should not find profile flag string %q in stdout")
```
amacaskill: "Should not find profile flag string %q in stdout" -> "Should not find profile flag in stdout"

test/e2e/testsuites/mount.go
```diff
@@ -219,6 +219,27 @@ func (t *gcsFuseCSIMountTestSuite) DefineTests(driver storageframework.TestDrive
...
+		tPod := specs.NewTestPod(f.ClientSet, f.Namespace)
+		profileMO := "profile=aiml-training,profile:aiml-training"
+		tPod.SetupVolume(l.volumeResource, volumeName, mountPath, false, profileMO)
+
+		tPod.Create(ctx)
+
+		ginkgo.By("Checking pod is running")
+		tPod.WaitForRunning(ctx)
+
+		ginkgo.By("Checking pod is does not have --profile=aiml-training passed to gcsfuse")
```
amacaskill, line 236: Checking pod is does not have profile aiml-training passed to gcsfuse in mount flag or config file format. 

If it works you can leave it for now, and we can improve it in the follow up PR. I think it also makes sense in follow up PR to not test config file and mount flag format in the same test, because for the positive case when we do want to test passing profile to gcsfuse, you can't pass both as it will return a failure IIUC. 

In next PR, let's make VerifyProfileFlagsAreNotPassed more flexible to check for either format (indicated by parameters), and whether or not they should be set or not.

PR #985, opened 2025-09-30
dummy commit

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


test/e2e/testsuites/failed_mount.go
```diff
@@ -242,23 +240,28 @@ func (t *gcsFuseCSIFailedMountTestSuite) DefineTests(driver storageframework.Tes
...
-
 		ginkgo.By("Checking that the pod has failed mount error Unauthenticated")
 		tPod.WaitForFailedMountError(ctx, codes.Unauthenticated.String())
 		tPod.WaitForFailedMountError(ctx, "storage service manager failed to setup service: context deadline exceeded")
 	}
 
 	ginkgo.It("should fail when the specified service account does not have access to the GCS bucket", func() {
+		if pattern.VolType == storageframework.DynamicPV {
+			e2eskipper.Skipf("skip for volume type %v", storageframework.DynamicPV)
+		}
```
amacaskill, line 286: Why do we skip for these? please add comment.

test/e2e/testsuites/failed_mount.go
```diff
@@ -190,8 +189,7 @@ func (t *gcsFuseCSIFailedMountTestSuite) DefineTests(driver storageframework.Tes
 		}
 		v, err := version.ParseSemantic(gcsfuseVersionStr)
 		if configPrefix == specs.SkipCSIBucketAccessCheckAndInvalidVolumePrefix && (err != nil || v.AtLeast(version.MustParseSemantic("v2.9.0-gke.0"))) {
-			tPod.WaitForFailedMountError(ctx, codes.InvalidArgument.String())
-			tPod.WaitForFailedMountError(ctx, "name should be a valid bucket resource name")
+			tPod.WaitForFailedMountOrContainerError(ctx, "Error: failed to reserve container name", codes.InvalidArgument.String())
 		} else {
 			tPod.WaitForFailedMountError(ctx, codes.NotFound.String())
 			tPod.WaitForFailedMountError(ctx, "storage: bucket doesn't exist")
```
amacaskill, line 210: What's the benefit of using WaitForFailedMountError in some places and WaitForFailedMountOrContainerError in other places? In the future, when Sidecar bucket access check is the default, won't this confuse developers trying to write tests for their features? how will they know which one to use / can we make this easier for them somehow so they don't pick the wrong one, and waste cycles debugging?

test/e2e/testsuites/failed_mount.go
Sneha-at: Our stable overlay skips the tests for Dynamic PV. For dynamic PV the test is stuck at PV creation so pod is always in pending state.

test/e2e/testsuites/failed_mount.go
Sneha-at: I added a new function that first tries for FailedMountError which is the case currently but with sidecar bucket access check it will fail with CreateContainerError. So this new function expects the pod to error out with either of the FailedMount or CreateContainer error. The only case where this can be an issue is if we have sidecar bucket access check enabled and the tests still fail with FailedMountError which is unlikely since we always retry in Mount

test/e2e/specs/specs.go
```diff
@@ -260,6 +260,32 @@ func (t *TestPod) WaitForFailedContainerError(ctx context.Context, msg string) {
...
 
+func (t *TestPod) WaitForFailedMountOrContainerError(ctx context.Context, containerErrMsg string, mountErrMsg string) {
+
+	failedMountErr := e2eevents.WaitTimeoutForEvent(
+		ctx,
+		t.client,
+		t.namespace.Name,
+		fields.Set{"reason": events.FailedMountVolume}.AsSelector().String(),
+		mountErrMsg,
+		pollTimeoutSlow)
```
amacaskill: Do we need to wait 20min for this failedMount case? 

With the indefinite retry bucket access check feature in sidecar we never get terminal FailedMountError instead the pods fail to start with CreateContainerError so this test will always take 20min. Can we reduce timeout for both waiting for failed mount and waiting for createContainer? it should happen fast

test/e2e/testsuites/failed_mount.go
amacaskill: > I added a new function that first tries for FailedMountError which is the case currently but with sidecar bucket access check it will fail with CreateContainerError.So this new function expects the pod to error out with either of the FailedMount or CreateContainer error. 

Yea I understand the couple cases where this is needed, but I'm wondering why don't we just make all the code use WaitForFailedMountOrContainerError. I guess downside is we could un-intentionally make change and cause some thing to fail with CreateContainer instead of failedMount which we wouldn't catch, so I guess it is ok to leave it as is but please add comments explaining why in places where you changed to WaitForFailedMountOrContainerError. 

However, did we address all of the places where we need to change WaitForFailedMountError -> WaitForFailedMountOrContainerError?

I don't see testCaseNonExistentBucket being changed: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/06ab41c19fdfc2cf24dab249d6ac123268cb1837/test/e2e/testsuites/failed_mount.go#L128-L141

I see these cases, do we need to add all of these, or only a subset?
```
Cloud Storage FUSE CSI Driver.[It] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] failedMount [csi-skip-bucket-access-check] should fail when invalid mount options are passedCloud Storage FUSE CSI Driver.[It] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] failedMount [csi-skip-bucket-access-check] should fail when the gcsfuse processes got killed due to OOMCloud Storage FUSE CSI Driver.[It] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] failedMount [csi-skip-bucket-access-check] should fail when the sidecar container is not injectedCloud Storage FUSE CSI Driver.[It] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] failedMount [csi-skip-bucket-access-check] should fail when the specified GCS bucket does not existCloud Storage FUSE CSI Driver.[It] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] failedMount [csi-skip-bucket-access-check] should fail when the specified GCS bucket name is invalidCloud Storage FUSE CSI Driver.[It] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] failedMount [csi-skip-bucket-access-check] should fail when the specified service account does not have access to the GCS bucket
```

Sneha-at: I updated only the failing test cases. 
For testCaseNonExistentBucket case a. It was looking for FailedMountError which we don't expect so that is removed, post this it checks for GCS Fuse sidecar log line `bucket does not exist` and thus doesn't need this change. Though this brings up the question do we want to look for a log line for invalid bucket name as well instead of CreateContainerError?
```
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false OVERLAY=dev STAGINGVERSION=$STAGINGVERSION  REGISTRY=$REGISTRY E2E_TEST_FOCUS="failedMount.*"
...


Ran 39 of 442 Specs in 1241.539 seconds
SUCCESS! -- 39 Passed | 0 Failed | 0 Pending | 403 Skipped
```

test/e2e/specs/specs.go
Sneha-at: Updated this to 5 min. Tested for both `OVERLAY=dev` and stable
```
make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false  STAGINGVERSION=$STAGINGVERSION  REGISTRY=$REGISTRY E2E_TEST_FOCUS="failedMount.*" 
..
Ran 39 of 442 Specs in 101.400 seconds
SUCCESS! -- 39 Passed | 0 Failed | 0 Pending | 403 Skipped



make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=false BUILD_GCSFUSE_FROM_SOURCE=false  STAGINGVERSION=$STAGINGVERSION  REGISTRY=$REGISTRY E2E_TEST_FOCUS="failedMount.*"  OVERLAY=dev
..
Ran 39 of 442 Specs in 101.400 seconds
SUCCESS! -- 39 Passed | 0 Failed | 0 Pending | 403 Skipped
```

test/e2e/testsuites/failed_mount.go
amacaskill: > Though this brings up the question do we want to look for a log line for invalid bucket name as well instead of CreateContainerError?

If its easy to do so yes. Checking for the log line is a more accurate test since containerCreate error can happen for a lot of different reasons; however, we should make sure we don't change the sidecar GCS Fuse sidecar log line `bucket does not exist` otherwise it will cause this test to fail also, but hopefully we have e2e presubmit tests setup before that happens

test/e2e/specs/specs.go
Sneha-at: This is no longer relevant, I updated the logic to use env variable

test/e2e/testsuites/failed_mount.go
Sneha-at: Updated the logic for tests.

PR #986, opened 2025-10-01
Bucket Scanner Controller - Bucket Scan Override Status

uriel-guzman:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


pkg/scanner/scanner.go
```diff
@@ -1047,14 +1138,40 @@ func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume) (*bucketInfo, error)
...
+		dir:              dir,
+		onlyDirSpecified: onlyDirSpecified,
+	}
+
+	// Handle the override annotation, if set.
+	if bucketStatus, ok := pv.Annotations[annotationStatus]; ok && bucketStatus == scanOverride {
+		// Enforce required annotations for override mode and validate formats.
+		overrideInfo, err := parseOverrideStatus(pv)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "failed to validate arguments for PV %q with override mode: %v", pv.Name, err)
```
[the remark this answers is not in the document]
uriel-guzman: These are just nits, so ignoring.

examples/gcsfuse-profiles/pv-pvc-deployment.yaml
```diff
@@ -17,6 +17,12 @@ apiVersion: v1
 kind: PersistentVolume
 metadata:
   name: my-pv
+  annotations:
+    # Optional: Uncomment the following annotations to add custom bucket annotations and bypass the scanner.
+    # gke-gcsfuse/bucket-scan-status: "override"             # Required for "override": Will bypass the scanner with custom results.
+    # gke-gcsfuse/bucket-scan-num-objects: "42"              # Required for "override": Total number of objects in the bucket / directory.
```
amacaskill, line 23: In profile docs, we should provide the commands (gcloud, gsutil, etc) customers can run to get these numbers for their bucket. Not sure if your adding docs here, but please capture somewhere , wherever you are tracking those

Same for bucket-scan-total-size-bytes below. I think hns is straightforward, but maybe for completeness?

pkg/scanner/scanner.go
```diff
@@ -586,11 +597,15 @@ func (s *Scanner) syncPod(ctx context.Context, key string) error {
...
 		bucketI, err := s.checkPVRelevance(pv)
 		if err != nil {
-			return fmt.Errorf("error checking PV %q relevance for Pod %q: %v", pvName, key, err)
+			return fmt.Errorf("error checking PV %q relevance for Pod %q: %w", pvName, key, err)
 		}
 
-		if bucketI != nil {
-			klog.Infof("Pod %q uses relevant PV %q (PVC %q/%q)", key, pvName, namespace, pvcName)
+		// If bucketI is nil, PV is not relevant or TTL is not expired.
+		// If bucketI.isOverride is true, PV is relevant but scanning is bypassed.
```
amacaskill, line 604: >// If bucketI.isOverride is true, PV is relevant but scanning is bypassed.

Is it true bucketI.isOverride=true means PV is relevant AND scanning is bypassed? I thought it just means scanning is bypassed and `bucketI != nil && !bucketI.isOverride`  means PV is relevant but scanning is bypassed

amacaskill: sometimes, you can make this more readable by doing things like this so it reads more like english and you need less comments, but its up to you!
```
pvIsRelevant := bucketI != nil  
if !pvIsRelevant && !bucketI.isOverride` 
``

File: pkg/scanner/scanner.go [hunk 4]
```diff
@@ -669,6 +684,21 @@ func (s *Scanner) getDurationAttribute(pv *v1.PersistentVolume, attributeKey str
...
+func (s *Scanner) bypassScanForOverride(ctx context.Context, pv *v1.PersistentVolume, key string, bucketI *bucketInfo) error {
+	klog.Infof("PV %q is set to 'override' mode. Bypassing bucket scan and applying user-provided annotations.", key)
+
+	// The status annotation is already "override", but we patch it here along with
+	// the timestamp to mark the operation as complete and update the in-memory map.
+	s.eventRecorder.Eventf(pv, v1.EventTypeNormal, reasonScanOperationSucceeded, "Override mode detected for PV %q. Bypassing scan and using user-provided values: %d objects, %d bytes, HNS enabled: %t", pv.Name, bucketI.numObjects, bucketI.totalSizeBytes, bucketI.isHNSEnabled)
+	if patchErr := s.updatePVScanResult(ctx, pv, bucketI, scanOverride); patchErr != nil {
+		return patchErr
+	}
+	return nil // Remove since this is a complete and successful "scan" (bypass).
```
amacaskill: Remove from queue?

pkg/scanner/scanner.go
```diff
@@ -1000,10 +1036,65 @@ func (s *Scanner) updatePVScanResult(ctx context.Context, pv *v1.PersistentVolum
...
+			bucketI.totalSizeBytes = val
+		case annotationHNSEnabled:
+			val, err := strconv.ParseBool(pv.Annotations[annotationHNSEnabled])
+			if err != nil {
+				return nil, fmt.Errorf("invalid value for annotation %q: %w", key, err)
+			}
+			bucketI.isHNSEnabled = val
+		default:
+			// This should never happen, but it's safer to check.
+			return nil, fmt.Errorf("unexpected key for override mode requiredAnnotations: %q", key)
```
amacaskill: Wouldn't this be internal error then? You wrap all errors returned from this function as InvalidArgument, but if we ever hit this, should it be Internal so we don't filter it out from slo?

pkg/scanner/scanner.go
```diff
@@ -1047,14 +1138,40 @@ func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume) (*bucketInfo, error)
...
+	bucketI := &bucketInfo{
+		name:             bucketName,
+		dir:              dir,
+		onlyDirSpecified: onlyDirSpecified,
+	}
+
+	// Handle the override annotation, if set.
+	if bucketStatus, ok := pv.Annotations[annotationStatus]; ok && bucketStatus == scanOverride {
+		// Enforce required annotations for override mode and validate formats.
+		overrideInfo, err := parseOverrideStatus(pv)
```
amacaskill, line 1168: Do you want to do the annotation validation of formats in GCW so we don't let the PVs be created with invalid annotations in the first place? Just a thought

pkg/scanner/scanner.go
```diff
@@ -1156,22 +1289,7 @@ func (s *Scanner) addPV(obj any) {
...
-		klog.Errorf("PV %q: Error checking relevance for %q: %v", eventType, pv.Name, err)
-		return
-	}
-
-	if bucketI == nil {
-		klog.V(6).Infof("PV %q: %q - Not relevant", eventType, pv.Name)
-		return
-	}
-
+	klog.V(6).Infof("PV ADDED: %q", pv.Name)
```
amacaskill: Looks like debugging formatted statement? If this is intentional to be here, maybe we should add more context? I don't think if I saw PV ADDED in logs, I would know what this means

pkg/scanner/scanner.go
```diff
@@ -1086,30 +1203,46 @@ func (s *Scanner) checkPVRelevance(pv *v1.PersistentVolume) (*bucketInfo, error)
 
 	if found {
 		elapsed := timeNow().Sub(lastScanTime)
+
```
amacaskill, line 1188: nit: Several nested if/else blocks and if found / if !found checks that make it hard to follow. We could use a helper function here to simplify this and flatten the if /else nesting. https://engdoc.corp.google.com/eng/doc/tott/episodes/733.md?cl=head

You could put finding last scan time into its own helper function and then you would just be left with if last scan time was found, then do X otherwise, do Y. It's up to you how / if you simplify, but regardless, please add just a couple comments here at the beginning part where we try to find last scan then do stuff based on  ttl

pkg/scanner/scanner_test.go
```diff
@@ -424,6 +439,60 @@ func TestCheckPVRelevance(t *testing.T) {
 			wantRelevant: true,
 			wantBucket:   testBucketName,
 		},
+		{
+			name:           "Relevant - Override mode",
```
amacaskill: Nit: Include the Scenario and Expected Outcome in Test Names https://engdoc.corp.google.com/eng/doc/tott/episodes/660.md?cl=head

by looking at test name, you should be able to see scenario and  expected outcome. You only include scenario in these unit test names. You don't need to change all of your unit tests you have implemented to this style but for these in this PR, and going forward, try to follow this best practice. 

For example for "Irrelevant - Override mode - Missing Annotations" you would add "Irrelevant - Override mode - Missing Annotations - should return error" or whatever outcome is

examples/gcsfuse-profiles/pv-pvc-deployment.yaml
uriel-guzman: I added a note in the design document to include CLI commands / instructions to easily obtain these numbers.

amacaskill (approved):

pkg/scanner/scanner.go
uriel-guzman: > Is it true bucketI.isOverride=true means PV is relevant AND scanning is bypassed? I thought it just means scanning is bypassed and bucketI != nil && !bucketI.isOverride means PV is relevant but scanning is bypassed

`bucketI != nil` means that the PV is relevant. `bucketI.isOverride` means that the PV is relevant AND is using the override feature. So your understanding is correct.

We could make this easier to read by doing:
```
pvIsRelevant := bucketI != nil  
if pvIsRelevant && !bucketI.isOverride` 
```
Added to make this more readable.

pkg/scanner/scanner.go
uriel-guzman: Fixed.

pkg/scanner/scanner.go
uriel-guzman: Good catch. I just categorized this as Internal and the rest as InvalidArgument. Also removed the InvalidArgument categorization from the calling function.

pkg/scanner/scanner.go
uriel-guzman: Sure, that's a great suggestion. I created a bug to follow-up with this in the mutating webhook work as an improvement. For now, I'll just add TODOs here as reminders.

pkg/scanner/scanner.go
uriel-guzman: This is a verbosity level 6 log that won't appear in customer logs (we default to level 5 in the non-managed / managed driver). The purpose is only to debug the informer handlers if required (e.g. answer the question of whether a PV was added or not to workqueue).

Added a bit more detail (context about where this log is coming from, i.e. AddFunc Pod) to make it more clear.

pkg/scanner/scanner.go
uriel-guzman: Thanks for the suggestion. I refactored this code and put it in a helper function to minimize the nested statements.

pkg/scanner/scanner_test.go
uriel-guzman: Fixed.

PR #1224, opened 2026-02-17
Disable metrics on GCS Fuse CSI volume counts exceeding threshold of 10

thrivikram-karur-g:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR disables the metric collectors when the gcsfuse volumes count in the pod spec is more than a given threshold(10 in this case) as it is considered unbounded without that and contributes to theoretically infinite cardinality and is not accepted for making these metrics live. Hence, disabling metrics collector when the number of gcsfuse volumes exceeds given threshold has been incorporated, accordingly corresponding tests have been added.

Fixes #

```release-note
      Metrics from GCSFuse CSI driver wouldn't be exported if the number of GCS CSI volumes are more than the current threshold of 10
```

### DIFF
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -222,10 +222,12 @@ func main() {
 
 		clientset.ConfigurePodLister(ctx, *nodeID)
 		clientset.ConfigureNodeLister(ctx, *nodeID)
+		// Both PVs & PVCs are needed to check count of GCSFuseVolumes
+		clientset.ConfigurePVLister(ctx)
+		clientset.ConfigurePVCLister(ctx)
 
 		if featureOptions.FeatureGCSFuseProfiles.Enabled {
 			// Curently, only the gcsfuse profiles feature actually uses these listers.
-			clientset.ConfigurePVLister(ctx)
 			clientset.ConfigureSCLister(ctx)
 		}
 
--- a/deploy/base/node/node_setup.yaml
+++ b/deploy/base/node/node_setup.yaml
@@ -42,6 +42,14 @@ rules:
   - apiGroups: [""]
     resources: ["serviceaccounts"]
     verbs: ["get"]
+# Required to list PV annotations.
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch"]
+# Required to map PVC to PV from Pod.
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -158,6 +158,12 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 			},
 			Spec: corev1.PersistentVolumeSpec{
 				StorageClassName: pvObj.Spec.StorageClassName, // Required by the gcsfuse profiles feature to map PV to SC.
+				PersistentVolumeSource: corev1.PersistentVolumeSource{
+					CSI: &corev1.CSIPersistentVolumeSource{
+						Driver:       pvObj.Spec.CSI.Driver, // Required by cardinality mitigation to disable metric collectors when gcsfuse volumes exceed threshold
+						VolumeHandle: pvObj.Spec.CSI.VolumeHandle,
+					},
+				},
 			},
 		}, nil
 	}
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -44,10 +44,12 @@ type Interface interface {
 	ConfigurePodLister(ctx context.Context, nodeName string)
 	ConfigureNodeLister(ctx context.Context, nodeName string)
 	ConfigurePVLister(ctx context.Context)
+	ConfigurePVCLister(ctx context.Context)
 	ConfigureSCLister(ctx context.Context)
 	GetPod(namespace, name string) (*corev1.Pod, error)
 	GetNode(name string) (*corev1.Node, error)
 	GetPV(name string) (*corev1.PersistentVolume, error)
+	GetPVC(namespace string, name string) (*corev1.PersistentVolumeClaim, error)
 	GetSC(name string) (*storagev1.StorageClass, error)
 	CreateServiceAccountToken(ctx context.Context, namespace, name string, tokenRequest *authenticationv1.TokenRequest) (*authenticationv1.TokenRequest, error)
 	GetGCPServiceAccountName(ctx context.Context, namespace, name string) (string, error)
@@ -63,6 +65,7 @@ type Clientset struct {
 	podLister                 listersv1.PodLister
 	nodeLister                listersv1.NodeLister
 	pvLister                  listersv1.PersistentVolumeLister
+	pvcLister                 listersv1.PersistentVolumeClaimLister
 	scLister                  storagelisters.StorageClassLister
 	informerResyncDurationSec int
 }
@@ -170,6 +173,37 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 	informerFactory.WaitForCacheSync(ctx.Done())
 
 	c.pvLister = pvLister
+}
+
+func (c *Clientset) ConfigurePVCLister(ctx context.Context) {
+	trim := func(obj any) (any, error) {
+		pvcObj, ok := obj.(*corev1.PersistentVolumeClaim)
+		if !ok {
+			return obj, nil
+		}
+		return &corev1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      pvcObj.ObjectMeta.Name,
+				Namespace: pvcObj.ObjectMeta.Namespace,
+			},
+			Spec: corev1.PersistentVolumeClaimSpec{
+				VolumeName: pvcObj.Spec.VolumeName,
+			},
+		}, nil
+	}
+
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
+		c.k8sClients,
+		time.Duration(c.informerResyncDurationSec)*time.Second,
+		informers.WithTransform(trim),
+	)
+
+	pvcLister := informerFactory.Core().V1().PersistentVolumeClaims().Lister()
+
+	informerFactory.Start(ctx.Done())
+	informerFactory.WaitForCacheSync(ctx.Done())
+
+	c.pvcLister = pvcLister
 }
 
 func (c *Clientset) ConfigureSCLister(ctx context.Context) {
@@ -319,6 +353,14 @@ func (c *Clientset) GetPV(name string) (*corev1.PersistentVolume, error) {
 	}
 
 	return c.pvLister.Get(name)
+}
+
+func (c *Clientset) GetPVC(namespace, name string) (*corev1.PersistentVolumeClaim, error) {
+	if c.pvcLister == nil {
+		return nil, errors.New("pvc informer is not ready")
+	}
+
+	return c.pvcLister.PersistentVolumeClaims(namespace).Get(name)
 }
 
 func (c *Clientset) GetSC(name string) (*storagev1.StorageClass, error) {
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -148,24 +148,29 @@ func (c *Clientset) ConfigureNodeLister(ctx context.Context, nodeName string) {
 func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 	trim := func(obj any) (any, error) {
 		pvObj, ok := obj.(*corev1.PersistentVolume)
-		if !ok {
+
+		if !ok || pvObj == nil {
 			return obj, nil
 		}
-		return &corev1.PersistentVolume{
+
+		trimmedPV := &corev1.PersistentVolume{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:        pvObj.ObjectMeta.Name,
-				Annotations: pvObj.ObjectMeta.Annotations, // Required by the gcsfuse profiles feature to calculate smart cache recommendations.
+				Annotations: pvObj.ObjectMeta.Annotations,
 			},
 			Spec: corev1.PersistentVolumeSpec{
-				StorageClassName: pvObj.Spec.StorageClassName, // Required by the gcsfuse profiles feature to map PV to SC.
-				PersistentVolumeSource: corev1.PersistentVolumeSource{
-					CSI: &corev1.CSIPersistentVolumeSource{
-						Driver:       pvObj.Spec.CSI.Driver, // Required by cardinality mitigation to disable metric collectors when gcsfuse volumes exceed threshold
-						VolumeHandle: pvObj.Spec.CSI.VolumeHandle,
-					},
-				},
+				StorageClassName: pvObj.Spec.StorageClassName,
 			},
-		}, nil
+		}
+
+		if pvObj.Spec.CSI != nil {
+			trimmedPV.Spec.PersistentVolumeSource.CSI = &corev1.CSIPersistentVolumeSource{
+				Driver:       pvObj.Spec.CSI.Driver,
+				VolumeHandle: pvObj.Spec.CSI.VolumeHandle,
+			}
+		}
+
+		return trimmedPV, nil
 	}
 
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(
--- a/pkg/cloud_provider/clientset/fake.go
+++ b/pkg/cloud_provider/clientset/fake.go
@@ -151,8 +151,9 @@ func (c *FakeClientset) CreateNode(nodeConfig FakeNodeConfig) {
 func (c *FakeClientset) CreatePV(pvConfig FakePVConfig) {
 	pv := &corev1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   pvConfig.Name,
-			Labels: map[string]string{},
+			Name:        pvConfig.Name,
+			Labels:      map[string]string{},
+			Annotations: pvConfig.Annotations,
 		},
 		Spec: corev1.PersistentVolumeSpec{
 			StorageClassName: pvConfig.SCName,
@@ -170,7 +171,7 @@ func (c *FakeClientset) CreatePV(pvConfig FakePVConfig) {
 func (c *FakeClientset) CreatePVC(pvcConfig FakePVCConfig) {
 	pvc := &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: pvcConfig.Name,
+			Name:   pvcConfig.Name,
 			Labels: map[string]string{},
 		},
 	}
--- a/pkg/cloud_provider/clientset/fake.go
+++ b/pkg/cloud_provider/clientset/fake.go
@@ -73,6 +73,7 @@ func NewFakeClientset() *FakeClientset {
 	fakeClientSet.CreatePod(FakePodConfig{HostNetworkEnabled: false})
 	fakeClientSet.CreateNode(FakeNodeConfig{IsWorkloadIdentityEnabled: true})
 	fakeClientSet.CreatePV(FakePVConfig{})
+	fakeClientSet.CreatePVC(FakePVCConfig{})
 	fakeClientSet.CreateSC(FakeSCConfig{})
 
 	return fakeClientSet
@@ -208,8 +209,8 @@ func (c *FakeClientset) GetPV(name string) (*corev1.PersistentVolume, error) {
 }
 
 func (c *FakeClientset) GetPVC(namespace, name string) (*corev1.PersistentVolumeClaim, error) {
-	c.fakePV.ObjectMeta.Name = name
-	c.fakePV.ObjectMeta.Namespace = namespace
+	c.fakePVC.ObjectMeta.Name = name
+	c.fakePVC.ObjectMeta.Namespace = namespace
 
 	return c.fakePVC, nil
 }
--- a/pkg/cloud_provider/clientset/fake.go
+++ b/pkg/cloud_provider/clientset/fake.go
@@ -50,6 +50,7 @@ type FakePVConfig struct {
 type FakePVCConfig struct {
 	Name       string
 	VolumeName string
+	Namespace  string
 }
 
 type FakeSCConfig struct {
--- a/pkg/cloud_provider/clientset/fake.go
+++ b/pkg/cloud_provider/clientset/fake.go
@@ -45,6 +45,14 @@ type FakePVConfig struct {
 	SCName      string
 }
 
+type FakePVCSpec struct {
+	VolumeName string
+}
+
+type FakePVCConfig struct {
+	Spec FakePVCSpec
+}
+
 type FakeSCConfig struct {
 	Labels       map[string]string
 	Parameters   map[string]string
@@ -55,6 +63,7 @@ type FakeClientset struct {
 	fakePod  *corev1.Pod
 	fakeNode *corev1.Node
 	fakePV   *corev1.PersistentVolume
+	fakePVC  *corev1.PersistentVolumeClaim
 	fakeSC   *storagev1.StorageClass
 }
 
@@ -78,6 +87,8 @@ func (c *FakeClientset) ConfigurePodLister(_ context.Context, _ string) {}
 func (c *FakeClientset) ConfigureNodeLister(_ context.Context, _ string) {}
 
 func (c *FakeClientset) ConfigurePVLister(_ context.Context) {}
+
+func (c *FakeClientset) ConfigurePVCLister(_ context.Context) {}
 
 func (c *FakeClientset) ConfigureSCLister(_ context.Context) {}
 
@@ -145,6 +156,19 @@ func (c *FakeClientset) CreatePV(pvConfig FakePVConfig) {
 	c.fakePV.Spec.StorageClassName = pvConfig.SCName
 }
 
+func (c *FakeClientset) CreatePVC(pvcConfig FakePVCConfig) {
+	c.fakePVC = &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "",
+			Labels: map[string]string{},
+		},
+	}
+
+	c.fakePVC.Spec = corev1.PersistentVolumeClaimSpec{
+		VolumeName: pvcConfig.Spec.VolumeName,
+	}
+}
+
 func (c *FakeClientset) CreateSC(scConfig FakeSCConfig) {
 	c.fakeSC = &storagev1.StorageClass{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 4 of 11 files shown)


pkg/csi_driver/node.go
```diff
@@ -239,8 +239,14 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
-		klog.V(6).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
-		s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, args.bucketName)
+		gcsFuseVolumeCount := s.countGcsFuseVolumes(pod)
+
+		if gcsFuseVolumeCount > 10 {
+			klog.Warningf("Metrics collection is disabled for Pod %s/%s as the number of GCS FUSE volumes is %d, which is greater than the limit of 10.", pod.Namespace, pod.Name, gcsFuseVolumeCount)
+		} else {
+			klog.V(6).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
+			s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, args.bucketName)
+		}
```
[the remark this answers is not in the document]
thrivikram-karur-g: updated

pkg/csi_driver/node.go
```diff
@@ -539,3 +546,18 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
 	}
 	return ""
 }
+
+func (s *nodeServer) countGcsFuseVolumes(pod *corev1.Pod) int {
```
amacaskill: countGcsFuseVolumes only checks for CSI ephemeral volumes: https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-ephemeral

You also need to check for persistent volumes as the pod could have either: https://docs.cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-pv

Checking for PV is less trivial as you need to make additional API calls to get PVC / PV object. We don't just want to call  

```
pvc, err := s.driver.config.K8sClients.GetPersistentVolumeClaim(ctx, pod.Namespace, pvcName)
pv, err := s.driver.config.K8sClients.GetPersistentVolume(ctx, pvc.Spec.VolumeName)
```

We instead want to Update the Clientset Interface In pkg/cloud_provider/clientset/clientset.go, add methods for fetching PVCs to the main interface ([fetching PVs already exist](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/4b8e5781038c41a40631c3defde8ac3684167cf8/pkg/cloud_provider/clientset/clientset.go#L316)). When you implement GetPVC, you also need to implement a pvcLister. In K8s client-go, a Lister is a tool that allows you to fetch Kubernetes objects (like Pods or PersistentVolumes) from a local, in-memory cache rather than making a direct HTTP request to the Kubernetes API server for every query. This cache is continuously kept up-to-date in the background by an Informer, which "watches" the API server for any create, update, or delete events. This prevents making API call during the critical path of volume mounting (NodePublishVolume). @uriel-guzman Recently added one for GCSFuse profiles, so he can help you with this if you have questions.

pkg/csi_driver/node_test.go
```diff
@@ -654,3 +659,253 @@ func TestNodePublishVolumeEnableGCSFuseKernelParams(t *testing.T) {
 		})
 	}
 }
+
+func TestCountGcsFuseVolumes(t *testing.T) {
```
amacaskill: In addition to these CSI ephemeral volume tests, please add unit tests for persistent volumes + mixed GCSFuse volumes (meaning a pod that has both CSI ephemeral + persistent volumes in various orders). PV + CSI , CSI + PV, etc. 

See https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1224/changes#r2834314334 for details for what I mean  for CSI ephemeral volume vs Persistent volume

pkg/csi_driver/node.go
```diff
@@ -239,8 +240,14 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	// Register metrics collector.
 	// It is idempotent to register the same collector in node republish calls.
 	if s.driver.config.MetricsManager != nil && !args.disableMetricsCollection {
-		klog.V(6).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
-		s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, args.bucketName)
+		gcsFuseVolumeCount := s.countGcsFuseVolumes(pod)
+
+		if gcsFuseVolumeCount > maxGcsFuseVolumesForMetrics {
+			klog.Warningf("Metrics collection is disabled for Pod %s/%s as the number of GCS FUSE volumes is %d, which is greater than the limit of 10.", pod.Namespace, pod.Name, gcsFuseVolumeCount)
```
amacaskill: Replace with : 

```
			klog.Warningf("Metrics collection is disabled for Pod %s/%s as the number of GCS FUSE volumes is %d, which is greater than the limit of %d.", pod.Namespace, pod.Name, gcsFuseVolumeCount, maxGcsFuseVolumesForMetrics)
```

pkg/metrics/fake.go
```diff
@@ -17,13 +17,44 @@ limitations under the License.
...
+// UnregisterMetricsCollector records the unregistration of a metrics collector.
+func (f *FakeMetricsManager) UnregisterMetricsCollector(identifier string) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	delete(f.collectors, identifier)
+}
 
-func (*FakeMetricsManager) UnregisterMetricsCollector(_ string) {}
+// GetCollectors returns the map of registered collectors.
+func (f *FakeMetricsManager) GetCollectors() map[string]string {
```
amacaskill, line 53: From gemini: returning the map directly here creates a data race. In Go, maps are reference types, so returning `f.collectors` means the mutex only protects the pointer return, not the map's contents. Because this test suite uses `t.Parallel()`, if one test reads the map while another calls `RegisterMetricsCollector`, the test suite will panic with a concurrent map read/write error.

Check to see if returning a copy of the map instead fixes this

pkg/csi_driver/node_test.go
```diff
@@ -654,3 +659,253 @@ func TestNodePublishVolumeEnableGCSFuseKernelParams(t *testing.T) {
...
+			if tc.disableMetricsCollection {
+				req.VolumeContext["disableMetrics"] = "true"
+			}
+
+			// Call NodePublishVolume
+			_, err := ns.NodePublishVolume(context.Background(), req)
+			if err != nil {
+				// The fake clientset does not have the pod annotations,
+				// which will cause the sidecar check to fail.
+				// See if we can find a workaround.
```
amacaskill: "// See if we can find a workaround." 

What do we need to fix here? Can you please provide details?

thrivikram-karur-g: No, I do not think we need any fix here, it is an older comment, removing it.

pkg/metrics/fake.go
thrivikram-karur-g: Makse sense, yes, copying the map should fix this.

thrivikram-karur-g: have updated

pkg/csi_driver/node.go
thrivikram-karur-g: updated

pkg/csi_driver/node.go
thrivikram-karur-g: thanks for all this context, this really helps, working on adding support for PVCLister

thrivikram-karur-g: Updated here but still need to add/update unittests for mixed volumes (ephemeral & persistent).

pkg/csi_driver/node_test.go
thrivikram-karur-g: Added unittests for both

cmd/csi_driver/main.go
```diff
@@ -222,10 +222,12 @@ func main() {
 
 		clientset.ConfigurePodLister(ctx, *nodeID)
 		clientset.ConfigureNodeLister(ctx, *nodeID)
+		// Both PVs & PVCs are needed to check count of GCSFuseVolumes
+		clientset.ConfigurePVLister(ctx)
```
thrivikram-karur-g, line 226: cc: @uriel-guzman  , could this cause any issues or any kind of perf concerns ?

deploy/base/node/node_setup.yaml
```diff
@@ -42,6 +42,14 @@ rules:
   - apiGroups: [""]
     resources: ["serviceaccounts"]
     verbs: ["get"]
+# Required to list PV annotations.
```
thrivikram-karur-g, line 45: @uriel-guzman  , wanted to check if this can result in any kind of backward incompatibility changes with release as permissions are being upgraded.

cmd/csi_driver/main.go
uriel-guzman: No, the PV lister was already there, and profiles is now enabled by default, so this doesn't actually change anything.

deploy/base/node/node_setup.yaml
uriel-guzman: Adding more permissions won't cause any issues. Removing them yes, but this is not the case, so this should be fine.

pkg/cloud_provider/clientset/fake.go
```diff
@@ -43,6 +43,13 @@ type FakePodConfig struct {
 type FakePVConfig struct {
 	Annotations map[string]string
 	SCName      string
+	DriverName  string
+	Name        string
+}
+
+type FakePVCConfig struct {
+	Name       string
```
uriel-guzman, line 51: Doesn't the PVC also need a namespace?

pkg/csi_driver/node.go
```diff
@@ -45,6 +46,7 @@ const (
 	// parameters file is polled and any changes to kernel parameter files are applied.
 	GCSFuseKernelParamsFilePollInterval = time.Second * 5
 	FuseMountType                       = "fuse"
+	maxGcsFuseVolumesForMetrics         = 10
```
uriel-guzman: Nit: `maxGcsFuseVolumesForMetrics` -> `maxGCSFuseVolumesForMetrics`

pkg/csi_driver/node.go
```diff
@@ -533,3 +541,51 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+		return gcsFuseVolumeCount
+	}
+
+	for _, v := range pod.Spec.Volumes {
+		// Count ephemeral gcsfuse volumes
+		if v.PersistentVolumeClaim == nil && v.CSI != nil && v.CSI.Driver == s.driver.config.Name {
+			gcsFuseVolumeCount++
+			continue
+		}
+
```
uriel-guzman, line 560: Can't this just be simplified to:

```
if v.CSI != nil && v.CSI.Driver == s.driver.config.Name {
			gcsFuseVolumeCount++
			continue
		}
```

The PVC doesn't have a `CSI` field, so this PVC nil checking seems redundant and more complicated than necessary: https://docs.cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-pv#create-a-persistentvolumeclaim. 

For example, this webhook code is checking if a volume is an ephemeral volume by just checking `volume.CSI` and `volume.CSI.Driver == gcsFuseCsiDriverName`, so this should be fine: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/volumes.go#L41

pkg/csi_driver/node.go
```diff
@@ -533,3 +541,51 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+		if v.PersistentVolumeClaim == nil {
+			continue
+		}
+
+		// Count persistent gcsfuse volumes
+		pvc, err := s.k8sClients.GetPVC(pod.Namespace, v.PersistentVolumeClaim.ClaimName)
+
+		// If an error occurs in getting information regarding the persistent volume, we ignore it's impact for metrics cardinality volume count restriction
+		if err != nil {
+			if apierrors.IsNotFound(err) {
```
uriel-guzman, line 570: You're handling the case where the API server returns `NotFound` correctly by skipping and logging a warning, but what about every other error? This GET call could fail for any reason, including missing permissions or internal errors. It seems like they are just being ignored and not even logged, masking real issues.

I think instead, this should return an an internal error. Otherwise, we won't get alerted if this is behaving incorrectly.

pkg/csi_driver/node.go
```diff
@@ -533,3 +541,51 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+			}
+
+			continue
+		}
+
+		pv, err := s.k8sClients.GetPV(pvc.Spec.VolumeName)
+
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("pv %q not found: %v", pvc.Spec.VolumeName, err)
```
uriel-guzman, line 583: Same comment about lack of error handling for this scenario.

pkg/csi_driver/node.go
```diff
@@ -533,3 +541,51 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+		// If an error occurs in getting information regarding the persistent volume, we ignore it's impact for metrics cardinality volume count restriction
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("pvc %q not found: %v", v.Name, err)
+			}
+
+			continue
+		}
+
+		pv, err := s.k8sClients.GetPV(pvc.Spec.VolumeName)
```
uriel-guzman, line 579: Can `pvc.Spec` be nil? Otherwise, this will panic if not checked.

pkg/csi_driver/node.go
```diff
@@ -533,3 +541,51 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("pv %q not found: %v", pvc.Spec.VolumeName, err)
+			}
+
+			continue
+		}
+
+		if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == s.driver.config.Name {
```
uriel-guzman, line 591: Can `pv.Spec` be nil? Otherwise, this will panic if not checked.

pkg/csi_driver/node_test.go
```diff
@@ -654,3 +659,357 @@ func TestNodePublishVolumeEnableGCSFuseKernelParams(t *testing.T) {
 		})
 	}
 }
+
+func TestCountGcsFuseVolumes(t *testing.T) {
+	testEnv := initTestNodeServer(t)
+	ns, ok := testEnv.ns.(*nodeServer)
+	if !ok {
+		t.Fatalf("Failed to cast NodeServer to *nodeServer")
+	}
```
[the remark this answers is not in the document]
uriel-guzman: +1 please add unit tests for PVs too.

pkg/csi_driver/node.go
```diff
@@ -533,3 +541,51 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+		}
+
+		// Count persistent gcsfuse volumes
+		pvc, err := s.k8sClients.GetPVC(pod.Namespace, v.PersistentVolumeClaim.ClaimName)
+
+		// If an error occurs in getting information regarding the persistent volume, we ignore it's impact for metrics cardinality volume count restriction
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("pvc %q not found: %v", v.Name, err)
+			}
```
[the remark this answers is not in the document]
uriel-guzman: +1 to this comment. Please fix.

pkg/csi_driver/node.go
thrivikram-karur-g: Not very sure but I will add it just in case.

pkg/csi_driver/node.go
thrivikram-karur-g: will check

pkg/csi_driver/node.go
thrivikram-karur-g: I agree that the bare minimum we should have is a visibility into the errors happening but at the same time I am not sure we should return an internal error in a critical path like NodePublishVolume for a non critical usecase of metrics collector registration.

thrivikram-karur-g: I will make the changes accordingly, let me know if you still have suggestions or workaround around this, we can discuss when we catchup on call.

pkg/csi_driver/node.go
thrivikram-karur-g: same reply as above

pkg/csi_driver/node.go
thrivikram-karur-g: I have checked, when you add such a check, go complier says below:

invalid operation: pv.Spec != nil (mismatched types "k8s.io/api/core/v1".PersistentVolumeSpec and untyped nil)compilerMismatchedTypes

nil is a predeclared identifier representing the zero value for a pointer, channel, func, interface, map, or slice type.

Explanation from why this check wouldn't be needed.

type PersistentVolume struct {
    // ...
    Spec PersistentVolumeSpec // <-- No asterisk (*) here!
    // ...
}

Because it is not a pointer (*PersistentVolumeSpec), it cannot be nil. Its "zero value" is simply an empty struct where all of its internal fields are set to their zero values. When you type pv.Spec != nil, the Go compiler throws its hands up because it's like asking if the number 5 is nil

If you already know pv is not nil (e.g., you successfully fetched it from the API), then pv.Spec is guaranteed to exist. You can safely delete the pv.Spec != nil part of your if statement.

pkg/csi_driver/node.go
thrivikram-karur-g: Agree, I too felt it was unnecessary but I guess in adding tests or some existing tests had inaccurate I ended up adding this, I will deprecate it.

pkg/csi_driver/node.go
thrivikram-karur-g: As already discussed above, both pvc.Spec & pv.Spec cannot be nil

pkg/csi_driver/node_test.go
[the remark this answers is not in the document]
thrivikram-karur-g: Added it.

thrivikram-karur-g:
> Please fix sanitiy test, you need to add your NewFakeMetricsManager() to the sanity test here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/test/sanity/sanity_test.go#L73, otherwise you're creating a panic because of an entry to the nil map of collectors.

Fixed it

pkg/csi_driver/node.go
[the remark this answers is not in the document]
thrivikram-karur-g: fixed

pkg/csi_driver/node.go
thrivikram-karur-g: resolving

pkg/csi_driver/node.go
thrivikram-karur-g: added it, if error exists in any flow, disabling registering the metrics collector. Let me know if you think otherwise if error needs to be raised and nodePublishVolume itself needs to fail.

pkg/cloud_provider/clientset/fake.go
thrivikram-karur-g: yeah, added. wil try using it in fakeClientSet as well.

pkg/csi_driver/node.go
thrivikram-karur-g: fixed

pkg/csi_driver/node.go
thrivikram-karur-g: I removed and seems like unittests atleast are running fine.

pkg/csi_driver/node.go
thrivikram-karur-g: have made the changes but am not raising error and failing the critical nodePublishVolume, if you still think we need to, can make the change and we can discuss this on call.

uriel-guzman: Ok, I agree we shouldn't block a critical path on a non-critical failure. If the PVC fails to be get, then this will just cause the metrics collector to be disabled and log an error, which I think is reasonable, since some customers might not even care about metrics and disabling the collector is good mitigation for when we can't know the number of PVs because of some API failure.

pkg/csi_driver/node.go
```diff
@@ -533,3 +543,55 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+		}
+
+		if v.PersistentVolumeClaim == nil {
+			continue
+		}
+
+		// Count persistent gcsfuse volumes
+		pvc, err := s.k8sClients.GetPVC(pod.Namespace, v.PersistentVolumeClaim.ClaimName)
+
+		// If an error occurs in getting information regarding the persistent volume, we ignore it's impact for metrics cardinality volume count restriction
```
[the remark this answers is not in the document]
uriel-guzman: +1 please fix the nit

uriel-guzman (approved):
Leaving my approval contingent on successful E2E tests. Please don't submit until @amacaskill approves.

pkg/csi_driver/node.go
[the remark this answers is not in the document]
thrivikram-karur-g: fixed it, thanks

thrivikram-karur-g:
Ran the e2e tests with the command.

make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false \
        E2E_TEST_BUILD_DRIVER=true \
        BUILD_GCSFUSE_FROM_SOURCE=true \
        REGISTRY=$REGISTRY \
        GCSFUSE_TAG="v3.7.1"

Below are the test results:

[ReportAfterSuite] PASSED [4155.235 seconds]
autogenerated by Ginkgo

Summarizing 1 Failure:
  [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: Pre-provisioned PV (default fs)] metrics [It] should emit metrics until a threshold of 10 gcsfuse volumes
  /home/thrivikramks_google_com/workspace/gcs-fuse-csi-driver/test/e2e/testsuites/metrics.go:259

Ran 347 of 481 Specs in 10382.277 seconds
FAIL! -- 346 Passed | 1 Failed | 0 Pending | 134 Skipped


Ginkgo ran 1 suite in 2h53m14.778640261s

Test Suite Failed
F0228 12:35:38.830157 1649086 main.go:136] Failed to run e2e test: failed to run e2e tests with ginkgo: exit status 1
make: *** [Makefile:313: e2e-test] Error 255
Completed running e2e tests for testing the PR

Have rerun the failed metrics emit test using the below command.

make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false \
        E2E_TEST_BUILD_DRIVER=true \
        BUILD_GCSFUSE_FROM_SOURCE=true \
        REGISTRY=$REGISTRY \
        GCSFUSE_TAG="v3.7.1" \
        E2E_TEST_FOCUS=.*emit.*

Below are the test results:

[ReportAfterSuite] PASSED [297.662 seconds]
autogenerated by Ginkgo

Ran 8 of 481 Specs in 297.542 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 473 Skipped


Ginkgo ran 1 suite in 5m1.99969443s
Test Suite Passed
Completed running e2e tests for testing the PR

So seems like the test failure was momentary.

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -155,6 +158,12 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 			},
 			Spec: corev1.PersistentVolumeSpec{
 				StorageClassName: pvObj.Spec.StorageClassName, // Required by the gcsfuse profiles feature to map PV to SC.
+				PersistentVolumeSource: corev1.PersistentVolumeSource{
+					CSI: &corev1.CSIPersistentVolumeSource{
+						Driver:       pvObj.Spec.CSI.Driver, // Required by cardinality mitigation to disable metric collectors when gcsfuse volumes exceed threshold
```
amacaskill: From gemini: In the ConfigurePVLister function, you added a trim transformer to cache the necessary PV fields. However, you are unconditionally accessing pvObj.Spec.CSI.Driver and pvObj.Spec.CSI.VolumeHandle. If a Kubernetes cluster contains any PersistentVolumes that are not CSI-based (for example, hostPath, nfs, or cloud provider-specific legacy volumes), pvObj.Spec.CSI will be nil. The informer will immediately panic with a nil pointer dereference when attempting to access pvObj.Spec.CSI.Driver. 

Recommendation: Add a nil check for the CSI source before assigning it to the stripped PV object.

pkg/csi_driver/node.go
```diff
@@ -533,3 +543,55 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+
+		pv, err := s.k8sClients.GetPV(pvc.Spec.VolumeName)
+
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("pv %q not found: %v", pvc.Spec.VolumeName, err)
+				continue
+			}
+
+			klog.Errorf("internal error getting PV %q: %v. GCS Fuse metric count may be inaccurate", pvc.Spec.VolumeName, err)
```
amacaskill: Instead of "internal error getting PV %q: %v. GCS Fuse metric count may be inaccurate" can you just say "internal error getting PV %q: %v. Setting GCS Fuse metric count to 0"

pkg/csi_driver/node.go
```diff
@@ -533,3 +543,55 @@ func gcsFuseSidecarContainerImage(pod *corev1.Pod) string {
...
+		pvc, err := s.k8sClients.GetPVC(pod.Namespace, v.PersistentVolumeClaim.ClaimName)
+
+		// A NotFound error is tolerated, but other errors will abort the count and disable metrics.
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("pvc %q not found: %v", v.PersistentVolumeClaim.ClaimName, err)
+				continue
+			}
+
+			klog.Errorf("internal error getting PVC %q: %v. GCS Fuse metric count may be inaccurate", v.PersistentVolumeClaim.ClaimName, err)
```
amacaskill: Same thing https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1224/changes#r2875470170

amacaskill (approved):

pkg/cloud_provider/clientset/clientset.go
thrivikram-karur-g: updated, thanks

pkg/csi_driver/node.go
thrivikram-karur-g: updated

pkg/csi_driver/node.go
thrivikram-karur-g: updated

flagbug:
We've upgraded our GKE cluster from version `1.34.6-gke.1154000` to version `1.35.3-gke.1389000` and ever since we're seeing all our `gcsfusecsi-node` instances fail because the `gcs-fuse-csi-driver` runs into its `100M` memory limit.

I have a strong suspicion that the change in this PR that introduces a PVC listener is the root cause, as we have around 40k PVCs in our cluster and I've seen listeners of other components misbehave before with that many PVCs.

This is more of a FYI that we're running into this issue, as I understand that we have an exceptional amount of PVCs in our cluster, but it also means that we now have to find a solution to *massively* increase the memory limits of `gcs-fuse-csi-driver`.

PR #1297, opened 2026-04-11
Set GOMEMLIMIT based on container memory limits

kislaykishore:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Configure the GOMEMLIMIT environment variable in the node daemonset to automatically match the container's memory limits. This helps the Go runtime optimize garbage collection and avoid OOM kills when running in constrained environments.

Fixes #

```release-note

```

### DIFF
--- a/deploy/base/node/node.yaml
+++ b/deploy/base/node/node.yaml
@@ -80,6 +80,10 @@ spec:
               value: ""
             - name: IDENTITY_POOL
               value: ""
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                resource: limits.memory
           volumeMounts:
             - name: kubelet-dir
               mountPath: /var/lib/kubelet/pods


amacaskill:
We need to set GOMEMLIMIT to the container memory limit, with 5-10% headroom to account for the other binary memory usage that the runtime isn't aware of. See cl/485451455 for more details. 

Setting 100% is probably worse than leaving it unlimited (https://go.dev/doc/gc-guide, https://yaqs.corp.google.com/eng/q/3887607034002014208):
- When you don't set it, the application might occasionally OOM if memory scales too fast, but the application will run at full speed until that happens. If you set GOMEMLIMIT to 100%, the Go runtime does not account for memory used outside the heap (such as the binary itself, CGO memory, and OS caches). As the heap usage approaches 100%, the garbage collector will run almost continuously to try and keep the memory down, causing a "death spiral" where performance drops to a crawl. 
- Furthermore, Go has a built-in escape hatch for this exact scenario: if the runtime detects that it is spending over 50% of its CPU time doing garbage collection, it will stop enforcing the GOMEMLIMIT to allow the program to do actual work. Because the unmanaged memory (which the GC can't control) pushes the total memory usage over the container limit anyway, the container will still be OOM killed. 
- So 100% gives you the worst of both worlds: severe application hangs followed by an OOM crash.

So to fix this, we need to set GOMEMLIMIT to be 95% of whatever the memory limit is, in both non managed driver (what you have updated here), as well as the managed driver yaml. This probably shouldn't be a hard coded value, because we currently have customers who deploy this mutating webhook to [increase the memory limit of the driver](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/384/changes). If you hardcode GOMEMLIMIT to 95MiB in the daemonset manifest:
- When a customer uses the webhook to increase the container memory limit to 200Mi, the container will technically have 200MiB of RAM available.
- However, the GOMEMLIMIT environment variable will still be stuck at 95MiB.
- As a result, the Go runtime will not realize it has more memory available and will continue to aggressively run garbage collection to stay under 95MiB, completely wasting the extra resources the customer provisioned.

To handle this properly, we can import the github.com/KimMachineGun/automemlimit library directly in the driver's Go main.go. This library automatically inspects the Linux cgroups of the container at runtime and sets GOMEMLIMIT to a safe percentage (90% by default) of the actual memory limit applied to the container. This gracefully handles any mutations made by webhooks without requiring complex manifest changes.

You cannot use the Kubernetes Downward API (like done in this PR) to do this, because while the Downward API correctly reflects the mutated limit, Kubernetes does not support mathematical operations (like calculating 95%) directly within the valueFrom environment variable fields.

kislaykishore:
Thanks! That makes sense. I will make the changes to use the github.com/KimMachineGun/automemlimit library and keep a threshold of 10%.

> We need to set GOMEMLIMIT to the container memory limit, with 5-10% headroom to account for the other binary memory usage that the runtime isn't aware of. See cl/485451455 for more details.
> 
> Setting 100% is probably worse than leaving it unlimited (https://go.dev/doc/gc-guide, https://yaqs.corp.google.com/eng/q/3887607034002014208):
> 
> * When you don't set it, the application might occasionally OOM if memory scales too fast, but the application will run at full speed until that happens. If you set GOMEMLIMIT to 100%, the Go runtime does not account for memory used outside the heap (such as the binary itself, CGO memory, and OS caches). As the heap usage approaches 100%, the garbage collector will run almost continuously to try and keep the memory down, causing a "death spiral" where performance drops to a crawl.
> * Furthermore, Go has a built-in escape hatch for this exact scenario: if the runtime detects that it is spending over 50% of its CPU time doing garbage collection, it will stop enforcing the GOMEMLIMIT to allow the program to do actual work. Because the unmanaged memory (which the GC can't control) pushes the total memory usage over the container limit anyway, the container will still be OOM killed.
> * So 100% gives you the worst of both worlds: severe application hangs followed by an OOM crash.
> 
> So to fix this, we need to set GOMEMLIMIT to be 95% of whatever the memory limit is, in both non managed driver (what you have updated here), as well as the managed driver yaml. This probably shouldn't be a hard coded value, because we currently have customers who deploy this mutating webhook to [increase the memory limit of the driver](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/384/changes). If you hardcode GOMEMLIMIT to 95MiB in the daemonset manifest:
> 
> * When a customer uses the webhook to increase the container memory limit to 200Mi, the container will technically have 200MiB of RAM available.
> * However, the GOMEMLIMIT environment variable will still be stuck at 95MiB.
> * As a result, the Go runtime will not realize it has more memory available and will continue to aggressively run garbage collection to stay under 95MiB, completely wasting the extra resources the customer provisioned.
> 
> To handle this properly, we can import the github.com/KimMachineGun/automemlimit library directly in the driver's Go main.go. This library automatically inspects the Linux cgroups of the container at runtime and sets GOMEMLIMIT to a safe percentage (90% by default) of the actual memory limit applied to the container. This gracefully handles any mutations made by webhooks without requiring complex manifest changes.
> 
> You cannot use the Kubernetes Downward API (like done in this PR) to do this, because while the Downward API correctly reflects the mutated limit, Kubernetes does not support mathematical operations (like calculating 95%) directly within the valueFrom environment variable fields.

Thanks! That makes sense. I will make the changes to use the github.com/KimMachineGun/automemlimit library and keep a threshold of 10%.

PR #1307, opened 2026-04-15
Enable cloud profiler in GCSFuse CSI driver

yaozile123:
**What type of PR is this?**
/kind feature
This pull request implements the strategy to use the Kubernetes Pod Name and Pod UID as the Cloud Profiler `ServiceVersion` for both the Node Driver, Sidecar Mounter and GCSFuse process, and enables the profiler by default in the Node Driver deployment.
Key implementations:
1. **Node Driver**:
    - Enabled Cloud Profiler by adding the argument `--enable-cloud-profiler-for-driver=true` in the node.yaml.
    - Injected the `POD_NAME` and `POD_UID` environment variables via the Downward API.
    - Updated the driver initialization to read both `POD_NAME` and `POD_UID` and assign `<pod_name>_<pod_uid>` to `ServiceVersion`.
2. **Sidecar Mounter**:
    - Updated the profiler initialization to use the `PodName` and `PodUID` passed from the driver as the `ServiceVersion` (formatted as `<pod_name>_<pod_uid>`) to ensure strict uniqueness across instances.
3. **GCSFuse Process**:
    - When the user enables the profiler for the sidecar, the sidecar mounter now automatically injects the CLI flag `--enable-cloud-profiler` to the spawned `gcsfuse` process.
    - It also injects a custom label `--cloud-profiler-label` formatted as `<pod_name>_<pod_uid>`.
4. **E2E Test Suite**:
    - Added a new test suite `test/e2e/testsuites/cloud_profiler.go` to verify that profiles are successfully generated and uploaded to the Cloud Profiler API for both the sidecar and gcsfuse processes.
    - Disabled the test suite on managed driver until the CRP component released.
    - Changed polling to exponential backoff when creating iam roles to avoid conflicts when concurrent tests bind IAM roles simultaneously.
    - In scenarios where the current project has large amount of profiler data, the Cloud Profiler GCSFuse integration test frequently timed out in 10-minute threshold.Bumped timeout from 10m to 30m. 

**Dependency Updates**:
- Added `google.golang.org/api/cloudprofiler/v2` package and executed `go mod tidy` and `go mod vendor` in the `test` directory.

Fixes #

We use the combination of Kubernetes Pod Name and Pod UID as the service version for Cloud Profiler, the current cloud profiler testsuite will disabled on managed driver until CRP component released.

Ran the test on non-managed driver:
```
make build-image-and-push-multi-arch REGISTRY=$REGISTRY STAGINGVERSION=$STAGINGVERSION

make e2e-test \
    E2E_TEST_GINKGO_PROCS=20 \
    REGISTRY=$REGISTRY \
    STAGINGVERSION=$STAGINGVERSION

Ran 433 of 558 Specs in 12239.952 seconds
  SUCCESS! -- 433 Passed | 0 Failed | 0 Pending | 125 Skipped


  Ginkgo ran 1 suite in 3h24m14.193229708s
  Test Suite Passed
```

I also ran the test on the internal boskos project, and the test all passed.

```release-note
NONE

### CODE REVIEW

File: pkg/sidecar_mounter/sidecar_mounter_config.go [hunk 1]
```diff
@@ -340,6 +342,30 @@ func (mc *MountConfig) prepareMountArgs() {
...
+		hostname, err := os.Hostname()
+		if err != nil {
+			klog.Warningf("failed to get hostname for cloud profiler: %v", err)
+			hostname = "unknown-host"
+		}
+
+		// Use first 4 characters of UUID as random suffix to ensure uniqueness.
+		randomSuffix := uuid.New().String()[:4]
+
+		customLabel := fmt.Sprintf("%s-%s", hostname, randomSuffix)
```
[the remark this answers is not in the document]
yaozile123: done

cmd/sidecar_mounter/main.go
```diff
@@ -84,13 +85,21 @@ func main() {
...
+				Service:        "gke-gcsfuse-sidecar",
+				ServiceVersion: hostname + "-" + uuid.New().String()[:4],
 			}
 			if err := profiler.Start(cfg); err != nil {
 				klog.Errorf("Errored while starting cloud profiler, got %v", err)
 			} else {
-				klog.Infof("Running cloud profiler on %s", cfg.Service)
+				klog.Infof("Running cloud profiler on %s with version %s", cfg.Service, cfg.ServiceVersion)
 			}
 		}
```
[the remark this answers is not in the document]
yaozile123: done

cmd/csi_driver/main.go
```diff
@@ -132,13 +132,16 @@ func main() {
...
-			Service: "gcs-fuse-csi-driver",
+			Service:        "gcs-fuse-csi-driver",
+			ServiceVersion: serviceVersion,
 		}
 		if err := profiler.Start(cfg); err != nil {
 			klog.Errorf("Errored while starting cloud profiler, got %v", err)
 		} else {
-			klog.Infof("Running cloud profiler on %s", cfg.Service)
+			klog.Infof("Running cloud profiler on %s with version %s", cfg.Service, cfg.ServiceVersion)
 		}
```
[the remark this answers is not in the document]
amacaskill: Please address this comment. If it isn't needed, explain why.  Seems l ike the issue is still there.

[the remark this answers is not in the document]
yaozile123: The pod name and UID will be passed from node driver via UDS. Can the pod name or UID be empty in this case?

pkg/csi_driver/node.go
```diff
@@ -210,7 +210,11 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	if args.enableCloudProfilerForSidecar && gcsFuseSidecarImage != "" && s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarCloudProfilerMinVersion) {
-		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.EnableCloudProfilerForSidecarConst + "=" + strconv.FormatBool(args.enableCloudProfilerForSidecar)})
+		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{
+			util.EnableCloudProfilerForSidecarConst + "=" + strconv.FormatBool(args.enableCloudProfilerForSidecar),
+			util.PodNameConst + "=" + vc[VolumeContextKeyPodName],
+			util.PodUIDConst + "=" + string(pod.UID),
```
amacaskill, line 216: Seems like you are passing pod UID / pod Name from node driver to sidecar mounter? Is that correct? 

The sidecar mounter and driver pods are different pods, so we can't use the same pod UID

yaozile123: Based on line [103-104](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/273d4c55705b73466a1601f26d55fea3132b1b45/pkg/csi_driver/node.go#L102-L104), the pod object should be the workload pod not the driver pod right？The volume context should come from the workload pod rather than node driver pod.

cmd/csi_driver/main.go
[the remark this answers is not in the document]
amacaskill: I don't know, but if it is empty, I don't think we want to start cloud profiler since we won't be able to tell what its profiling. Please add the logic to not enable if both pod_name and pod_UID are empty. If only 1 is empty, I think its fine to still enable it.

pkg/csi_driver/node.go
amacaskill: Oops I missed that thank you

pkg/csi_driver/node.go
```diff
@@ -210,7 +210,11 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	if args.enableCloudProfilerForSidecar && gcsFuseSidecarImage != "" && s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarCloudProfilerMinVersion) {
```
amacaskill: Can we return false from isSidecarVersionSupportedForGivenFeature if gcsFuseSidecarImage != "" for all cases, then clean up the code in pkg/csi_driver/node.go everywhere it calls isSidecarVersionSupportedForGivenFeature

amacaskill: Do we need to change SidecarCloudProfilerMinVersion now, since the logic to enable it properly isn't available until this PR? https://github.com/yaozile123/gcs-fuse-csi-driver/blob/90fe88b4d14f70d6107506782a92d8af49045293/pkg/csi_driver/utils.go#L80

test/e2e/specs/testdriver.go
```diff
@@ -168,8 +169,8 @@ func (n *GCSFuseCSITestDriver) PrepareTest(ctx context.Context, f *e2eframework.
 		e2eframework.Logf("Creating Project IAM Policy Bindings for Cloud Profiler (Agent and User)...")
 		profilerAgentBinding.Create(ctx)
 		profilerUserBinding.Create(ctx)
-		e2eframework.Logf("Waiting 5 minutes for Profiler IAM policy propagation to prevent 403 flakiness...")
-		time.Sleep(5 * time.Minute)
+		e2eframework.Logf("Waiting for Profiler IAM policy propagation to prevent 403 flakiness...")
```
amacaskill, line 172: Why. do we need this?

cmd/sidecar_mounter/main.go
```diff
@@ -84,14 +87,20 @@ func main() {
...
-			}
-			if err := profiler.Start(cfg); err != nil {
-				klog.Errorf("Errored while starting cloud profiler, got %v", err)
-			} else {
-				klog.Infof("Running cloud profiler on %s", cfg.Service)
-			}
+			// Use sync.Once to ensure Cloud Profiler only start once per process.
+			startProfilerOnce.Do(func() {
+				// In hostNetwork mode, hostname is the node name, otherwise it is the pod name.
+				// Appending a 4-character suffix ensures uniqueness for Cloud Profiler.
```
amacaskill: Please remove this comment. We aren't appending 4 char suffix anymore

amacaskill: Same with // In hostNetwork mode, hostname is the node name, otherwise it is the pod name.. I don't think we are using hostname anymore correct?

cmd/sidecar_mounter/main.go
```diff
@@ -84,14 +87,20 @@ func main() {
 		time.Sleep(1500 * time.Millisecond)
 		mc := sidecarmounter.NewMountConfig(sp, flagsFromDriver)
 		if mc.EnableCloudProfilerForSidecar {
```
amacaskill: mc can be nil. Please handle nil condition

test/e2e/utils/handler.go
```diff
@@ -394,6 +394,9 @@ func generateTestSkip(testParams *TestParameters) string {
 	if testParams.UseGKEManagedDriver {
 		skipTests = append(skipTests, "should.not.pass.profile") // Skipping for managed as changes have not been picked up yet
 
+		// TODO(yaozile): remove it after b/500830509
```
amacaskill: Please remove internal bug link. Just say remove it once it is released or something.

test/e2e/testsuites/cloud_profiler.go
```diff
@@ -0,0 +1,189 @@
...
+		expectedVersion := fmt.Sprintf("%s_%s", livePod.Name, livePod.UID)
+
+		ginkgo.By("Checking that the sidecar logs the correct profiler version string")
+		expectedLogLine := fmt.Sprintf("Running cloud profiler on %v with version %s", sidecarServiceName, expectedVersion)
+
+		// Wait for the log to appear
+		tPod.WaitForLog(ctx, sidecarServiceName, expectedLogLine)
+
+		ginkgo.By("Generating load")
+		tPod.VerifyExecInPodSucceed(f, specs.TesterContainerName, fmt.Sprintf("head -c 10485760 </dev/urandom > %s/test.bin", mountPath))
```
amacaskill, line 129: Please explain what this does in a short comment.

test/e2e/testsuites/cloud_profiler.go
```diff
@@ -0,0 +1,189 @@
...
+		tPod.Create(ctx)
+		defer tPod.Cleanup(ctx)
+
+		ginkgo.By("Checking that the pod is running")
+		tPod.WaitForRunning(ctx)
+
+		ginkgo.By("Fetching pod UID to build expected profiler version string")
+		livePod, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Get(ctx, tPod.GetPodName(), metav1.GetOptions{})
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		expectedVersion := fmt.Sprintf("%s_%s", livePod.Name, livePod.UID)
```
amacaskill: Why not import util.GetCloudProfilerServiceVersion, and use that instead?

test/e2e/testsuites/cloud_profiler.go
amacaskill, line 1: IS there a way you can add a test for the node driver profiler? If not, please explain why

test/e2e/testsuites/cloud_profiler.go
```diff
@@ -0,0 +1,189 @@
...
+			g.Expect(gcsfuseOk).To(gomega.BeTrue(), fmt.Sprintf("gcsfuse profile does not exist yet for version %s", expectedVersion))
+		}, "10m", "10s").Should(gomega.Succeed())
+	}
+
+	ginkgo.It("cloud_profiler should create profiles for sidecar and gcsfuse", ginkgo.SpecPriority(10), func() {
+		testCaseCloudProfilerProfileCreation("")
+	})
+}
+
+func checkIfProfileExistForServiceAndVersion(ctx context.Context, serviceName, version string) (bool, error) {
```
amacaskill, line 155: Is this what the GCSFuse Integration profiles test does?

test/e2e/specs/testdriver.go
yaozile123: IAM policy changes take time to propagate to the data plane. If we start the test immediately after creating the role binding, the Cloud Profiler agent might hit 403 Permission Denied errors because the permissions haven't fully taken effect yet, which leads to flaky tests.

test/e2e/testsuites/cloud_profiler.go
yaozile123: Yes, I was mimic [what they were doing](https://github.com/GoogleCloudPlatform/gcsfuse/blob/8180eb250653673c2ed6cfadd3988da3754c478d/tools/integration_tests/cloud_profiler/with_gcp_profiler_service_test.go#L48-L90) for our sidecar and gcsfuse process.

test/e2e/testsuites/cloud_profiler.go
yaozile123: We could technically add a test for it, but the challenge is that the node driver runs continuously as a DaemonSet and is always generating profiler data. And we can't cleanly isolate or recreate a fresh Cloud Profiler state from scratch just for the test.

test/e2e/utils/handler.go
amacaskill: You will still need to add some version skip here once you release it on maanaged driver, to prevent the tests from running when GKE version is less than X.

test/e2e/testsuites/cloud_profiler.go
yaozile123: Done

cmd/csi_driver/main.go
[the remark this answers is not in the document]
yaozile123: Done

pkg/csi_driver/node.go

cmd/sidecar_mounter/main.go
yaozile123: Done

cmd/sidecar_mounter/main.go
yaozile123: Done

test/e2e/testsuites/cloud_profiler.go
yaozile123: Done

pkg/csi_driver/node.go
yaozile123: Sure, do you know which min version should I use? Will it be something like v1.23.1-gke.0?

test/e2e/utils/handler.go
yaozile123: Add the version skip for cloud profiler test unless the gke version is at least 1.36

pkg/csi_driver/node.go
amacaskill: The min version should be v999.999 or something, then after you cut the release tag that contains this PR, that release becomes the version you put here, and then you cut another release tag and that's the one we release.

test/e2e/utils/handler.go
amacaskill: 1.36 won't be the correct version since we already have a 1.36.0 release out there. I would make it something like 1.36.X where its the number it hasn't reached yet, Maybe make it like 1.36.9 or something, then once you cut the release, check the version the non managed testgrid is using, and update it here. Please add a TODO for this.

test/e2e/utils/cluster.go
```diff
@@ -40,6 +40,7 @@ var (
 	supportsMachineTypeAutoConfigMinimumVersion = version.MustParseGeneric("1.33.0")
 	sidecarBucketAccessCheckMinimumVersion      = version.MustParseGeneric("1.34.1")
 	gcsfuseProfilesMinimumVersion               = version.MustParseGeneric("1.35.1")
+	cloudProfilerMinimumVersion                 = version.MustParseGeneric("1.36.0")
```
amacaskill: 1.36.0 already exists so needs to be something higher then you can change it back lower once we know the version

test/e2e/testsuites/gcsfuse_integration.go
```diff
@@ -85,6 +85,7 @@ var testPackageTimeoutMap = map[string]int{
 	testNameWriteLargeFiles: 60,
 	testNameReadLargeFiles:  60,
 	testNameRapidAppends:    60,
+	testNameCloudProfiler:   30,
```
amacaskill, line 88: Why does this need so long to complete?

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+
```
amacaskill: We can remove this line

amacaskill:
Were you able to test that these new tests succeed on a project within our test project pool? To ensure everything is set up properly so when we merge, it also passes on testgrid? If so, then I think once my other comments are addressed, you can merge. After we see green testgrid results, we can cherrypick back to release-1.23, then update the min version in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1307#discussion_r3118699968 to be this tag, and cut a new tag and release that new tag.

amacaskill (approved):

test/e2e/testsuites/gcsfuse_integration.go
yaozile123: I encountered situations where projects had a large amount of profiler data caused the integration test to exceed the 10-minute threshold. This occurs because the Cloud Profiler API does not currently offer a direct way to check for the existence of a specific service name and version, necessitating a one-by-one check of the data which can be time-consuming.

test/e2e/utils/handler.go
yaozile123: Done

test/e2e/utils/cluster.go
yaozile123: Bumped to 1.36.9

pkg/sidecar_mounter/sidecar_mounter_config.go

PR #1341, opened 2026-05-01
Document Cloud Profiler version requirements in releases.md

yaozile123:
**What type of PR is this?**
/kind documentation

Document the minimum GKE and GCS FUSE version requirements for the Cloud Profiler integration in docs/releases.md, linking to the public documentation page.

Fixes #

```release-note
NONE

### CODE REVIEW

File: docs/cloud-profiler.md [hunk 1]
```diff
@@ -0,0 +1,94 @@
+# Cloud Profiler Integration for GCSFuse CSI Driver
```
amacaskill: I left comments on your CL. Please apply the same changes here, after I will do a final approval . Thank you for making this change!!

amacaskill: I actually don't really think we need this doc update in both places. If its already on public docs and is the exact same, we probably don't need it in both places. Some times we have public doc updates for the how to, but troubleshooting in OSS, but since you put both in public docs you can probably remove all these docs, and just update the version map here with the min supported version: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/docs/releases.md#google-cloud-storage-fuse-csi-driver-release-notes

yaozile123: I agree, let me update the PR

amacaskill: Cool, please ping me once its ready for another review

amacaskill (approved):

PR #1342, opened 2026-05-05
Add skipped WIF node-identity fallback security test

su-sudhir:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it:**

Adds a skipped e2e test case to the `workload-identity-federation` test suite to validate a security vulnerability where gcsfuse silently falls back to the node's compute service account when WIF credentials are absent at pod creation time:

- **Node SA has access, pod has no WIF credentials**: When the node compute SA has bucket access but the pod has no WIF credential ConfigMap annotation, gcsfuse should deny the mount instead of silently inheriting the node's GCS permissions. The test is skipped until the node-identity fallback security bug is fixed.

**Which issue(s) this PR fixes:**

N/A

**Does this PR introduce a user-facing change?:**

```release-note
NONE

### DIFF
--- a/test/e2e/testsuites/gcsfuse_oidc_auth.go
+++ b/test/e2e/testsuites/gcsfuse_oidc_auth.go
@@ -37,6 +37,7 @@ import (
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	e2evolume "k8s.io/kubernetes/test/e2e/framework/volume"
 	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
 	admissionapi "k8s.io/pod-security-admission/api"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/e2e/testsuites/gcsfuse_oidc_auth.go [hunk 1]
```diff
@@ -385,6 +386,53 @@ func (t *gcsFuseCSIOIDCTestSuite) DefineTests(driver storageframework.TestDriver
 	ginkgo.It("should fail when CSI bucket access check is enabled with OIDC authentication", func() {
 		testCaseOIDCWithCSIBucketAccessCheck()
 	})
+
+	// Security scenario: node SA has bucket access but the pod has no WIF credentials.
+	// The admission webhook must reject pod creation when the WIF credential ConfigMap
+	// annotation is empty, preventing any fallback to the node's identity.
+	//
+	// TODO: Remove the skip below once the node-identity-fallback security bug is fixed.
+	ginkgo.It("should fail GCS mount when WIF credential config is absent and node SA has bucket access", func() {
```
mattcary: Change the description to something like "it should fail to start pod when..." as that's more precise w/rt the behavior of the test.

su-sudhir: Renamed the test-title to  `should fail to start pod when WIF credential config is absent and node SA has bucket access`

mattcary:
Waiting on manual e2e test run.

su-sudhir:
> Waiting on manual e2e test run.

Since this is a Skipped Test case, the manual run logs are possible once the Security leak issue is fixed.

mattcary (approved):

mattcary:
> > Waiting on manual e2e test run.
> 
> Since this is a Skipped Test case, the manual run logs are possible once the Security leak issue is fixed.

ok, sgtm

PR #1352, opened 2026-05-12
test(e2e): support GCSFuse v3.10+ cloud profiler service name optimization and skip on older versions

PranjalC100:
**What type of PR is this?**
/kind failing-test

The `cloud_profiler` test running on GKE currently experiences intermittent `context deadline exceeded` errors. This happens because the test project accumulates a massive amount of profile data under the default `gcsfuse` service name, causing the profile lookup pagination to time out while iterating through unrelated profiles.

An optimization was introduced in GCSFuse v3.9.0 to use a unique, lexicographically decreasing custom service name for each test run. This allows the API pagination to break early since the results are sorted alphabetically. 

This PR updates the GKE integration test framework to support this optimization by:
1. Generating a lexicographically decreasing string using base36 encoding.
2. Exporting `PROFILE_SERVICE_NAME` and `PROFILE_LABEL` variables into the test pod environment, which are picked up by `test_config.yaml` to dynamically set `--cloud-profiler-service-name` and `--cloud-profiler-label`.
3. Skipping the `cloud_profiler` test for GCSFuse versions older than `v3.9.0-gke.0` (such as `v3.8`), as this optimization involves core code changes not backported to older versions and the v3.9 test grid is already stable.

Fixes b/512210213

This brings the GKE integration tests in sync with the GCSFuse integration tests changes made in GoogleCloudPlatform/gcsfuse#4538.

```release-note
NONE
```
Test gPaste output: https://paste.googleplex.com/6206146571534336

### DIFF
--- a/test/e2e/testsuites/gcsfuse_integration.go
+++ b/test/e2e/testsuites/gcsfuse_integration.go
@@ -243,10 +243,10 @@ func runIntegrationTest(ctx context.Context, f *framework.Framework, driver stor
 	if opts.TestPkg == testNameCloudProfiler || opts.TestName == testNameCloudProfiler {
 		prefix := getDecreasingString()
 		profileLabel = fmt.Sprintf("%s-cloud-profiler-test", prefix)
-		// serviceName and profileLabel can be identical or different. They must follow 
-		// a decreasing lexicographical order per run to avoid excessive pagination 
+		// serviceName and profileLabel can be identical or different. They must follow
+		// a decreasing lexicographical order per run to avoid excessive pagination
 		// and potential timeouts when retrieving profiles from the GCP project.
-		serviceName = profileLabel 
+		serviceName = profileLabel
 
 		vars["PROFILE_LABEL"] = profileLabel
 		vars["PROFILE_SERVICE_NAME"] = serviceName
--- a/test/e2e/testsuites/gcsfuse_integration.go
+++ b/test/e2e/testsuites/gcsfuse_integration.go
@@ -243,7 +243,10 @@ func runIntegrationTest(ctx context.Context, f *framework.Framework, driver stor
 	if opts.TestPkg == testNameCloudProfiler || opts.TestName == testNameCloudProfiler {
 		prefix := getDecreasingString()
 		profileLabel = fmt.Sprintf("%s-cloud-profiler-test", prefix)
-		serviceName = profileLabel
+		// serviceName and profileLabel can be identical or different. They must follow 
+		// a decreasing lexicographical order per run to avoid excessive pagination 
+		// and potential timeouts when retrieving profiles from the GCP project.
+		serviceName = profileLabel 
 
 		vars["PROFILE_LABEL"] = profileLabel
 		vars["PROFILE_SERVICE_NAME"] = serviceName
--- a/test/e2e/testsuites/gcsfuse_integration.go
+++ b/test/e2e/testsuites/gcsfuse_integration.go
@@ -20,12 +20,13 @@ package testsuites
 import (
 	"context"
 	"fmt"
+	"math"
 	"strings"
+	"time"
 
 	"local/test/e2e/specs"
 	"local/test/e2e/utils"
 
-	"github.com/google/uuid"
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -164,6 +165,7 @@ type TestCommandConfig struct {
 	BucketName         string
 	OnlyDir            string
 	ProfileLabel       string
+	ProfileServiceName string
 	ReadOnly           bool
 	BillingProject     string
 }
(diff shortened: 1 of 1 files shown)


test/e2e/testsuites/gcsfuse_integration.go
```diff
@@ -343,6 +369,10 @@ func generateTestCommand(opts TestCommandConfig) string {
 		commandArgs = append(commandArgs, fmt.Sprintf("export PROFILE_LABEL=%s", opts.ProfileLabel))
 	}
 
+	if opts.ProfileServiceName != "" {
```
meet2mky, line 375: Hi @yaozile123, Is this code executed even for the v3.9.0 GCSFuse version testing. If yes then could you tell what version of test_config.yaml will be used if this is going to be used from v3.9.0 or current master. Because if it's going to be used from  the v3.9.0 release branch then the handling logic for this PROFILE_SERVICE_NAME env variable is not present in GCSFuse.

test/e2e/testsuites/gcsfuse_integration.go
```diff
@@ -516,6 +546,11 @@ func (t *gcsFuseCSIGCSFuseIntegrationTestSuite) DefineTests(driver storageframew
 			e2eskipper.Skipf("skip gcsfuse integration test %v on gcsfuse version %v", testNameRequesterPaysBucket, v.String())
 		}
 
+		// GCSFuse cloud_profiler tests are supported after v3.9.0.
+		if !v.AtLeast(version.MustParseSemantic("v3.9.0-gke.0")) && testName == testNameCloudProfiler {
```
meet2mky: It should be v3.10.0 if the test_config.yaml is taken from the same branch as per https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1352/changes#r3228100288 which I think should be the case.

@PranjalC100

PranjalC100: Sure I'll change it according to how `test_config.yaml` is picked up from our repo.

test/e2e/testsuites/gcsfuse_integration.go
yaozile123: Why we need export PROFILE_SERVICE_NAME? This test code on managed driver 1.36 will be executed on GCSFuse release branch v3.9.0. Should we disable the test in v3.9.0 as well?

yaozile123: I saw we are [extracting service version](https://github.com/GoogleCloudPlatform/gcsfuse/blob/79c1cbabec2e9ed331d9082938bc2179dc6831ba/tools/integration_tests/cloud_profiler/cloud_profiler_test.go#L118) in v3.9.0, will this behavior changed in v3.10? If so, please disable the test in v3.9 as well.

PranjalC100: Support for the `cloud_profiler` package can be applied as a patch to the v3.9.0 branch. Since the modifications only affect `test_config.yaml` and packages, there is no impact on core gcsfuse code. The changes were picked up in v3.9.0 (commit [4b7892b](https://github.com/GoogleCloudPlatform/gcsfuse/commit/4b7892bca5cb2fea0034056f253ab5b911c9e2e9)), we will also apply a [patch fix ](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4526)to the v3.9.0 branch.

test/e2e/testsuites/gcsfuse_integration.go
PranjalC100: I'll patch fix the v3.9.0 release branch with this [commit](https://github.com/GoogleCloudPlatform/gcsfuse/pull/4526) after merging it.

test/e2e/testsuites/gcsfuse_integration.go
yaozile123: Please explain why do we need export PROFILE_SERVICE_NAME but not for PROFILE_LABEL

test/e2e/testsuites/gcsfuse_integration.go
yaozile123: @meet2mky Should we disable the test until v3.10? We need to see the test passed on the testgrid against master branch first.

test/e2e/testsuites/gcsfuse_integration.go
```diff
@@ -218,10 +239,14 @@ func runIntegrationTest(ctx context.Context, f *framework.Framework, driver stor
 
 	vars := map[string]string{"BUCKET_NAME": bucketName}
 
-	var profileLabel string
+	var profileLabel, serviceName string
 	if opts.TestPkg == testNameCloudProfiler || opts.TestName == testNameCloudProfiler {
-		profileLabel = fmt.Sprintf("ve2e0.0.0-%s", strings.ReplaceAll(uuid.New().String(), "-", "")[:8])
+		prefix := getDecreasingString()
+		profileLabel = fmt.Sprintf("%s-cloud-profiler-test", prefix)
+		serviceName = profileLabel // kept identical
```
yaozile123: why we need service name and profile label be the same?  can you explain it here

test/e2e/testsuites/gcsfuse_integration.go
meet2mky: Yes let's enable it on v3.10 and newer versions.

test/e2e/testsuites/gcsfuse_integration.go
PranjalC100: The `PROFILE_LABEL` export was already present(just 5 lines above this) in the integration test framework. This PR adds the export for `PROFILE_SERVICE_NAME` to support the new -**-cloud-profiler-service-name** flag introduced in GCSFuse v3.9.0+

test/e2e/testsuites/gcsfuse_integration.go
PranjalC100: They can be identical or different. Both are fine. They just need to be in decreasing lexicographical order within the GCP Project everytime the test runs.

yaozile123: Can you add this explanation to the comment instead of saying keep identical?

yaozile123:
Can you change the title of your PR and squash your commit? Then I will approve it

test/e2e/testsuites/gcsfuse_integration.go
PranjalC100: Added!

PranjalC100:
> Can you change the title of your PR and squash your commit? Then I will approve it

I'll squash and merge in the after the approval! Works the same way:)

yaozile123 (approved):

yaozile123:
LGTM, please squash the commit before you merge. Thanks!

PR #1376, opened 2026-05-19
Allow users to explicitly disable GCSFuse Cloud Profiler via mountOpt…

yaozile123:
…ions

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

- **Support GCSFuse Profiler Opt-Out::** This PR allows users to explicitly disable GCSFuse profiling (via `enable-cloud-profiler=false` in mount options) when sidecar profiling is enabled (`enableCloudProfilerForSidecar: "true"`).
- **Bumped minimum sidecar version:** Raised `SidecarCloudProfilerMinVersion` to `v1.23.7-gke.0` in `pkg/csi_driver/utils.go`.
- **Added mounter unit tests:** Added unit test case `should not enable cloud profiler when enable-cloud-profiler=false is explicitly passed` in `sidecar_mounter_config_test.go`.
- **Added E2E integration tests:** Introduced the `cloud_profiler should only create profiles for sidecar when gcsfuse profiler is disabled` E2E test in `cloud_profiler.go`.

Fixes #

Tested the cloud profiler e2e test
```
```bash
export PROJECT_ID=yaozile-gke-dev
export REGISTRY=us-central1-docker.pkg.dev/$PROJECT_ID/gcs-fuse-csi-driver
export STAGINGVERSION=prow-gob-internal-boskos-200

make build-image-and-push-multi-arch REGISTRY=$REGISTRY STAGINGVERSION=$STAGINGVERSION

REGISTRY=$REGISTRY \
E2E_TEST_FOCUS="cloudProfiler" \
make e2e-test

  Ran 2 of 587 Specs in 3623.292 seconds
  SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 585 Skipped


  Ginkgo ran 1 suite in 1h0m31.151604346s
  Test Suite Passed
```

User can choose to opt out profiling on GCSFuse instance by pass `enable-cloud-profiler=false` in mount options when sidecar profiling is enabled (`enableCloudProfilerForSidecar: "true"`)
```release-note

```


pkg/csi_driver/utils.go
```diff
@@ -77,7 +77,7 @@ const (
 	VolumeContextKeyBucketName             = "bucketName"
 	TokenServerSidecarMinVersion           = "v1.17.2-gke.0" // #nosec G101
 	SidecarBucketAccessCheckMinVersion     = "v1.20.0-gke.0"
-	SidecarCloudProfilerMinVersion         = "v1.23.5-gke.0"
+	SidecarCloudProfilerMinVersion         = "v1.23.7-gke.0"
```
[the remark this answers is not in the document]
yaozile123: This is intentional since we need to support disable the cloud profiler for GCSFuse vs sidecar mounter separately

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -356,11 +356,15 @@ func (mc *MountConfig) prepareMountArgs() {
 	}
 
 	if mc.EnableCloudProfilerForSidecar {
-		// Inject Cloud Profiler flags supported by gcsfuse.
-		flagMap["enable-cloud-profiler"] = ""
-		customLabel := util.GetCloudProfilerServiceVersion(mc.PodName, mc.PodUID)
-		flagMap["cloud-profiler-label"] = customLabel
-		klog.Infof("Setting label in GCSFuse mount options: %q", customLabel)
+		if val, ok := flagMap["enable-cloud-profiler"]; !ok || val != "false" {
```
amacaskill: Pretty sure more than just "false" is valid value for false. Use ParseBool

pkg/csi_driver/utils.go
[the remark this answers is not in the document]
amacaskill: We already rolled out a component that enables the feature by default for node driver. Are we sure we can just bump this with no consequences? if so please explain why in this comment.

pkg/sidecar_mounter/sidecar_mounter_config.go

pkg/csi_driver/utils.go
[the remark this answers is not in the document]
yaozile123: IIUC, the managed node driver will automatically updated once we rolled out the new component.  For already injected sidecars, the profiler is already spun up and will continue to run.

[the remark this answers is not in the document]
yaozile123: We need to bump it to `v1.23.7-gke.0` otherwise user won't able to disable the profiler in GCSFuse instance if they were using sidecar version `v1.23.5-gke.0` and  `v1.23.6-gke.0`

test/e2e/testsuites/cloud_profiler.go
```diff
@@ -124,6 +127,12 @@ func (t *gcsFuseCSICloudProfilerTestSuite) DefineTests(driver storageframework.T
 		// Wait for the log to appear
 		tPod.WaitForLog(ctx, sidecarServiceName, expectedLogLine)
 
+		if disableGcsfuseProfiler {
```
amacaskill, line 130: Any test skip needed for this test for the managed testgrid since older versions of managed driver don't have your fix?

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -356,11 +356,23 @@ func (mc *MountConfig) prepareMountArgs() {
 	}
 
 	if mc.EnableCloudProfilerForSidecar {
-		// Inject Cloud Profiler flags supported by gcsfuse.
-		flagMap["enable-cloud-profiler"] = ""
-		customLabel := util.GetCloudProfilerServiceVersion(mc.PodName, mc.PodUID)
-		flagMap["cloud-profiler-label"] = customLabel
-		klog.Infof("Setting label in GCSFuse mount options: %q", customLabel)
+		enableProfiler := true
+		val, ok := flagMap["enable-cloud-profiler"]
```
amacaskill: Can we use a constant for "enable-cloud-profiler"?

amacaskill (approved):

test/e2e/testsuites/cloud_profiler.go
yaozile123: The `cloud_profiler` test will only run after [GKE version `v1.36.1`](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/b664a41bed9a8602bbad56b503fbc65041a06683/test/e2e/utils/cluster.go#L44) so I think at that time the sidecar version should be greater than `1.23.7-gke.0` which will have the fix.

pkg/sidecar_mounter/sidecar_mounter_config.go

PR #1378, opened 2026-05-19
Override storage endpoint with options.go endpoint

uriel-guzman:
**What type of PR is this?**
kind feature

Currently, customers need to specify `custom-endpoint` to use the GCSFuse CSI Driver in non-GDU universes. This PR makes the detection of the storage endpoint automatic, by passing an internal flag 'storage-endpoint-internal' to the sidecar, so that it also creates the clients in the correct universe.

Fixes #

```release-note
Automatically determine storage endpoint.
```

### DIFF
--- a/deploy/base/node/node.yaml
+++ b/deploy/base/node/node.yaml
@@ -64,6 +64,7 @@ spec:
             - --enable-cloud-profiler-for-driver=true
             - --enable-auto-gomemlimit=true
             - --auto-gomemlimit-ratio=0.95
+            - --universe-domain=$(UNIVERSE_DOMAIN)
           ports:
             - containerPort: 9920
               name: metrics
@@ -91,6 +92,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.uid
+            - name: UNIVERSE_DOMAIN
+              valueFrom:
+                configMapKeyRef:
+                  name: gcsfusecsi-node-config
+                  key: universe-domain
           volumeMounts:
             - name: kubelet-dir
               mountPath: /var/lib/kubelet/pods
@@ -172,3 +178,10 @@ metadata:
 data:
   sidecar-image: gke.gcr.io/gcs-fuse-csi-driver-sidecar-mounter
   metadata-sidecar-image: gke.gcr.io/gcs-fuse-csi-driver-metadata-prefetch
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gcsfusecsi-node-config
+data:
+  universe-domain: googleapis.com
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -164,9 +164,10 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Check if the given Service Account has the access to the GCS bucket, and the bucket exists.
 	// skip check if it has ever succeeded
+	storageEndpoint := storageEndpointFromUniverseDomain(s.driver.config.UniverseDomain)
 	if vs != nil && !args.skipCSIBucketAccessCheck {
 		if !vs.BucketAccessCheckPassed {
-			storageService, err := s.prepareStorageService(ctx, vc, args.customEndpoint)
+			storageService, err := s.prepareStorageService(ctx, vc, storageEndpoint)
 			if err != nil {
 				return nil, status.Errorf(codes.Unauthenticated, "failed to prepare storage service: %v", err)
 			}
@@ -383,6 +384,12 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	disallowedFlags := s.driver.generateDisallowedFlagsMap(gcsFuseSidecarImage)
 	args.fuseMountOptions = removeDisallowedMountOptions(args.fuseMountOptions, disallowedFlags)
+
+	// Pass the universe-aware storage endpoint if the sidecar version supports it.
+	if s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, StorageEndpointInternalMinVersion) {
+		args.fuseMountOptions = overrideStorageEndpointInternal(args.fuseMountOptions, storageEndpoint)
+	}
+
 	// Start to mount
 	if err = s.mounter.Mount(args.bucketName, targetPath, FuseMountType, args.fuseMountOptions); err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to mount volume %q to target path %q: %v", args.bucketName, targetPath, err)
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -90,6 +90,8 @@ const (
 	FlagFileForDefaultingPath              = "flags-for-defaulting"
 	GCSFuseProfileFlag                     = "profile"
 	LocalSocketAddressArg                  = "experimental-local-socket-address"
+	// TODO(urielguzman): Change this to the actual version where it's supported.
+	StorageEndpointInternalMinVersion = "v1.999.0-gke.0"
 )
 
 var (
@@ -118,7 +120,6 @@ type requestArgs struct {
 	optInHostnetworkKSA           bool
 	enableCloudProfilerForSidecar bool
 	multiNICIndex                 int
-	customEndpoint                string
 }
 
 func NewControllerServiceCapability(c csi.ControllerServiceCapability_RPC_Type) *csi.ControllerServiceCapability {
@@ -374,10 +375,40 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest, vc map[string]stri
 
 	args, err := parseVolumeAttributes(fuseMountOptions, vc)
 	args.bucketName = bucketName
-	if customEndpointVal := util.CustomEndpointFromOpts(args.fuseMountOptions); customEndpointVal != "" {
-		args.customEndpoint = customEndpointVal
-	}
 	return args, err
+}
+
+// storageEndpointFromUniverseDomain returns the storage endpoint for the universe domain,
+// in the format storage.UNIVERSE_DOMAIN:443. If no universe domain is provided, googleapis.com
+// will be used as default.
+func storageEndpointFromUniverseDomain(universeDomain string) string {
+	if universeDomain == "" {
+		universeDomain = "googleapis.com"
+	}
+	return fmt.Sprintf("storage.%s:443", universeDomain)
+}
+
+// overrideStorageEndpointInternal adds the storage-endpoint-internal flag to the mountOptions.
+// If it already exists, it will be overridden.
+func overrideStorageEndpointInternal(opts []string, storageEndpoint string) []string {
+	prefix := fmt.Sprintf("%s=", util.StorageEndpointInternal)
+	replacement := fmt.Sprintf("%s%s", prefix, storageEndpoint)
+	newMountOptions := make([]string, 0, len(opts)+1)
+
+	for _, opt := range opts {
+		if strings.HasPrefix(opt, prefix) {
+			if opt != replacement {
+				// Log a warning if we are actively changing an existing value.
+				klog.Warningf("Found flag %q, replaced with %q", opt, replacement)
+			}
+			// Skip adding the old/incorrect flag to the new slice.
+			continue
+		}
+		newMountOptions = append(newMountOptions, opt)
+	}
+
+	// Always append the correct/updated flag at the end.
+	return append(newMountOptions, replacement)
 }
 
 func putExitFile(pod *corev1.Pod, targetPath string) error {
--- a/pkg/sidecar_mounter/sidecar_mounter_config.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_config.go
@@ -72,6 +72,7 @@ type MountConfig struct {
 	CustomEndpoint                 string                `json:"-"`
 	EnableAutoGoMemLimit           bool                  `json:"-"`
 	AutoGoMemLimitRatio            float64               `json:"-"`
+	StorageEndpoint                string                `json:"-"`
 }
 
 // sidecarRetryConfig controls the retry configurations for sidecarRetry behivior for storage service creation and bucket access check.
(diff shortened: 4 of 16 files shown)


pkg/csi_driver/utils.go
```diff
@@ -370,12 +369,56 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest, vc map[string]stri
...
+}
+
+// storageEndpointFromUniverseDomain returns the storage endpoint for the universe domain,
+// in the format storage.UNIVERSE_DOMAIN:443. If no universe domain is provided, googleapis.com
+// will be used as default.
+func storageEndpointFromUniverseDomain(universeDomain string) string {
+	if universeDomain == "" {
+		universeDomain = "googleapis.com"
+	}
+	return fmt.Sprintf("storage.%s:443", universeDomain)
```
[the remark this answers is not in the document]
uriel-guzman: we tell customers to use the other format: https://docs.cloud.google.com/storage/docs/cloud-storage-fuse/config-file#custom-endpoint

deploy/base/node/node.yaml
```diff
@@ -89,6 +90,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.uid
+            - name: UNIVERSE_DOMAIN
+              valueFrom:
+                configMapKeyRef:
+                  name: gcsfusecsi-node-config
```
amacaskill, line 98: Why do we a config map for this?

uriel-guzman: This pattern already exists two other times: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/Makefile#L308-L311, is there a better way?

pkg/csi_driver/utils.go
```diff
@@ -370,12 +369,56 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest, vc map[string]stri
...
+}
+
+// overrideCustomEndpointWithStorageEndpoint replaces any existing custom endpoint
+// configurations—whether specified via CLI flags or config file prefixes—with a unified
+// CLI flag pointing to the provided universe storage endpoint.
+func overrideCustomEndpointWithStorageEndpoint(opts []string, storageEndpoint string) []string {
+	// If there's already a user-specified custom endpoint that doesn't match the storage endpoint,
+	// log a warning, since it's likely misconfigured and will fail later anyway if we don't replace
+	// it.
+	if val := util.CustomEndpointFromOpts(opts); val != "" && val != storageEndpoint {
```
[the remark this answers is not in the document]
amacaskill: normalize the endpoint strings (e.g., by stripping schemes and trailing slashes) before comparing them . I don't think we need a warning generated in this case correct?

		klog.Warningf("Found custom endpoint with value: https://storage.googleapis.com, overriding with storage endpoint: [%q](storage.googleapis.com:443)"

pkg/csi_driver/utils.go
```diff
@@ -85,10 +85,12 @@ const (
 	GCSFuseProfilesMinVersion              = "v1.19.3-gke.0"
 	GCSFuseFileCacheMediumMinVersion       = "v1.21.0-gke.0"
 	GCSFuseKernelParamsMinVersion          = "v1.22.0-gke.0"
-	MultiNICMinVersion                     = "v1.22.2-gke.0"
-	FlagFileForDefaultingPath              = "flags-for-defaulting"
-	GCSFuseProfileFlag                     = "profile"
-	LocalSocketAddressArg                  = "experimental-local-socket-address"
+	// TODO(urielguzman): Change this to the actual version where it's supported.
+	StorageEndpointInternalMinVersion = "v1.999.0-gke.0"
```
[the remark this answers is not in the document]
uriel-guzman: This is fine, we don't want to enable it until we know the version.

pkg/csi_driver/node.go
```diff
@@ -164,9 +164,10 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Check if the given Service Account has the access to the GCS bucket, and the bucket exists.
 	// skip check if it has ever succeeded
+	storageEndpoint := storageEndpointFromUniverseDomain(s.driver.config.UniverseDomain)
 	if vs != nil && !args.skipCSIBucketAccessCheck {
 		if !vs.BucketAccessCheckPassed {
-			storageService, err := s.prepareStorageService(ctx, vc, args.customEndpoint)
+			storageService, err := s.prepareStorageService(ctx, vc, storageEndpoint)
```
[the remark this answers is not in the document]
uriel-guzman: The custom-endpoint was never intended to be the SoT for sidecat bucket access check. Customers who are using custom-endpoint aren't affected because it was originally intended as a GCSFuse mount option.

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -315,6 +311,10 @@ func (mc *MountConfig) prepareMountArgs() {
 		case util.EnableGCSFuseKernelParams:
 			mc.EnableGCSFuseKernelParams = value == util.TrueStr
 			continue
+
+		case util.StorageEndpointInternal:
+			mc.StorageEndpoint = value
+			continue
```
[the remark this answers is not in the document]
uriel-guzman: That's fine. If the customer who was using custom-endpoint in non-GDU upgrades, and the CSI Driver is newer, and the sidecar is older, `custom-endpoint` will be used to setup the sidecar clients, so they don't break. We should not continue using custom-endpoint, because a malicious agent can inject arbitrary endpoints, so we want to move away from this.

deploy/base/node/node.yaml
amacaskill, line 1: I know profiles doesn't support TPC at the moment, but can we just add this to controller.yaml anyways? SO we don't forget to add it in controller.yaml later and make TPC test fail when we test profiles TPC support?

deploy/base/node/node.yaml
amacaskill: Discussed offline. We do this Because you want it to be dynamic based on cluster config Rather than hardcoding the universe flag in the deployment yaml before deploying

pkg/csi_driver/utils.go
```diff
@@ -373,12 +374,46 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest, vc map[string]stri
...
+	for _, opt := range opts {
+		if strings.HasPrefix(opt, prefix) {
+			replaced = append(replaced, opt)
+			continue
+		}
+		newMountOptions = append(newMountOptions, opt)
+	}
+	// Add the new replacement option at the end.
+	if len(replaced) > 0 {
+		klog.Warningf("Found flags %v, replaced with %q", replaced, replacement)
```
amacaskill: Doesn't this fire if any matching prefix option is found in the slice, even if the value is completely identical to the driver's calculated replacement?

mountOptions:
  - o=noexec
  - uid=1000
  - gid=1000
 

overrideMountOption(opts, "uid=", "uid=1000")
Store "uid=1000" in the replaced tracker slice.
W0522] Found flags [uid=1000], replaced with "uid=1000"

amacaskill: If its generic function, it needs to be generic in case someone calls it in the future.

pkg/csi_driver/utils.go
```diff
@@ -370,12 +369,46 @@ func parseRequestArguments(req *csi.NodePublishVolumeRequest, vc map[string]stri
...
-	if customEndpointVal := util.CustomEndpointFromOpts(args.fuseMountOptions); customEndpointVal != "" {
-		args.customEndpoint = customEndpointVal
-	}
 	return args, err
 }
 
+// overrideMountOption filters out any mount options that match the given prefix patterns,
+// appends the replacement option, and returns the updated slice.
+func overrideMountOption(opts []string, prefixPatterns []string, replacement string) []string {
+	var newMountOptions []string
```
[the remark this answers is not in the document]
amacaskill: Why do we not want to do this? 

Worst-Case Allocation (No Prefix Matches): If zero elements match the prefix, the loop appends all N elements into newMountOptions. The final return appends 1 more element (replacement), bringing the final slice size to N+1. So I don't 

If capacity is N: The loop completes without resizing, but the final append(newMountOptions, replacement) exceeds capacity, forcing Go to allocate a new, larger array and copy all N elements over.

If capacity is unallocated (var newMountOptions []string): Go progressively re-allocates and copies elements multiple times as the slice grows dynamically from 0→1→2→4…

pkg/csi_driver/node.go
```diff
@@ -347,6 +348,12 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	disallowedFlags := s.driver.generateDisallowedFlagsMap(gcsFuseSidecarImage)
 	args.fuseMountOptions = removeDisallowedMountOptions(args.fuseMountOptions, disallowedFlags)
+
+	// Pass the universe-aware storage endpoint if the sidecar version supports it.
+	if s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, StorageEndpointInternalMinVersion) {
+		args.fuseMountOptions = overrideStorageEndpointInternal(args.fuseMountOptions, storageEndpoint)
+	}
+
```
amacaskill, line 392: we can add unit test to node_test to make sure 
- The driver does not inject the internal flag if the sidecar image version is older than StorageEndpointInternalMinVersion
- the driver correctly computes the universe-aware endpoint and injects storage-endpoint-internal=<endpoint> when the sidecar version supports it

deploy/base/node/node.yaml
```diff
@@ -170,3 +176,9 @@ metadata:
 data:
   sidecar-image: gke.gcr.io/gcs-fuse-csi-driver-sidecar-mounter
   metadata-sidecar-image: gke.gcr.io/gcs-fuse-csi-driver-metadata-prefetch
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gcsfusecsi-node-config
+data: {}
```
amacaskill: If a user or a deployment pipeline installs the CSI driver using the base manifests directly (without applying the Makefile/kustomize overlays, which is common for self-managed or third-party Kubernetes distros) ([ source](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1378/changes?content_ref=docker_buildx_args+quiet#diff-c7f2e26441c7b79b970577a423bab66c277383cad02fc8fcc0b238efaf0f086f)), the key universe-domain is missing.

Which would cause  CreateContainerConfigError when trying to statr node daemonet . Should we add default vaou

amacaskill (approved):

pkg/csi_driver/utils.go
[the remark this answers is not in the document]
uriel-guzman: Your math on the worst-case allocation behavior is completely correct, and in a high-throughput, hot-path loop handling thousands of elements, pre-allocating would absolutely be the right call.

However, in this specific context, users realistically have maybe a few dozen of mount options. At this scale, the difference between dynamic resizing and pre-allocation is a matter of nanoseconds and a few bytes of memory, which has zero measurable impact on the application's performance or footprint. Given that the current code is clean and readable, keeping it as-is avoids over-engineering for a microscopic performance gain, which could sometimes even be a tiny bit slower due to the overhead ot setting up the capacity on the heap rather than letting small slcies stay on the stack. I'd prefer to leave it as a minor style variance.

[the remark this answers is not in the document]
amacaskill: > users realistically have maybe a few dozen of mount options. At this scale, the difference between dynamic resizing and pre-allocation is a matter of nanoseconds and a few bytes of memory, which has zero measurable impact on the application's performance or footprint.

Sure, maybe when you are looking at a single, isolated execution path. But this is called in the node driver (which there is only 1 pod per node) . Agentic Customers are starting to want to max out pods per node, and pod limit per node is going to be raised for these customers. We have 1 node driver pod with limited memory and we are already seeing OOM issues. Under load, the cumulative effects of dynamic memory allocations lead to significant garbage collection overhead and memory spikes.   Since our node driver runs under strict resource constraints and is already experiencing OOM, so I think let's just make the change. Its just a 1 line change. I don't really think it hurts the style, and I think the garbage collection // memory at scale is more important than the style

deploy/base/node/node.yaml
uriel-guzman: I think it would be best to not make any changes to the controller, if they are not currently required, to avoid any surprises. I left a TODO instead, so we don't forget to set the flag in the future when we do the profiles TPC fix.

Also, the controller.yaml doesn't have the config-map, so adding the flag here won't really do more than a TODO, because the cmd/main.go file already defaults it to googleapis.com. It's also not going to work anyway if we try to add the config-file, because the compute client will fail anyway. This change is for storage client only.

pkg/csi_driver/utils.go
uriel-guzman: I changed the function to be specific to storage-endpoint-internal to simplify. Otherwise, we'd need to handle config-file and CLI key format stuff for that generic function, which I think would be out of scope of the goal of this PR.

pkg/csi_driver/node.go
uriel-guzman: done

deploy/base/node/node.yaml
uriel-guzman: done

pkg/csi_driver/utils.go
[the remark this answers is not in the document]
uriel-guzman: This is an old comment made by Gemini, on a piece of code that doesn't exist anymore.

pkg/csi_driver/utils.go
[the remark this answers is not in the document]
uriel-guzman: Ack, added in the change.

uriel-guzman:
The unit-test is failing, I don't know why. It was also failing on https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1364/changes/4e541340ddf33d90598031a8fc12e9ae208d312b when it got merged: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/actions/runs/26299254579/job/77419884488, seems like it's timing out, I don't think this is related to this PR.

deploy/base/node/node.yaml
amacaskill: Sounds good

PR #1412, opened 2026-06-01
feat(sidecar): respect first appearance of enable-sidecar-bucket-acce…

yangspirit:
…ss-check

**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Currently, GCS FUSE CSI inline ephemeral volumes silently override user-specified overrides of `enable-sidecar-bucket-access-check=false` (or `skipCSIBucketAccessCheck: "true"`). This happens because the CSI driver appends its global default `enable-sidecar-bucket-access-check=true` after the user options. Because the sidecar processed the list sequentially and used the last-seen value, the default `true` always won, blocking non-Google auth flows from mounting GCS buckets via inline volumes.
This PR implements a fast-path sidecar fix to respect the first occurrence of `enable-sidecar-bucket-access-check` (first-wins parsing). Because mount options are sorted alphabetically by the CSI driver before being passed to the sidecar, `"enable-sidecar-bucket-access-check=false"` always sorts before `"true"` and is processed first. Using a state guard, we capture the first-seen value (user's false preference) and ignore subsequent ones.
This immediately unblocks customers using custom sidecar image overrides.

Fixes # b/495760030

This is the short-term sidecar-only fix meant to be released in a new sidecar container image immediately. The robust driver-side deduplication fix will follow as a separate PR for the next standard driver release.

```release-note
Fixed a bug where user-specified `enable-sidecar-bucket-access-check=false` mount option was ignored in GCS FUSE CSI inline ephemeral volumes.
```

### DIFF
--- a/pkg/sidecar_mounter/sidecar_mounter_config.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_config.go
@@ -207,6 +207,7 @@ func (mc *MountConfig) prepareMountArgs() {
 	invalidArgs := []string{}
 
 	mc.GcsFuseNumaNode = -1
+	enableSidecarBucketAccessCheckParsed := false
 	for _, arg := range mc.Options {
 		klog.Infof("processing mount arg %v", arg)
 
@@ -276,7 +277,10 @@ func (mc *MountConfig) prepareMountArgs() {
 			continue
 
 		case util.EnableSidecarBucketAccessCheckConst:
-			mc.EnableSidecarBucketAccessCheck = value == util.TrueStr
+			if !enableSidecarBucketAccessCheckParsed {
+				mc.EnableSidecarBucketAccessCheck = value == util.TrueStr
+				enableSidecarBucketAccessCheckParsed = true
+			}
 			continue
 
 		case util.PodNamespaceConst:
--- a/pkg/csi_driver/utils_test.go
+++ b/pkg/csi_driver/utils_test.go
@@ -178,6 +178,41 @@ func TestJoinMountOptions(t *testing.T) {
 	}
 }
 
+func TestJoinMountOptionsSorting(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name            string
+		existingOptions []string
+		newOptions      []string
+		expectedOptions []string
+	}{
+		{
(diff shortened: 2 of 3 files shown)


pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -276,7 +277,10 @@ func (mc *MountConfig) prepareMountArgs() {
 			continue
 
 		case util.EnableSidecarBucketAccessCheckConst:
```
uriel-guzman, line 279: Can we just add a comment here explaining why we are doing this and mention that it will be removed in a later release once the "long term" solution is released? (in case someone else thinks they can start relying on this)

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -276,7 +277,10 @@ func (mc *MountConfig) prepareMountArgs() {
...
-			mc.EnableSidecarBucketAccessCheck = value == util.TrueStr
+			if !enableSidecarBucketAccessCheckParsed {
```
uriel-guzman, line 280: Can we just manually verify that this works the customer before merging? You can just create a pod with `skipCSIBucketAccessCheck=true` and `mountOptions:enable-sidecar-bucket-access-check=false` and verify the sidecar performs no access checks. To verify that, you can create a pod and skip giving the KSA access to the bucket: https://docs.cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-setup#authentication

If this works properly, then you shouldn't see an error coming from the sidecar mounter: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/sidecar_mounter/sidecar_mounter.go#L103

Instead, you should see the GCSFuse mount attempted: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/sidecar_mounter/sidecar_mounter.go#L107, and then it will fail when GCSFuse mounts, but that proves the sidecar bucket access checks were all skipped.

uriel-guzman (approved):

PR #1413, opened 2026-06-02
Implement ControllerPublishVolume skeleton

uriel-guzman:
**What type of PR is this?**
/kind feature

This PR implements dummy code for ControllerPublishVolume for the upcoming shared node mount feature. It does a simple check to understand if the `sharedMount` has been passed or not from the VolumeContext.

Currently, this change is a no-op, because the driver has not yet implemented "attachRequired: true".

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -127,6 +127,43 @@ func (s *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req *
 	}, nil
 }
 
+func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
+	// Validate arguments
+	volumeID := req.GetVolumeId()
+	if len(volumeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "ControllerPublishVolume Volume ID must be provided")
+	}
+	if err := s.driver.validateVolumeCapabilities([]*csi.VolumeCapability{req.GetVolumeCapability()}); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	nodeID := req.GetNodeId()
+	if len(nodeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "ControllerPublishVolume Node ID must be provided")
+	}
+
+	// Acquires the lock for the volume on that node only, because we need to support the ability
+	// to publish the same volume onto different nodes concurrently
+	lockingVolumeID := fmt.Sprintf("%s/%s", nodeID, volumeID)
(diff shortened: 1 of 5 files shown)


pkg/csi_driver/controller.go
```diff
@@ -127,6 +127,43 @@ func (s *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req *
 	}, nil
 }
 
+func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
+	// Validate arguments
+	volumeID := req.GetVolumeId()
+	if len(volumeID) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "ControllerPublishVolume Volume ID must be provided")
```
chrisThePattyEater, line 134: I don't think we need to log controllerpublish volume. Based on NPV logs it already says which call fails 

That applies to all error logs

chrisThePattyEater:
I think you are also going to need to add this to advertise the fact that the controller can do CPV
`cscap:   []*csi.ControllerServiceCapability{
			NewControllerServiceCapability(csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME),
		},`
here 
https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/d0b1500af4866e2d744aef362291da3631ddd0e1/pkg/csi_driver/gcs_fuse_driver.go#L94

Parallelstore also does it

chrisThePattyEater (approved):

uriel-guzman:
> I think you are also going to need to add this to advertise the fact that the controller can do CPV `cscap: []*csi.ControllerServiceCapability{ NewControllerServiceCapability(csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME), },` here https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/d0b1500af4866e2d744aef362291da3631ddd0e1/pkg/csi_driver/gcs_fuse_driver.go#L94
> 
> Parallelstore also does it https://screenshot.googleplex.com/Ayt8s6ohSuBqc4X

We can added this once @yaozile123 adds CSIDriverSpec attachRequired: true.

pkg/csi_driver/controller.go
uriel-guzman: NodePublishVolume logs the name too, even when it throws an error: https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/d0b1500af4866e2d744aef362291da3631ddd0e1/pkg/csi_driver/node.go#L93

I'd prefer to follow existing conventions in the codebase for consistency. If we one day decide to remove the name of the gRPC call, we can remove all of them at once.

PR #1416, opened 2026-06-02
Feat/proper sidecar override dedup

yangspirit:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR respects the user-specified `enable-sidecar-bucket-access-check` mount option in `NodePublishVolume` to allow overriding the global driver configuration (`EnableSidecarBucketAccessCheck`).
Specifically, it:
* Inspects the CSI volume mount options (`fuseMountOptions`) for `enable-sidecar-bucket-access-check`. If found, it overrides the driver's global default behavior for that volume.
* Registers `enable-sidecar-bucket-access-check` as an overwritable mount option in `joinMountOptions` within `pkg/csi_driver/utils.go`. This avoids duplicate flag generation by resolving the option key to its last occurrence.
* Implements helper functions `hasMountOption` and `getMountOptionValue` in `pkg/csi_driver/utils.go` for parsing key-value pairs in string array options.
* Adds comprehensive unit tests in `pkg/csi_driver/utils_test.go` (e.g. `TestHasMountOption`, `TestGetMountOptionValue`, and overwriting assertions in `TestJoinMountOptions`) to verify parsing and deduplication mechanics.
* Adds comprehensive unit tests in `pkg/csi_driver/node_test.go` (`TestNodePublishVolumeRespectEnableSidecarBucketAccessCheck`) to verify various combinations of global settings, sidecar version compatibility, and explicit user overrides.

Fixes # b/495760030

```release-note
Added support to respect and deduplicate user-provided `enable-sidecar-bucket-access-check` mount options, allowing per-volume overrides of the global sidecar bucket access check setting.
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -195,7 +195,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 		klog.V(6).Infof("NodePublishVolume populating identity provider %q in mount options", identityProvider)
 		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.OptInHnw + "=true", util.TokenServerIdentityProviderConst + "=" + identityProvider})
-	} else if enableSidecarBucketAccessCheckForSidecarVersion {
+	} else if enableSidecarBucketAccessCheckForSidecarVersion && !userExplicitlyEnabled {
 		//Enable sidecar bucket access check only for Workload Identity workloads. This feature consumes additional quota for Host Network pods as we do not have token caching.
 		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.EnableSidecarBucketAccessCheckConst + "=true"})
 	}
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -181,11 +181,11 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 	}
 
-sidecarCheckVal := getInternalMountOptionValue(args.fuseMountOptions, util.EnableSidecarBucketAccessCheckConst)
-userExplicitlyEnabled := sidecarCheckVal == util.TrueStr
-userExplicitlyDisabled := sidecarCheckVal == util.FalseStr
-enableSidecarBucketAccessCheckForSidecarVersion := s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarBucketAccessCheckMinVersion) &&
-	(userExplicitlyEnabled || (s.driver.config.EnableSidecarBucketAccessCheck && !userExplicitlyDisabled))
+	sidecarCheckVal := getInternalMountOptionValue(args.fuseMountOptions, util.EnableSidecarBucketAccessCheckConst)
+	userExplicitlyEnabled := sidecarCheckVal == util.TrueStr
+	userExplicitlyDisabled := sidecarCheckVal == util.FalseStr
+	enableSidecarBucketAccessCheckForSidecarVersion := s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarBucketAccessCheckMinVersion) &&
+		(userExplicitlyEnabled || (s.driver.config.EnableSidecarBucketAccessCheck && !userExplicitlyDisabled))
 	identityProvider := ""
 	if s.shouldPopulateIdentityProvider(pod, args.optInHostnetworkKSA, args.userSpecifiedIdentityProvider != "") {
 		if args.userSpecifiedIdentityProvider != "" {
@@ -197,7 +197,7 @@ enableSidecarBucketAccessCheckForSidecarVersion := s.driver.isSidecarVersionSupp
 		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.OptInHnw + "=true", util.TokenServerIdentityProviderConst + "=" + identityProvider})
 	} else if enableSidecarBucketAccessCheckForSidecarVersion {
 		//Enable sidecar bucket access check only for Workload Identity workloads. This feature consumes additional quota for Host Network pods as we do not have token caching.
-	args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.EnableSidecarBucketAccessCheckConst + "=true"})
+		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.EnableSidecarBucketAccessCheckConst + "=true"})
 	}
 
 	if enableSidecarBucketAccessCheckForSidecarVersion {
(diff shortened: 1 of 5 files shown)


pkg/csi_driver/utils.go
```diff
@@ -778,3 +779,26 @@ func transformKeysToSet(inputMap map[string]string) map[string]struct{} {
 	}
 	return outputSet
 }
+
+func hasMountOption(options []string, key string) bool {
```
uriel-guzman: GCSFuse mount options are **always** either in [CLI format](https://docs.cloud.google.com/storage/docs/cloud-storage-fuse/cli-options#options) or [config-file format](https://docs.cloud.google.com/storage/docs/cloud-storage-fuse/config-file#format-and-fields). 

If this function is generic, it should support both formats. So, I think either support config-file format too, or change the name to something else, like `hasInternalMountOption`. `enable-sidecar-bucket-access-check` is internal only, so it's only delimeted by a `=`, not `:` like config-file format.

pkg/csi_driver/node.go
```diff
@@ -181,7 +181,10 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 	}
 
-	enableSidecarBucketAccessCheckForSidecarVersion := s.driver.config.EnableSidecarBucketAccessCheck && s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarBucketAccessCheckMinVersion)
+	userExplicitlyEnabled := hasMountOption(args.fuseMountOptions, util.EnableSidecarBucketAccessCheckConst) && getMountOptionValue(args.fuseMountOptions, util.EnableSidecarBucketAccessCheckConst) == util.TrueStr
+	userExplicitlyDisabled := hasMountOption(args.fuseMountOptions, util.EnableSidecarBucketAccessCheckConst) && getMountOptionValue(args.fuseMountOptions, util.EnableSidecarBucketAccessCheckConst) == util.FalseStr
+	enableSidecarBucketAccessCheckForSidecarVersion := s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarBucketAccessCheckMinVersion) &&
+		(userExplicitlyEnabled || (s.driver.config.EnableSidecarBucketAccessCheck && !userExplicitlyDisabled))
```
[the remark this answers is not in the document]
uriel-guzman: +1, this suggestion also makes the code more readable.

pkg/csi_driver/node.go
```diff
@@ -193,7 +196,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.OptInHnw + "=true", util.TokenServerIdentityProviderConst + "=" + identityProvider})
 	} else if enableSidecarBucketAccessCheckForSidecarVersion {
 		//Enable sidecar bucket access check only for Workload Identity workloads. This feature consumes additional quota for Host Network pods as we do not have token caching.
-		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.EnableSidecarBucketAccessCheckConst + "=" + strconv.FormatBool(s.driver.config.EnableSidecarBucketAccessCheck)})
+		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{util.EnableSidecarBucketAccessCheckConst + "=" + strconv.FormatBool(enableSidecarBucketAccessCheckForSidecarVersion)})
```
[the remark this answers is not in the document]
uriel-guzman: +1, please apply suggestion.

pkg/csi_driver/utils.go
```diff
@@ -183,9 +183,10 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
 // joinMountOptions joins mount options eliminating duplicates.
 func joinMountOptions(existingOptions []string, newOptions []string) []string {
 	overwritableOptions := map[string]string{
-		"gid":       "",
-		"file-mode": "",
-		"dir-mode":  "",
+		"gid":                                    "",
+		"file-mode":                              "",
+		"dir-mode":                               "",
+		util.EnableSidecarBucketAccessCheckConst: "",
```
uriel-guzman: I'm not sure why we need this. Do we need it? In node.go, you're already changing the value to false or true, depending on the outcome of your logic, so in which case is this needed?

uriel-guzman (changes requested):

pkg/csi_driver/utils.go
yangspirit: Yes, without this, both options would remain in the final args.fuseMountOptions slice.

pkg/csi_driver/utils.go
yangspirit: Changed it to hasInternalMountOption

pkg/csi_driver/utils.go
```diff
@@ -778,3 +779,26 @@ func transformKeysToSet(inputMap map[string]string) map[string]struct{} {
 	}
 	return outputSet
 }
+
+func hasInternalMountOption(options []string, key string) bool {
```
uriel-guzman: Is this used? I just see it referenced in the unit tests. If not, please remove.

yangspirit: Done.

pkg/csi_driver/utils.go
uriel-guzman: if either `userExplicitlyEnabled` or `userExplicitlyDisabled` are true, then I don't think we need to pass `enable-sidecar-bucket-access-check` ourselves, right? That's the only scenario where you have both options in the mountOptions. Because the customer already passed it, why add it again?

Adding it to the map can be avoided by having the CSI driver only pass `enable-sidecar-bucket-access-check` when `userExplicitlyEnabled` and `userExplicitlyDisabled` are both false. That way, we don't need to add this in `overwritableOptions`.

uriel-guzman (changes requested):

pkg/csi_driver/utils.go
yangspirit: Done.

pkg/webhook/injection_test.go
```diff
@@ -1939,7 +1939,6 @@ func TestInjectMetadataPrefetchSidecarWithSC(t *testing.T) {
 			si.ScLister = informer.Storage().V1().StorageClasses().Lister()
 			si.PvLister = informer.Core().V1().PersistentVolumes().Lister()
 			si.PvcLister = informer.Core().V1().PersistentVolumeClaims().Lister()
-			informer.Start(ctx.Done())
```
uriel-guzman, line 1942: Why is this gone?

yangspirit: it's moved to line 1973  to fix a race condition. 
Before (with race condition): If informer.Start() is called before creating resources on fakeClient. 
After (fixed): By creating the resources first and then calling informer.Start(), the resources are included in the informer's initial List operation.

uriel-guzman (approved):

PR #1423, opened 2026-06-03
Fix ClusterName env var for failing prow tests

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Recently we have had a lot of test failures in our prow because there was a set of tests added that rely on CLUSTER_NAME env var it seems that the prow tests used to run these do not have access to this field are return nil. This issue is likely due to the difference in kube contexts between local runs and the prow env. To resolve the failing tests this pr implements a more robust parsing logic which fallsback to parsing the IDENTITY_PROVIDER value which is available both in prow and locally. The idenity provider format is "https://container.googleapis.com/v1/projects/prow-gob-internal-boskos-371/locations/us-central1/clusters/gcsfuseca8e". 

This change has been tested locally without the falback logic to verify that the parsing occurs as expected

Fixes #

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,14 +56,14 @@ ifeq ($(SELF_MANAGED_K8S), true)
 else
 # GKE-Specific Logic (Default)
     export WI_NODE_LABEL_CHECK ?= true
-    # assume that a GKE cluster identifier follows the format gke_{project-name}_{location}_{cluster-name}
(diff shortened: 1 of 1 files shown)


mattcary (approved):
We may want to think of something more maintainable than complicated shell logic pipes, but this should be effective for now.

PR #1426, opened 2026-06-04
remove fallback cluster details logic

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This is a follow up pr that removes fallback logic introduced in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1423. In the most recent run it was identified that the fallback logic does not work. Empty strings do not fallback to the Identity Provider logic. This pr removes the fallback logic and only uses provider.
 
Fixes #

```release-note

```


mattcary (approved):
Let's give it a shot!

PR #1437, opened 2026-06-08
Implement shared node mount support in the mutating webhook

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind feature

This PR implements shared node mount support in the mutating webhook of the GCS FUSE CSI Driver. 
Specifically, the webhook is updated to:

1. **Label PVs**: Automatically labels `PersistentVolume` objects using shared node mount with `gcsfuse.csi.storage.gke.io/shared-mount=true`.
2. **Control Sidecar Injection**: Bypasses sidecar injection for Pods using shared node mount volumes, and rejects Pods that mix using shared node mount and non-shared node mount GCSFuse volumes.
3. **Validate fsGroup**: Enforces that the Pod's `fsGroup` matches the one specified in the volume's `PodTemplate` for shared node mount volumes.
4. **Mutate for Profiles**: Injects GCSFuse profiles managed label and scheduling gate to the Pod spec when profiles are enabled for shared mount.

Fixes #

Tested e2e tests on non managed driver, using 10 n2-standard-4 nodes on GKE 1.36
```
Summarizing 9 Failures:
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should successfully authenticate multiple pods using same federation configuration
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should isolate workload identity federation access for Kubernetes service accounts with the same name across different namespaces
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should fail write operations when WI principal has read-only storage role
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should fail GCS access when WI principal has no storage role
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:382
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should fail GCS access when WI principal role is on a different bucket
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:382
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should successfully mount when pod KSA has WIF bucket access but node SA does not
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should fail GCS access after workload identity federation principal permissions are removed while pod is running
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should enforce different GCS bucket permissions for different Kubernetes service accounts
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325
    [FAIL] E2E Test Suite [Driver: gcsfuse.csi.storage.gke.io] [Testpattern: CSI Ephemeral-volume (default fs)] workload-identity-federation [It] should re-authenticate successfully after pod restart using federation
    /usr/local/google/home/yaozile/oss/gcs-fuse-csi-driver/test/e2e/specs/specs.go:325

  Ran 468 of 596 Specs in 7132.322 seconds
  FAIL! -- 459 Passed | 9 Failed | 0 Pending | 128 Skipped


  Ginkgo ran 1 suite in 1h59m5.624547168s

  Test Suite Failed
```
All the test are passed except the workload identity test cases, those will be fixed in another PR

```release-note
NONE

### DIFF
--- a/pkg/webhook/client.go
+++ b/pkg/webhook/client.go
@@ -94,3 +94,13 @@ func (si *SidecarInjector) GetVolumesStorageClass(volume *corev1.PersistentVolum
 	}
 	return sc, nil
 }
+
+// GetPodTemplate retrieves the referenced PodTemplate object from the informer cache.
+func (si *SidecarInjector) GetPodTemplate(namespace, name string) (*corev1.PodTemplate, error) {
+	podTemplate, err := si.PodTemplateLister.PodTemplates(namespace).Get(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return podTemplate, nil
+}
--- a/pkg/webhook/mutatingwebhook.go
+++ b/pkg/webhook/mutatingwebhook.go
@@ -73,13 +73,14 @@ type SidecarInjector struct {
 	PvcLister              listersv1.PersistentVolumeClaimLister
 	PvLister               listersv1.PersistentVolumeLister
 	ScLister               listerstoragev1.StorageClassLister
+	PodTemplateLister      listersv1.PodTemplateLister
 	ServerVersion          *version.Version
 	K8SClient              kubernetes.Interface
 }
 
 // Handle injects a gcsfuse sidecar container and a emptyDir to incoming qualified pods.
 func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) admission.Response {
-	if req.Kind.Kind == "PersistentVolume" && si.Config.EnableGcsfuseProfiles { // Currently only handling pvs for gcsfuse profiles
+	if req.Kind.Kind == "PersistentVolume" {
 		pv := &corev1.PersistentVolume{}
 		if err := si.Decoder.Decode(req, pv); err != nil {
 			klog.Errorf("Could not decode PersistentVolume object: %v", err)
@@ -87,10 +88,33 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 		}
 		klog.Infof("Mutating webhook is handling PersistentVolume: %s", pv.Name)
 
-		if err := si.validatePersistentVolumeForGCSFuseProfiles(pv); err != nil {
-			klog.Errorf("PersistentVolume validation failed: %v", err)
-			return admission.Errored(http.StatusBadRequest, err)
+		if pv.Spec.CSI == nil || pv.Spec.CSI.Driver != util.GCSFuseCsiDriverName {
+			// Not a gcsfuse CSI driver volume, skip validation/mutation
+			return admission.Allowed(fmt.Sprintf("No mutation Required on PersistentVolume: %s", pv.Name))
 		}
+
+		if si.Config.EnableGcsfuseProfiles {
+			if err := si.validatePersistentVolumeForGCSFuseProfiles(pv); err != nil {
+				klog.Errorf("PersistentVolume validation failed: %v", err)
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+		}
+
+		// Patch the label to PV if it uses shared node mount.
+		if si.Config.EnableSharedNodeMount && pv.Spec.CSI.VolumeAttributes != nil && pv.Spec.CSI.VolumeAttributes[SharedMountVolumeAttribute] == util.TrueStr {
+			if pv.Labels == nil {
+				pv.Labels = make(map[string]string)
+			}
+			if pv.Labels[SharedMountLabel] != util.TrueStr {
+				pv.Labels[SharedMountLabel] = util.TrueStr
+				marshaledPV, err := json.Marshal(pv)
+				if err != nil {
+					return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to marshal pv: %w", err))
+				}
+				return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPV)
+			}
+		}
+
 		return admission.Allowed(fmt.Sprintf("No mutation Required on PersistentVolume: %s", pv.Name))
 	}
 
@@ -105,6 +129,91 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 
 	if req.Operation != admissionv1.Create {
 		return admission.Allowed(fmt.Sprintf("No injection required for operation %v.", req.Operation))
+	}
+
+	// Classify Pod volumes to check if they are GCSFuse volumes and if they use shared node mount.
+	var hasSharedNodeMount bool
+	var hasStandardMount bool
+	var cacheCreatedByUser bool
+
+	for _, volume := range pod.Spec.Volumes {
+		isGcsfuse, _, volumeAttributes, pv, err := si.isGcsFuseCSIVolume(volume, pod.Namespace)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("PVC does not exist, skipping volume %q: %v", volume.Name, err)
+				continue
+			}
+			klog.Errorf("Failed to check if volume %q is GCSFuse volume: %v", volume.Name, err)
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		if !isGcsfuse {
+			continue
+		}
+
+		isSharedNodeMount := si.Config.EnableSharedNodeMount && pv != nil && volumeAttributes != nil && volumeAttributes[SharedMountVolumeAttribute] == util.TrueStr
+
+		if isSharedNodeMount {
+			hasSharedNodeMount = true
+			if volume.PersistentVolumeClaim != nil {
+				pvc, err := si.GetPVC(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
+				if err != nil {
+					return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to get PVC %q: %w", volume.PersistentVolumeClaim.ClaimName, err))
+				}
+
+				podTemplate, err := si.getMounterPodTemplate(pod.Namespace, pvc)
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						klog.Warningf("Skip volume %q verification because the referenced PodTemplate was not found: %v", volume.Name, err)
+						continue
+					}
+					klog.Errorf("failed to get mounter PodTemplate for PVC %q: %v", pvc.Name, err)
+					return admission.Errored(http.StatusInternalServerError, err)
+				}
+				if podTemplate == nil {
+					return admission.Errored(http.StatusBadRequest, fmt.Errorf("volume %q uses shared node mount but its PVC %q is missing a referenced PodTemplate annotation %q", volume.Name, pvc.Name, MounterPodTemplateAnnotation))
+				}
+
+				if volumeExists(podTemplate.Template.Spec.Volumes, SidecarContainerCacheVolumeName) {
+					cacheCreatedByUser = true
+				}
+
+				// Validate that Pod fsGroup matches the requirements of the referenced PVC PodTemplate.
+				if err := si.validateFSGroup(pod, podTemplate); err != nil {
+					klog.Errorf("fsGroup validation failed: %v", err)
+					return admission.Errored(http.StatusBadRequest, err)
+				}
+			}
+		} else {
+			hasStandardMount = true
+		}
+	}
+
+	// Prevent mixing shared node mount and standard mount GCSFuse volumes in the same Pod.
+	if hasSharedNodeMount && hasStandardMount {
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("mixing shared node mount and non-shared node mount GCSFuse volumes in the same Pod is not allowed"))
+	}
+
+	// Handle Pod mutation and validations for shared node mount mode.
+	if hasSharedNodeMount {
+		// Inject profiles label and scheduling gate if storage profiles are enabled and match.
+		if si.Config.EnableGcsfuseProfiles {
+			profilesEnabled, err := si.IsGCSFuseProfilesEnabled(pod)
+			if err != nil {
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+			if profilesEnabled {
+				if err := ModifyPodSpecForGCSFuseProfiles(pod, cacheCreatedByUser, true); err != nil {
+					return admission.Errored(http.StatusBadRequest, err)
+				}
+				marshaledPod, err := json.Marshal(pod)
+				if err != nil {
+					return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to marshal pod: %w", err))
+				}
+				return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+			}
+		}
+
+		return admission.Allowed("No sidecar injection required for shared node mount mode.")
 	}
 
 	enableGcsfuseVolumes, ok := pod.Annotations[GcsFuseVolumeEnableAnnotation]
@@ -177,7 +286,7 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 		pod.Spec.Volumes = append(pod.Spec.Volumes, GetSATokenVolume(audience))
 	}
 
-	cacheCreatedByUser := volumeExists(pod.Spec.Volumes, SidecarContainerCacheVolumeName)
+	cacheCreatedByUser = volumeExists(pod.Spec.Volumes, SidecarContainerCacheVolumeName)
 	pod.Spec.Volumes = append(GetSidecarContainerVolumeSpec(pod.Spec.Volumes...), pod.Spec.Volumes...)
 
 	// Inject metadata prefetch sidecar.
@@ -198,7 +307,7 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 		}
 		klog.Infof("Pod %q has at least one volume using gcsfuse profiles", pod.Name)
 		if profilesEnabled {
-			err = ModifyPodSpecForGCSFuseProfiles(pod, cacheCreatedByUser)
+			err = ModifyPodSpecForGCSFuseProfiles(pod, cacheCreatedByUser, false)
 			if err != nil {
 				return admission.Errored(http.StatusBadRequest, err)
 			}
@@ -251,7 +360,7 @@ func audienceForInjectedSATokenVolume(projectID string, pod *corev1.Pod) string 
 }
 
 // Modifies the pod spec to add gcsfuse profile related features. This includes adding a label, scheduling gate, and placeholder file cache volumes
-func ModifyPodSpecForGCSFuseProfiles(pod *corev1.Pod, cacheCreatedByUser bool) error {
+func ModifyPodSpecForGCSFuseProfiles(pod *corev1.Pod, cacheCreatedByUser bool, isSharedMount bool) error {
 	// Always apply the gcsfuse profile label when gcsfuse profiles are enabled for pod informer's Kubernetes API efficient filtering
 	if pod.Labels == nil {
 		pod.Labels = make(map[string]string)
@@ -278,6 +387,10 @@ func ModifyPodSpecForGCSFuseProfiles(pod *corev1.Pod, cacheCreatedByUser bool) e
 		pod.Spec.SchedulingGates = append(pod.Spec.SchedulingGates, profilesGate)
 	} else {
 		klog.Warningf("Pod %s/%s already has the %s scheduling gate, skipping injection of gcsfuse profiles scheduling gate.", pod.Namespace, pod.Name, BucketScanPendingSchedulingGate)
+	}
+
+	if isSharedMount {
+		return nil
 	}
 
 	// Inject placeholder file cache volumes
@@ -502,4 +615,47 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
 	}
 
 	return filename, &credConfig, nil
+}
+
+// getMounterPodTemplate retrieves the referenced mounter PodTemplate for a PVC if the annotation is present.
+func (si *SidecarInjector) getMounterPodTemplate(namespace string, pvc *corev1.PersistentVolumeClaim) (*corev1.PodTemplate, error) {
+	if pvc.Annotations == nil {
+		return nil, nil
+	}
+	templateName, ok := pvc.Annotations[MounterPodTemplateAnnotation]
+	if !ok || templateName == "" {
+		return nil, nil
+	}
+
+	podTemplate, err := si.GetPodTemplate(namespace, templateName)
+	if err != nil {
+		return nil, err
+	}
+	return podTemplate, nil
+}
+
+// validateFSGroup checks if the Pod's fsGroup matches the one specified in the volume's referenced mounter PodTemplate.
+func (si *SidecarInjector) validateFSGroup(pod *corev1.Pod, podTemplate *corev1.PodTemplate) error {
+	templateName := podTemplate.Name
+	templateHasFSGroup := podTemplate.Template.Spec.SecurityContext != nil && podTemplate.Template.Spec.SecurityContext.FSGroup != nil
+	podHasFSGroup := pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.FSGroup != nil
+
+	// Enforce strict matching between Pod and PodTemplate fsGroup settings to prevent shared mount permission conflicts.
+	if templateHasFSGroup {
+		expectedFSGroup := *podTemplate.Template.Spec.SecurityContext.FSGroup
+		if !podHasFSGroup {
+			return fmt.Errorf("Pod fsGroup is not set, but volume's PodTemplate %q requires fsGroup %d", templateName, expectedFSGroup)
+		}
+		actualFSGroup := *pod.Spec.SecurityContext.FSGroup
+		if actualFSGroup != expectedFSGroup {
+			return fmt.Errorf("Pod fsGroup %d does not match the one specified in volume's PodTemplate %q (%d)", actualFSGroup, templateName, expectedFSGroup)
+		}
+		return nil
+	}
+
+	if podHasFSGroup {
+		return fmt.Errorf("Pod has fsGroup set, but volume's PodTemplate %q does not specify one (not allowed for shared node mount)", templateName)
+	}
+
+	return nil
 }
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -30,8 +30,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
-	storagelisters "k8s.io/client-go/listers/storage/v1"
 	"k8s.io/client-go/rest"
+
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: pkg/webhook/mutatingwebhook.go [hunk 1]
```diff
@@ -503,3 +595,107 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+		}
+
+		if pvc.Spec.VolumeName != "" {
+			pv, err := si.PvLister.Get(pvc.Spec.VolumeName)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return false, false, nil
+				}
+				return false, false, err
+			}
```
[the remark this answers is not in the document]
yaozile123: I have refactored GetPV() and GetPVC() to include this logic

pkg/webhook/client.go
```diff
@@ -65,19 +68,35 @@ func (si *SidecarInjector) GetPreprovisionCSIVolume(csiDriver string, pvc *corev
...
 func (si *SidecarInjector) GetPV(name string) (*corev1.PersistentVolume, error) {
 	pv, err := si.PvLister.Get(name)
 	if err != nil {
-		return nil, err
+		if apierrors.IsNotFound(err) && si.K8SClient != nil {
+			pv, err = si.K8SClient.CoreV1().PersistentVolumes().Get(context.Background(), name, metav1.GetOptions{})
+		}
+		if err != nil {
+			return nil, err
+		}
```
uriel-guzman: Why do we need to do this in this PR? This is not an existing pattern across the repo. It's also an inconsistent change because for GetVolumeStorageClass in this same file you're not doing it: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/764e727ace683a1b10b04160e20e302070136406/pkg/webhook/client.go#L91

I would suggest to just revert this and not do it, since it's out of scope of this PR's goal to avoid unexpected breakages.

pkg/webhook/client.go
```diff
@@ -65,19 +68,35 @@ func (si *SidecarInjector) GetPreprovisionCSIVolume(csiDriver string, pvc *corev
...
 
+// GetPVC retrieves the PVC associated with the given namespace and name.
+// It first attempts to retrieve the PVC from the informer cache.
+// Falling back to a direct API call using si.K8SClient when the PVC is not found in the cache.
 func (si *SidecarInjector) GetPVC(namespace, name string) (*corev1.PersistentVolumeClaim, error) {
 	pvc, err := si.PvcLister.PersistentVolumeClaims(namespace).Get(name)
 	if err != nil {
-		return nil, err
+		if apierrors.IsNotFound(err) && si.K8SClient != nil {
+			pvc, err = si.K8SClient.CoreV1().PersistentVolumeClaims(namespace).Get(context.Background(), name, metav1.GetOptions{})
```
uriel-guzman: Same comment, I don't think this contributes to the PR's goal and seems like an unrelated enhncement. I'd suggest to not scope creep in this and future PRs, since we don't want to change existing behaviors.

pkg/webhook/mutatingwebhook.go
```diff
@@ -56,6 +56,9 @@ const (
 	metadataPrefetchMemoryRequestAnnotation          = "gke-gcsfuse/metadata-prefetch-memory-request"
 	GCPWorkloadIdentityCredentialConfigMapAnnotation = "gke-gcsfuse/workload-identity-credential-configmap"
 	NumaPinningAnnotation                            = "gke-gcsfuse/enable-numa-pinning"
+	SharedMountLabel                                 = "gke-gcsfuse/shared-mount"
+	MounterPodTemplateAnnotation                     = "gke-gcsfuse/mounter-pod-template"
+	SharedMountVolumeAttribute                       = "sharedMount"
```
uriel-guzman: There's already a place in the webhook package where webhook relevant volume attributes go. Let's move this to the appropriate place: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/764e727ace683a1b10b04160e20e302070136406/pkg/webhook/sidecar_spec.go#L45

pkg/webhook/client.go
yaozile123: Got it, I did this for resolving the comments leaved by Gemini. I will revert this change

pkg/webhook/mutatingwebhook.go
```diff
@@ -79,18 +82,44 @@ type SidecarInjector struct {
...
+			if err := si.validatePersistentVolumeForGCSFuseProfiles(pv); err != nil {
+				klog.Errorf("PersistentVolume validation failed: %v", err)
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+		}
+
+		// Patch the label to PV if it uses shared node mount.
+		mutated := false
+		if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == util.GCSFuseCsiDriverName {
+			if pv.Spec.CSI.VolumeAttributes != nil && pv.Spec.CSI.VolumeAttributes[SharedMountVolumeAttribute] == "true" {
```
uriel-guzman: Nit: Use `util.TrueStr` instead of `"true"`: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/764e727ace683a1b10b04160e20e302070136406/pkg/csi_driver/node.go#L277

pkg/webhook/mutatingwebhook.go
```diff
@@ -79,18 +82,44 @@ type SidecarInjector struct {
...
+				if pv.Labels == nil {
+					pv.Labels = make(map[string]string)
+				}
+				if pv.Labels[SharedMountLabel] != "true" {
```
uriel-guzman: Same here, use the const `util.TrueStr`.

pkg/webhook/mutatingwebhook.go
```diff
@@ -79,18 +82,44 @@ type SidecarInjector struct {
...
+		if si.Config.EnableGcsfuseProfiles {
+			if err := si.validatePersistentVolumeForGCSFuseProfiles(pv); err != nil {
+				klog.Errorf("PersistentVolume validation failed: %v", err)
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+		}
+
+		// Patch the label to PV if it uses shared node mount.
+		mutated := false
+		if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == util.GCSFuseCsiDriverName {
```
chrisThePattyEater: We do this same check in validatePersistentVolumeForGCSFuseProfiles, can we just check once early on to minimize if blocks. ex
if pv.Spec.CSI == nil || pv.Spec.CSI.Driver != util.GCSFuseCsiDriverName {
		// Not a gcsfuse CSI driver volume, skip validation
		return admission.Allowed(fmt.Sprintf("No mutation Required on PersistentVolume: %s", pv.Name))
	}

pkg/webhook/client.go
yaozile123: Done

pkg/webhook/mutatingwebhook.go
```diff
@@ -79,18 +82,44 @@ type SidecarInjector struct {
...
+				}
+				if pv.Labels[SharedMountLabel] != "true" {
+					pv.Labels[SharedMountLabel] = "true"
+					mutated = true
+				}
+			}
 		}
+
+		// Return patch response if PV was mutated, otherwise allow PV creation as-is.
+		if mutated {
```
chrisThePattyEater: Why do we need this flag, we know that when pv.Spec.CSI.VolumeAttributes[SharedMountVolumeAttribute] is "true" we will patch the label. We should always do this logic at the end of the `pv.Spec.CSI.VolumeAttributes != nil && pv.Spec.CSI.VolumeAttributes[SharedMountVolumeAttribute] == "true"` block

yaozile123: Done, I also applied the change for Pod volumes below

pkg/webhook/mutatingwebhook.go
yaozile123: Done

pkg/webhook/mutatingwebhook.go
yaozile123: Done

pkg/webhook/mutatingwebhook.go
yaozile123: Done, moved to sidecar_spec.go

pkg/webhook/mutatingwebhook.go
yaozile123: I have moved the check to the beginning of the Handle method. Should we also remove the redundant check inside validatePersistentVolumeForGCSFuseProfiles?

pkg/webhook/mutatingwebhook.go
```diff
@@ -107,6 +136,69 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 		return admission.Allowed(fmt.Sprintf("No injection required for operation %v.", req.Operation))
 	}
 
+	// Classify Pod volumes to check if they are GCSFuse volumes and if they use shared node mount.
+	var hasSharedMount bool
+	var hasStandardMount bool
+
+	for _, volume := range pod.Spec.Volumes {
+		isGcsfuse, isSharedMount, err := si.checkVolumeProperties(volume, pod.Namespace)
```
chrisThePattyEater: This is a nit so ignore or comment if you disagree but: It feels like it we are mixing two separate concerns: identifying if a volume belongs to the GCSFuse driver, and evaluating its internal configuration attributes (sharedMount).

It would be cleaner to split this logic:
isGCSFuseVolume(volume, namespace) (bool, error)
isSharedMountVolume(volume, namespace) (bool, error)

or do the check first and then isSharedMountVolume

pkg/webhook/mutatingwebhook.go
```diff
@@ -503,3 +583,107 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+		// Ephemeral inline volume does not support sharedMount.
+		return true, false, nil
+	}
+
+	// 2. Check PVC
+	if volume.PersistentVolumeClaim != nil {
+		pvcName := volume.PersistentVolumeClaim.ClaimName
+		pvc, err := si.GetPVC(namespace, pvcName)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
```
chrisThePattyEater: If you do separate concerns can we name this something more clear, ex isSharedNodeMounVolume or something?

pkg/webhook/mutatingwebhook.go
yaozile123: I have split it into two methods

pkg/webhook/mutatingwebhook.go
yaozile123: Done, I also split the method as discussed above

uriel-guzman (changes requested):

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +589,146 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+		return true, nil
+	}
+
+	// 2. Check PVC
+	if volume.PersistentVolumeClaim != nil {
+		pvcName := volume.PersistentVolumeClaim.ClaimName
+		pvc, err := si.GetPVC(namespace, pvcName)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
```
uriel-guzman: Log a warning here. Customers should usually create their Pods after their PVCs, or weird things can happen.

uriel-guzman (changes requested):

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +589,146 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+			}
+			return false, err
+		}
+
+		if pvc.Spec.VolumeName != "" {
+			pv, err := si.GetPV(pvc.Spec.VolumeName)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
```
uriel-guzman: Same, log a warning here, the PV is expected to exist before the workload.

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +589,146 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+					return false, nil
+				}
+				return false, err
+			}
+			if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == util.GCSFuseCsiDriverName {
```
uriel-guzman: This logic should be the very first thing that is checked. It should be moved up to line 612, immediately after you even check for an error from the PVC get command. Reasoning is that we don't want to reject the PVC creations from non-gcsfuse volumes, even if they have an error.

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +589,146 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+				return true, nil
+			}
+		} else if pvc.Spec.StorageClassName != nil && *pvc.Spec.StorageClassName != "" && si.ScLister != nil {
```
uriel-guzman: This logic says that if a PVC isn't bound to a PV, check StorageClass as fallback. However, I'm not sure this makes a lot of sense, because if a customer isn't using profiles (StorageClass), they'll just have some dummy StorageClass name, and if that isn't found, it's going to always return false? This makes the logic a bit inconsitent.

I think we should simplify this to just make it so that, if a PVC isn't bound yet (doesn't have the PV volume name), just log a warning instead.

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +589,146 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+}
+
+// isSharedNodeMountVolume checks if the GCSFuse volume is configured with sharedNodeMount.
+func (si *SidecarInjector) isSharedNodeMountVolume(volume corev1.Volume, namespace string) (bool, error) {
+	// 1. Check CSI Ephemeral inline volume. sharedNodeMount does not support ephemeral inline volume.
+	if volume.CSI != nil && volume.CSI.Driver == util.GCSFuseCsiDriverName {
+		return false, nil
+	}
+
+	// 2. Check PVC
```
uriel-guzman: This logic is almost all the same as isGCSFuseVolume. Please wrap the common logic into helper functions and call the helper functions instead of duplicating the code. For example, in both functions, L635 smells like a function that can be called `isGCSFuseEphemeralVolume`, and L641 can be called `pvcFromVolume`, and L652 can be called `pvFromPVC`, then do the relevant checks. Consider doing whatever you make sense, but let's just try to get rid of the duplicate code as much as possible.

uriel-guzman (changes requested):

pkg/webhook/mutatingwebhook.go
yaozile123: Done

pkg/webhook/mutatingwebhook.go
yaozile123: Done

pkg/webhook/mutatingwebhook.go
yaozile123: Done, I have added the warning

pkg/webhook/mutatingwebhook.go
yaozile123: Done, I have added several helper functions

pkg/webhook/mutatingwebhook.go
yaozile123: Could you elaborate on how to check if it's a GCSFuse volume when `GetPVC` returns an error, since we wouldn't have the volume details yet? 

To avoid rejecting non-gcsfuse volumes, should we simply log a warning and return `false, nil` instead of returning the error?

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +589,160 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+	}
+	return si.GetPV(pvc.Spec.VolumeName)
+}
+
+func isGCSFusePV(pv *corev1.PersistentVolume) bool {
+	return pv.Spec.CSI != nil && pv.Spec.CSI.Driver == util.GCSFuseCsiDriverName
+}
+
+// isGCSFuseVolume checks if the volume is a GCSFuse volume.
+func (si *SidecarInjector) isGCSFuseVolume(volume corev1.Volume, namespace string) (bool, error) {
```
uriel-guzman: Can we just use the existing function for this? https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/volumes.go#L37. This already seems to be returning all the values you're interested in (checking if it's an ephemeral volume, isGCSFuseCSIVolume, PV).

yaozile123: Done, I also updated the method to include some warnings when pv/pvc not found.

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +585,64 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+		if podHasFSGroup {
+			return fmt.Errorf("Pod has fsGroup set, but volume's PodTemplate %q does not specify one (not allowed for shared node mount)", templateName)
+		}
+	}
+
+	return nil
+}
+
+// ModifyPodSpecForSharedMountProfiles injects profiles label and scheduling gate to the workload Pod spec for shared mount profiles.
+func ModifyPodSpecForSharedMountProfiles(pod *corev1.Pod) {
```
uriel-guzman: Can we just extend https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/mutatingwebhook.go#L307 instead of creating a new function with duplicate logic? It seems like from this [line](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/mutatingwebhook.go#L336) up until the end of the function, we could just add a new boolean parameter called `isSharedMount`. If it is skip all that sidecar stuff. Otherwise, it seems like the rest of the logic applies to shared mount too?

uriel-guzman (changes requested):

pkg/webhook/mutatingwebhook.go
yaozile123: Done

pkg/webhook/mutatingwebhook.go
```diff
@@ -113,6 +130,80 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+			hasSharedNodeMount = true
+			if volume.PersistentVolumeClaim != nil {
+				pvc, err := si.GetPVC(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
+				if err != nil {
+					return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to get PVC %q: %w", volume.PersistentVolumeClaim.ClaimName, err))
+				}
+
+				podTemplate, err := si.getMounterPodTemplate(ctx, pod.Namespace, pvc)
+				if err != nil {
+					return admission.Errored(http.StatusBadRequest, err)
```
uriel-guzman: If the PodTemplate and the workload are created at the same time, it's possible there's a race condition where the informer's cache isn't synced yet with the PodTemplate and this returns a blocking error to the customer.

I think instead we should relax the validation re-use take the metadata pre-fetch pattern: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L227

In metadata pre-fetch, if the volume doesn't exist, it just logs a non-blocking error with klog.Error. This should be okay, because in the rare scenario that this happens, we should have a check during NodePublishVolume where we validate the fsGroup against the mounter pod's fsGroup anyway, and we can just skip the validation of fsGroup here and anything else related to PodTemplate in the future.

yaozile123: Done, I updated it to log a non blocking error and skip the validation here

pkg/webhook/mutatingwebhook.go
```diff
@@ -113,6 +130,81 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+		if !isGcsfuse {
+			continue
+		}
+
+		isSharedNodeMount := pv != nil && volumeAttributes != nil && volumeAttributes[SharedMountVolumeAttribute] == util.TrueStr
+
+		if isSharedNodeMount {
+			hasSharedNodeMount = true
+			if volume.PersistentVolumeClaim != nil {
+				pvc, err := si.GetPVC(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
```
uriel-guzman, line 158: This behavior isn't consistent with metadata pre-fetch. In metadata pre-fetch, similar to the PodTemplate discussion, if the volume isn't found, it just logs a warning and continues: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6630691093631f1c942a26bc1e0e628a39c2b4fd/pkg/webhook/volumes.go#L69

The current behavior of this PR is that it will block the Pod create API request. I think we should also relax the validation here, and same for the PV when it doesn't exist.

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +605,52 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+	templateName, ok := pvc.Annotations[MounterPodTemplateAnnotation]
+	if !ok || templateName == "" {
+		return nil, nil
+	}
+
+	if si.K8SClient == nil {
+		return nil, fmt.Errorf("kubernetes client is uninitialized, cannot fetch PodTemplate %q", templateName)
+	}
+
+	podTemplate, err := si.K8SClient.CoreV1().PodTemplates(namespace).Get(ctx, templateName, metav1.GetOptions{})
```
uriel-guzman: We should add a PodTemplate informer instead of directly querying the API server. This will potentially be executed O(P * V) times, P = num pods, V = num volumes. At high scale, P could be dozens of thousands, and V is also unbounded, which can overwhelm the API server. A PodTemplate informer will serve the Get directly from the in-memory cache, making it much more efficient.

pkg/webhook/mutatingwebhook.go
```diff
@@ -561,3 +605,52 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+	if templateHasFSGroup {
+		expectedFSGroup := *podTemplate.Template.Spec.SecurityContext.FSGroup
+		if !podHasFSGroup {
+			return fmt.Errorf("Pod fsGroup is not set, but volume's PodTemplate %q requires fsGroup %d", templateName, expectedFSGroup)
+		}
+		actualFSGroup := *pod.Spec.SecurityContext.FSGroup
+		if actualFSGroup != expectedFSGroup {
+			return fmt.Errorf("Pod fsGroup %d does not match the one specified in volume's PodTemplate %q (%d)", actualFSGroup, templateName, expectedFSGroup)
+		}
+	} else {
```
uriel-guzman: Nit: Inside `if templateHasFSGroup`, you can return nil at the end of the block. Then, you de-nest the subsequent `else` statement`. This improves readability since it reduces the number of nested conditionals.

pkg/webhook/volumes.go
```diff
@@ -60,14 +61,24 @@ func (si *SidecarInjector) isGcsFuseCSIVolume(volume corev1.Volume, namespace st
 	pvcName := pvc.ClaimName
 	pvcObj, err := si.GetPVC(namespace, pvcName)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warningf("PVC %s/%s is not found. PVCs should exist before Pods.", namespace, pvcName)
```
uriel-guzman: Currently, if an error is returned from `isGcsFuseCSIVolume`, it logs a klog.Errorf: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L227

Because you are now returning nil from `isGcsFuseCSIVolume` and logging a warning, when the volume doesn't exist, now it's going to be demoted to a klog.Warningf.

To ensure we don't change existing logging severity behavior, I think we should not modify `isGcsFuseCSIVolume` directly, and instead, handle the error from within the calling function. From the calling function, you can check if it's NotFound, and then log the error from there and skip.

pkg/webhook/mutatingwebhook.go
yaozile123: Done, I have added the PodTemplate informer and relevant unit tests

pkg/webhook/mutatingwebhook.go
yaozile123: Done

pkg/webhook/volumes.go
yaozile123: Done. I kept `isGcsFuseCSIVolume` unmodified to preserve existing error logging severity, and handled the error handling at the callers instead. For PodTemplate and PVC NotFound errors, we log a warning and skip, while returning admission.Errored for all other errors. For PV, the `isGcsFuseCSIVolume` will return `false` if is not found, which should also skip the following check.

pkg/webhook/mutatingwebhook.go
yaozile123: Done

uriel-guzman (approved):

pkg/webhook/mutatingwebhook.go
```diff
@@ -503,3 +614,49 @@ func parseCredentialConfigurationConfigMap(client kubernetes.Interface, pod *cor
...
+	}
+
+	podTemplate, err := si.GetPodTemplate(namespace, templateName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("referenced PodTemplate %q in PVC %s/%s was not found: %w", templateName, namespace, pvc.Name, err)
+		}
+		return nil, fmt.Errorf("failed to get PodTemplate %q: %w", templateName, err)
+	}
+	return podTemplate, nil
```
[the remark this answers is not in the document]
uriel-guzman: Please fix.

pkg/webhook/mutatingwebhook.go
```diff
@@ -107,6 +131,89 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+
+		if isSharedNodeMount {
+			hasSharedNodeMount = true
+			if volume.PersistentVolumeClaim != nil {
+				pvc, err := si.GetPVC(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
+				if err != nil {
+					return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to get PVC %q: %w", volume.PersistentVolumeClaim.ClaimName, err))
+				}
+
+				podTemplate, err := si.getMounterPodTemplate(ctx, pod.Namespace, pvc)
```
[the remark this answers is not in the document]
uriel-guzman: Please remove unused param.

pkg/webhook/mutatingwebhook.go
```diff
@@ -107,6 +131,89 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
...
+	// Handle Pod mutation and validations for shared node mount mode.
+	if hasSharedNodeMount {
+		// Inject profiles label and scheduling gate if storage profiles are enabled and match.
+		if si.Config.EnableGcsfuseProfiles {
+			profilesEnabled, err := si.IsGCSFuseProfilesEnabled(pod)
+			if err != nil {
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+			if profilesEnabled {
+				_ = ModifyPodSpecForGCSFuseProfiles(pod, cacheCreatedByUser, true)
```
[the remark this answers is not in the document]
uriel-guzman: Please handle error correctly.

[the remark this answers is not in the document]
yaozile123: Fixed

pkg/webhook/mutatingwebhook.go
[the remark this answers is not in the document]
yaozile123: Done

pkg/webhook/mutatingwebhook.go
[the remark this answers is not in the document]
yaozile123: Fixed, sorry I forgot to update here when we decide to handle not found error in the caller

PR #1441, opened 2026-06-09
[Shared Mount] Verify Mounter Pod exists during NodeStageVolume

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This is the second pr for nodestage volume it creates the skeleton NSV Execution by retrieving the mounter pod (fails if it doesn't exist), checking if the staging path is already mounted (succeeding if it is), and finally creating the staging path.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -671,11 +671,76 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
+	publishContext := req.GetPublishContext()
+	if publishContext == nil {
+		return nil, status.Error(codes.InvalidArgument, "publishContext must be provided")
+	}
+
+	podName, ok := publishContext[PublishContextKeyMounterPodName]
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "publishContext must contain mounter pod name")
+	}
+	if podName == "" {
+		return nil, status.Error(codes.InvalidArgument, "mounter pod name in publishContext cannot be empty")
+	}
+
+	podNamespace, ok := publishContext[PublishContextKeyMounterPodNamespace]
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "publishContext must contain mounter pod namespace")
+	}
+	if podNamespace == "" {
+		return nil, status.Error(codes.InvalidArgument, "mounter pod namespace in publishContext cannot be empty")
+	}
+
 	// Acquire a lock on the staging path.
 	if acquired := s.volumeLocks.TryAcquire(stagingPath); !acquired {
 		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, stagingPath)
 	}
 	defer s.volumeLocks.Release(stagingPath)
 
+	resp, err := s.executeNodeStageVolume(ctx, req)
+
+	if err != nil {
+		klog.Errorf("NodeStageVolume failed on staging path %q for volume %q: %v)", stagingPath, volumeID, err)
+	}
+
+	return resp, err
+}
+
+func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
+	clientset := s.driver.config.K8sClients
+	stagingPath := req.GetStagingTargetPath()
+
+	// Validate staging path and mounter pod information from the request.
+	publishContext := req.GetPublishContext()
+	podName := publishContext[PublishContextKeyMounterPodName]
+	podNamespace := publishContext[PublishContextKeyMounterPodNamespace]
+
+	klog.Infof("Executing NodeStageVolume. Mounter pod: %s/%s, node: %q, volume: %q, staging path: %q", podNamespace, podName, s.driver.config.NodeID, req.GetVolumeId(), stagingPath)
+
+	// Verify mounter pod exists.
+	pod, err := clientset.GetPod(podNamespace, podName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.FailedPrecondition, "mounter pod %s/%s expected to exist but was not found", podNamespace, podName)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get mounter pod %s/%s: %v", podNamespace, podName, err)
+	}
+	if pod == nil {
+		return nil, status.Errorf(codes.Internal, "mounter pod %s/%s can't be nil", podNamespace, podName)
+	}
+
+	// Make the staging path.
+	klog.Infof("NodeStageVolume attempting mkdir for staging path %q", stagingPath)
+	if err := os.MkdirAll(stagingPath, 0750); err != nil {
+		return nil, status.Errorf(codes.Internal, "mkdir failed for path %q: %v", stagingPath, err)
+	}
+
+	// TODO(FUECHR) Add empty dir creation flow.
+	// TODO(FUECHR) Wait for mounter pod running flow.
+	// TODO(FUECHR) Add start gcsfuse flow.
+	klog.Infof("Mounter pod %s/%s is running and staging path %s is mounted", podNamespace, podName, stagingPath)
+
+	klog.Infof("NodeStageVolume succeeded on staging path %q for volume %q", stagingPath, req.GetVolumeId())
 	return &csi.NodeStageVolumeResponse{}, nil
 }
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -650,14 +650,14 @@ func (s *nodeServer) countGcsFuseVolumes(pod *corev1.Pod) (int, error) {
 func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
 	// Validate arguments.
 	if req == nil {
-		return nil, status.Error(codes.InvalidArgument, "Request cannot be nil")
+		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Request cannot be nil")
 	}
 	volumeID := req.GetVolumeId()
 	if len(volumeID) == 0 {
-		return nil, status.Error(codes.InvalidArgument, "Volume ID must be provided")
+		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Volume ID must be provided")
 	}
 	if req.GetVolumeCapability() == nil {
-		return nil, status.Error(codes.InvalidArgument, "Volume capability must be provided")
+		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Volume capability must be provided")
 	}
 	if err := s.driver.validateVolumeCapabilities([]*csi.VolumeCapability{req.GetVolumeCapability()}); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
--- a/pkg/csi_driver/node_test.go
+++ b/pkg/csi_driver/node_test.go
@@ -37,6 +37,9 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	mount "k8s.io/mount-utils"
 )
 
@@ -58,7 +61,23 @@ type nodeServerTestEnv struct {
 func initTestNodeServer(t *testing.T) *nodeServerTestEnv {
 	t.Helper()
 	mounter := mount.NewFakeMounter([]mount.MountPoint{})
-	driver := initTestDriver(t, mounter, clientset.NewFakeClientset())
+	podName := createMounterPodName("test-node", testVolumeID)
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: "test-ns",
+		},
+	}
+	fakeClient := clientset.NewFakeClientset(pod)
+	fakeClient.CreatePV(clientset.FakePVConfig{
+		Name:         "test-pv",
+		VolumeHandle: testVolumeID,
+		ClaimRef: &corev1.ObjectReference{
+			Namespace: "test-ns",
+			Name:      "test-pvc",
+		},
+	})
+	driver := initTestDriver(t, mounter, fakeClient)
 	s, _ := driver.config.StorageServiceManager.SetupService(context.TODO(), nil, "")
 	if _, err := s.CreateBucket(context.Background(), &storage.ServiceBucket{Name: testVolumeID}); err != nil {
 		t.Fatalf("failed to create the fake bucket: %v", err)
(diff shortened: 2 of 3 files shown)


uriel-guzman:
Can you name the PR something else that is not "2nd PR for NodeStageVolume"? Maybe name what the PR is actually introducing to NodeStageVolume?

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,57 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 	}
 	defer s.volumeLocks.Release(stagingPath)
 
+	resp, err := s.executeNodeStageVolume(ctx, req)
```
uriel-guzman, line 701: Can you add a check that the volume is actually using the shared mount feature before entering executeNodeStageVolume and before we even acquire the lock? Otherwise, sidecar architecture is going to go through this logic and it shouldn't.

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,57 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+
+	return resp, err
+}
+
+func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
+	stagingPath := req.GetStagingTargetPath()
+	clientset := s.driver.config.K8sClients
+
+	pv, err := pvFromVolumeID(clientset, req.GetVolumeId())
```
uriel-guzman: You don't need to get the PV to get the namespace in NodeStageVolume, which involves listing.

ControllerPublishVolumeResponse should return the namespace as part of publish_context: https://github.com/container-storage-interface/spec/blob/master/csi.proto#L842

NodeStageVolumeRequest should contain the namesapce as part of publish_context: https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1414

We should pass information from ControllerPublishVolume to NodeStageVolume as much as possible to avoid re-calculating expensive tasks like this one. 

For this PR, you can just assume that the namespace is already contained inside NodeStageVolumeRequest's publish_context. 

Additionally, we could also pass the mounter pod name too like this, although that's optional, because generating the mounter pod name isn't expensive, although it would probably be nice, since it simplifies NodeStageVolume, so I think we should also do that.

Make sure unit tests cover these assumptions.

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,57 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get PV for volume %q: %v", req.GetVolumeId(), err)
+	}
+	if pv.Spec.ClaimRef == nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "PV %q does not have a ClaimRef", pv.Name)
+	}
+	podConfig := &mounterPodConfig{
+		podName:   createMounterPodName(s.driver.config.NodeID, req.GetVolumeId()),
+		namespace: pv.Spec.ClaimRef.Namespace,
+		nodeID:    s.driver.config.NodeID,
```
uriel-guzman: What are you actually using nodeID for? It doesn't seem like you're using it for anything. I don't think you need to create a podConfig either here. You just need mounter pod namespace and mounter pod name?

pkg/csi_driver/shared_mount.go
```diff
@@ -252,3 +252,16 @@ func waitForMounterPodScheduled(clientset clientset.Interface, ctx context.Conte
 		}
 	}
 }
+
+// mustGetMounterPod returns the pod if it exists or returns an error if it doesn't.
+// This function strictly expects the mounter pod to exist.
+func mustGetMounterPod(clientset clientset.Interface, ctx context.Context, config *mounterPodConfig) (*corev1.Pod, error) {
+	pod, err := clientset.GetPod(config.namespace, config.podName)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
uriel-guzman: This logic was copied from Parallelstore, but `getMounterPod` was replaced by `clientset.GetPod(config.namespace, config.podName)`, however, they are not the same. 

`getMounterPod` looks like this:
```
// verifyMounterPodExists returns the mounter pod if it exists, or returns an error.
--
func (nm *nodeMounterManager) getMounterPod(podName string) (*corev1.Pod, error) {
pod, err := nm.k8sClient.GetPod(nm.mounterPodNamespace, podName)
if err != nil {
if errors.IsNotFound(err) {
return nil, nil
}
return nil, err
}
return pod, nil
}
```
It returns nil error when the GET returns NotFound. In your logic, it's returning Internal, which is going to trigger SLOs when the customer accidentally deleted their mounter pod, which we don't want. Also, the pod == nil check in L263 doesn't really make sense anymore, because if the pod isn't found, then L261 will be executed and will return Internal error, so this entire funciton is wrong.

Instead, since you already got rid of the getMounterPod wrapper, maybe just directly call `clientset.GetPod(config.namespace, config.podName)` from executeNodeStageVolume. If it returns NotFound, return FailedPrecondition (it's filtered from our SLOs, same as NotFound), otherwise, return Internal error.

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,57 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+	klog.Infof("Executing NodeStageVolume. Mounter pod: %s/%s, node: %q, volume: %q, staging path: %q", podConfig.namespace, podConfig.podName, podConfig.nodeID, req.GetVolumeId(), stagingPath)
+
+	// Verify mounter pod exists.
+	_, err = mustGetMounterPod(clientset, ctx, podConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if the staging path is already mounted.
+	mounted, err := s.isDirMounted(stagingPath)
```
uriel-guzman: Can we just leave L710 - L718 for the next PR. I think for this PR. You're not really verifying this logic in the existing PR or adding them to any unit tests anyway, let's just focus on getting the "mounter pod exists" logic right first.

uriel-guzman (changes requested):

pkg/csi_driver/node.go
chrisThePattyEater: we akready do this with 	"if !sharedMount(req.GetVolumeContext()) {"

uriel-guzman: Where? I don't see it?

chrisThePattyEater: line 670 on the current pr state - 
`
	// Skip NodeStageVolume for sidecar mounted volumes.
	if !sharedMount(req.GetVolumeContext()) {
	}
`

pkg/csi_driver/node.go
chrisThePattyEater: agreed removed

pkg/csi_driver/node.go
chrisThePattyEater: agreed, when NPV logic is moved over we can just directly use s.driver.config.NodeID there is no need to create the mounterPodConfig

pkg/csi_driver/node.go
chrisThePattyEater: Yes this makes sense, I am now pulling PodName and namespace from publish context

pkg/csi_driver/shared_mount.go
chrisThePattyEater: Ive added this to execute nodestage volume but am concerned about missing out on issues with the mounter pod getting to running state. It seems like an issue our slo should track with the fact that a user might accidently delete the mounter pod being something that happens infrequently

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,59 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+	// Validate staging path and mounter pod information from the request.
+	stagingPath := req.GetStagingTargetPath()
+	if len(stagingPath) == 0 {
+		return nil, status.Error(codes.Internal, "staging target path must be provided")
+	}
+
+	publishContext := req.GetPublishContext()
+	podName := publishContext[VolumeContextKeyPodName]
+	if podName == "" {
+		return nil, status.Error(codes.Internal, "publishContext must contain Pod name")
```
uriel-guzman: mounter pod, not pod

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,59 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+	}
+	podNamespace := publishContext[VolumeContextKeyPodNamespace]
+	if podNamespace == "" {
+		return nil, status.Error(codes.Internal, "publishContext must contain Pod namespace")
```
uriel-guzman: mounter pod, not pod

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,59 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+	clientset := s.driver.config.K8sClients
+
+	// Validate staging path and mounter pod information from the request.
+	stagingPath := req.GetStagingTargetPath()
+	if len(stagingPath) == 0 {
+		return nil, status.Error(codes.Internal, "staging target path must be provided")
+	}
+
+	publishContext := req.GetPublishContext()
+	podName := publishContext[VolumeContextKeyPodName]
```
uriel-guzman: Consider using the "comma ok" idiom to check for the presence of VolumeContextKeyPodName in the publishContext map.

```
podName, ok := publishContext[VolumeContextKeyPodName]
if !ok {
    return nil, status.Error(codes.Internal, "mounter pod name must be provided")
}
if podName == "" {
    return nil, status.Error(codes.Internal, "mounter pod name must not be empty")
}
```
This pattern distinguishes between the key not being in the map (!ok) and the key being present but having an empty string value. Relying on the zero value ("") to check for presence can be ambiguous, as "" could potentially be a valid value.

Same comment below for podNamespace. See go/go-style/decisions#in-band-errors

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,59 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+
+	// Verify mounter pod exists.
+	pod, err := clientset.GetPod(podNamespace, podName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.FailedPrecondition, "mounter pod %s/%s expected to exist but was not found", podNamespace, podName)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get mounter pod %s/%s: %v", podNamespace, podName, err)
+	}
+	if pod == nil {
```
uriel-guzman, line 729: How could pod be nil? This would be unexpected behjavior of the Kubernetes API client to return both a nil Pod object and a nil error. This line should return "Internal" becaues this is unexpected behavior. If the pod was not found, it should have returned NotFound, and be caught in L713?

pkg/csi_driver/node_test.go
```diff
@@ -205,6 +224,118 @@ func TestNodePublishVolume(t *testing.T) {
...
+
+	cases := []struct {
+		name      string
+		req       *csi.NodeStageVolumeRequest
+		mounts    []mount.MountPoint
+		podExists bool
+		expectErr error
+	}{
+		{
+			name: "mounter pod does not exist",
```
uriel-guzman: Make the names of these tests have the form: "<thing being tested> - should <thing expected to happen>". This is common across most of our unit tests in the code base: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/controller_test.go#L213

e.g. this one should be called "mounter pod doesn't exist - should return FailedPrecondition error" or something like that.

uriel-guzman (changes requested):

pkg/csi_driver/node_test.go
```diff
@@ -205,6 +224,118 @@ func TestNodePublishVolume(t *testing.T) {
...
+			req: &csi.NodeStageVolumeRequest{
+				VolumeId:          volID,
+				StagingTargetPath: stagingPath,
+				PublishContext: map[string]string{
+					VolumeContextKeyPodNamespace: podNamespace,
```
uriel-guzman: This says `VolumeContextKeyPodNamespace` but you're actually reading this key from the PublishContext? 

also you seem to be conflating the volumecontext pod name and naemsapce providing by kubelet with the mounter pod's: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/d4bcea2ed7a7d136283b13e63a88efe85e634d35/pkg/cloud_provider/auth/token_manager.go#L29, please make them separate things, since we want to avoid potential conflicts.

pkg/csi_driver/node.go
chrisThePattyEater: Agreed, updated to internal

pkg/csi_driver/node.go
chrisThePattyEater: added

pkg/csi_driver/node.go
chrisThePattyEater: added

pkg/csi_driver/node_test.go
chrisThePattyEater: fixed

pkg/csi_driver/node.go
chrisThePattyEater: added this

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,84 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+		return nil, status.Errorf(codes.Internal, "mounter pod %s/%s expected to exist but was not found", podNamespace, podName)
```
uriel-guzman: I think this error message is misleading, mounter pod %s/%s expected to exist but was not found sounds like the mounter pod wasn't found. Actually, it's the same exact message as L704, the only difference is the error code, which can be super confusing to debug.

Instead, change the message to something like "mounter pod %s/%s can't be nil".

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,84 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+	// TODO(FUECHR) Add empty dir creation flow.
+	// TODO(FUECHR) Wait for mounter pod running flow.
+	// TODO(FUECHR) Add start gcsfuse flow.
+	klog.Infof("Mounter pod %s/%s is running and staging path %s is mounted", podNamespace, podName, stagingPath)
+
+	klog.Infof("NodeStageVolume succeeded on staging path %q for volume %q", stagingPath, req.GetVolumeId())
 	return &csi.NodeStageVolumeResponse{}, nil
 }
+
+func getStagingPathAndMounterPodInfo(req *csi.NodeStageVolumeRequest) (string, string, string, error) {
```
uriel-guzman: Why do we need this? The "helper" function just seems to return and validate fields that are hardly related to each other and doesn't seem to improve readability either. It doesn't seem very flexible / extensible either, and doesn't follow the same style as ControllerPublishVolume?

pkg/csi_driver/utils.go
```diff
@@ -92,6 +92,8 @@ const (
 	FlagFileForDefaultingPath              = "flags-for-defaulting"
 	GCSFuseProfileFlag                     = "profile"
 	LocalSocketAddressArg                  = "experimental-local-socket-address"
+	PublishContextKeyMounterPodName        = "shared-mount/mounter-pod-name"
```
uriel-guzman: I don't think we need "shared-mount" prefix, just call it mounter-pod-name, and mounter-pod-namespace. There's no other feature that uses the concept of mounter pods so this can be simplified.

uriel-guzman (changes requested):

pkg/csi_driver/utils.go
chrisThePattyEater: fixed

pkg/csi_driver/node.go
chrisThePattyEater: removed

pkg/csi_driver/node.go
chrisThePattyEater: fixed

pkg/csi_driver/node_test.go
chrisThePattyEater: added this

pkg/csi_driver/node.go
```diff
@@ -677,5 +677,66 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
...
+}
+
+func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
+	clientset := s.driver.config.K8sClients
+	stagingPath := req.GetStagingTargetPath()
+
+	// Validate staging path and mounter pod information from the request.
+	publishContext := req.GetPublishContext()
+	if publishContext == nil {
+		return nil, status.Error(codes.Internal, "publishContext must be provided")
```
uriel-guzman: Why is nil publish context is Internal, but the validations above for req and volume capabilities nil are InvalidArguments? If req is missing, then publish context is also missing, so why is publish context missing Internal but req missing InvalidArgument?

uriel-guzman (changes requested):

pkg/csi_driver/node.go
uriel-guzman: Also, probably put publishContext validation in the same place as all the other req validations? (in NodeStageVolume, outsode of executeNodeStageVolume).

uriel-guzman (approved):

PR #1450, opened 2026-06-11
[Shared Mount] NSV - Prepare Empty Dir for staging path

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds logic to create an empty dir for the mounter pod in its tmp volume. Additionally we now verify that the volume details are passed in the staging path during NSV validation. Unit tests verify that the empty dir was created if expected and not created if the staging path was already mounted.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -662,10 +662,15 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 	if err := s.driver.validateVolumeCapabilities([]*csi.VolumeCapability{req.GetVolumeCapability()}); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	// Validate the staging path.
 	stagingPath := req.GetStagingTargetPath()
 	if len(stagingPath) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Staging Target Path must be provided")
 	}
+	if matched := util.StagingPathRegexp.FindStringSubmatch(stagingPath); len(matched) < 2 {
+		return nil, status.Errorf(codes.InvalidArgument, "NodeStageVolume Staging Target Path %v does not contain volume information", stagingPath)
+	}
+
 	// Skip NodeStageVolume for sidecar mounted volumes.
 	if !sharedMount(req.GetVolumeContext()) {
 		return &csi.NodeStageVolumeResponse{}, nil
@@ -743,11 +748,16 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	// Make the staging path.
 	klog.Infof("NodeStageVolume attempting mkdir for staging path %q", stagingPath)
 	if err := os.MkdirAll(stagingPath, 0750); err != nil {
-		return nil, status.Errorf(codes.Internal, "mkdir failed for path %q: %v", stagingPath, err)
+		return nil, status.Errorf(codes.Internal, "mkdir failed for staging path %q: %v", stagingPath, err)
 	}
 
-	// TODO(FUECHR) Add empty dir creation flow.
-	// TODO(FUECHR) Wait for mounter pod running flow.
+	// Create mounter pod's emptyDir in the tmp volume.
+	emptyDirPath := util.MounterPodEmptyDirPath(stagingPath, string(pod.UID))
+	klog.Infof("NodeStageVolume attempting mkdir for emptyDir path %q", emptyDirPath)
+	if err := os.MkdirAll(emptyDirPath, 0750); err != nil {
+		return nil, status.Errorf(codes.Internal, "mkdir failed for emptyDir path %q: %v", emptyDirPath, err)
+	}
+	// TODO(FUECHR) Wait for the mounter pod to be ready.
 	// TODO(FUECHR) Add start gcsfuse flow.
 	klog.Infof("Mounter pod %s/%s is running and staging path %s is mounted", podNamespace, podName, stagingPath)
 
--- a/pkg/csi_driver/node_test.go
+++ b/pkg/csi_driver/node_test.go
@@ -40,8 +40,11 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
 	mount "k8s.io/mount-utils"
 )
+
+const testStagingPathStr = "/var/lib/kubelet/plugins/kubernetes.io/csi/gcsfuse.csi.storage.gke.io/test-volume-id/globalmount"
 
 var testVolumeCapability = &csi.VolumeCapability{
 	AccessType: &csi.VolumeCapability_Mount{
@@ -123,6 +126,20 @@ func setupMountTarget(t *testing.T) (string, func()) {
 		t.Fatalf("failed to setup target path: %v", err)
 	}
 	return testTargetPath, func() { defer os.RemoveAll(base) }
+}
+
+func setupTestDir(t *testing.T, path string) (string, func()) {
+	// Create a temporary directory for the test
+	tempDir := t.TempDir()
+	// Append the full path
+	fullPath := filepath.Join(tempDir, path)
+	// Create the directory path (including any parent directories)
+	err := os.MkdirAll(fullPath, 0755)
+	if err != nil {
+		t.Fatalf("failed to create path: %v", err)
+	}
+	klog.Infof("Created test path %q", fullPath)
+	return fullPath, func() { defer os.RemoveAll(tempDir) }
 }
 
 func TestNodePublishVolume(t *testing.T) {
@@ -227,7 +244,8 @@ func TestNodePublishVolume(t *testing.T) {
 func TestExecuteNodeStageVolume(t *testing.T) {
 	t.Parallel()
 
-	stagingPath, cleanup := setupMountTarget(t)
+	// Setup the staging path for the test.
+	stagingPath, cleanup := setupTestDir(t, testStagingPathStr)
 	defer cleanup()
 
 	nodeID := "test-node" // default from initTestDriver
@@ -297,6 +315,10 @@ func TestExecuteNodeStageVolume(t *testing.T) {
 					},
 				}
 				fakeClientSet = clientset.NewFakeClientset(pod)
+
+				// fakeClientset always overrides the pod uid to nil, so we need to set it manually.
+				pod, _ = fakeClientSet.GetPod(podNamespace, podName)
+				pod.UID = testPodUID
 			} else {
 				fakeClientSet = clientset.NewFakeClientset()
 				fakeClientSet.GetPodErr = apierrors.NewNotFound(schema.GroupResource{Resource: "pods"}, podName)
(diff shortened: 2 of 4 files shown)


pkg/util/util.go
```diff
@@ -76,10 +76,12 @@
...
 
 var (
-	volumeIDRegEx            = regexp.MustCompile(`:.*$`)
-	targetPathRegexp         = regexp.MustCompile(`/var/lib/kubelet/pods/(.*)/volumes/kubernetes\.io~csi/(.*)/mount`)
-	emptyReplacementRegexp   = regexp.MustCompile(`kubernetes\.io~csi/(.*)/mount`)
-	gkeIdentityProviderRegex = regexp.MustCompile(
+	volumeIDRegEx                     = regexp.MustCompile(`:.*$`)
+	targetPathRegexp                  = regexp.MustCompile(`/var/lib/kubelet/pods/(.*)/volumes/kubernetes\.io~csi/(.*)/mount`)
+	emptyReplacementRegexp            = regexp.MustCompile(`kubernetes\.io~csi/(.*)/mount`)
+	stagingPathRegexp                 = regexp.MustCompile(`/var/lib/kubelet/plugins/kubernetes\.io/csi/gcsfuse\.csi\.storage\.gke\.io/(.*)/globalmount`)
```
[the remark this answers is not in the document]
chrisThePattyEater: not valid this is for directories not urls

pkg/util/util.go
```diff
@@ -76,10 +76,12 @@
...
+	stagingPathEmptyReplacementRegexp = regexp.MustCompile(`plugins/kubernetes\.io/csi/gcsfuse\.csi\.storage\.gke\.io/(.*)/globalmount`)
```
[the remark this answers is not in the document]
chrisThePattyEater: same as above

pkg/util/util.go
```diff
@@ -76,10 +76,12 @@ const (
...
```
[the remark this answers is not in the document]
chrisThePattyEater: This may be true but i don't think --root-dir is something we support given that NodePublishVolume does this /var/lib/kubelet string. I think we should leave as is to stay in sync with current repo

pkg/util/util.go
```diff
@@ -182,13 +184,36 @@ func ParsePodIDVolumeFromTargetpath(targetPath string) (string, string, error) {
 	return podID, volume, nil
 }
 
+func DetermineGCSFuseTmpVolumePath(stagingPath string, podUID string) string {
+	return stagingPathEmptyReplacementRegexp.ReplaceAllString(stagingPath, fmt.Sprintf("pods/%s/volumes/kubernetes.io~empty-dir/%v/.volumes/$1", podUID, GCSFuseTmpVolumeName))
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: same as above

pkg/csi_driver/node.go
```diff
@@ -662,10 +662,16 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 	if err := s.driver.validateVolumeCapabilities([]*csi.VolumeCapability{req.GetVolumeCapability()}); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	// Validate the staging path.
 	stagingPath := req.GetStagingTargetPath()
 	if len(stagingPath) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Staging Target Path must be provided")
 	}
+	_, err := util.ParseVolumeFromStagingPath(stagingPath)
```
uriel-guzman: Why do we need this? This function is to parse the volume from the staging path. You're not parsing anything here.

pkg/csi_driver/node_test.go
```diff
@@ -345,6 +373,17 @@ func TestExecuteNodeStageVolume(t *testing.T) {
 				if !isAlreadyMounted && !dirExists {
 					t.Errorf("expected MkdirAll to be called, but directory was not created")
 				}
+
+				// Verify emptyDir creation
+				if !isAlreadyMounted {
```
uriel-guzman: The creation of the emptyDir isn't strictly dependent on whether the volume is mounted or not. It literally just depends on whether or not there were any errors before you got to that line of code. This type of conditional makes the testsuite unnecessarily complex and can cause eng toil / technical debt when updating NodeStageVolume in the future, i.e. moving lines of code around shouldn't make the testsuite fail, unless there is a strict dependency that has been changed.

Relax this validation so that it just checks that emptyDir was created. If any testcase causes an error, it will be caught before the validation exists, thus, removing the need for these conditionals.

pkg/csi_driver/node_test.go
```diff
@@ -289,6 +307,10 @@ func TestExecuteNodeStageVolume(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			var fakeClientSet *clientset.FakeClientset
+			// Note: clientset.NewFakeClientset internally overwrites the mock pod with a default template
+			// that doesn't have a UID set, meaning its UID defaults to an empty string.
+			// We set podUID to "" here so our verification matches the path generated by the test environment.
+			podUID := ""
```
uriel-guzman: Can't you just pass podUID to the clientset? Otherwise, how are you validating that the actual podUID from the mounter pod is the one being validated and not just some empty string?

pkg/util/util.go
```diff
@@ -59,8 +59,8 @@ const (
 	OptInHnw                            = "hnw-ksa"
 	EnableCloudProfilerForSidecarConst  = "enable-cloud-profiler-for-sidecar"
 	EnableCloudProfilerConst            = "enable-cloud-profiler"
-	SidecarContainerTmpVolumeName       = "gke-gcsfuse-tmp"
-	SidecarContainerTmpVolumePath       = "/gke-gcsfuse-tmp"
+	GCSFuseTmpVolumeName                = "gke-gcsfuse-tmp"
```
uriel-guzman: I don't think this rename is beneficial. GCSFuseTmpVolumeName can be confused with the --temp-dir: https://docs.cloud.google.com/storage/docs/cloud-storage-fuse/cli-options#temp-dir. I'd suggest just keeping it as was before. Internally, we know that the mounter pod and the sidecar share some configs, so I don't think it's a big deal.

pkg/util/util.go
```diff
@@ -182,13 +184,36 @@ func ParsePodIDVolumeFromTargetpath(targetPath string) (string, string, error) {
...
+func PrepareEmptyDirForStagingPath(stagingPath, podUID string) (string, error) {
+	emptyDirBasePath := DetermineGCSFuseTmpVolumePath(stagingPath, podUID)
+	if err := os.MkdirAll(emptyDirBasePath, 0o750); err != nil {
+		return "", fmt.Errorf("mkdir failed for path %q: %w", emptyDirBasePath, err)
+	}
+
+	return emptyDirBasePath, nil
+}
+
+func ParseVolumeFromStagingPath(stagingPath string) (string, error) {
```
uriel-guzman: I am very confused about why we need this. The only place it's used is in NodeStageVolume, and the volume return isn't even used there, it's ignored, and even if you needed the volume later, you already have it in volumeID via NodeStageVolumeRequest, so what's the purpose?

pkg/util/util.go
```diff
@@ -182,13 +184,36 @@ func ParsePodIDVolumeFromTargetpath(targetPath string) (string, string, error) {
 	return podID, volume, nil
 }
 
+func DetermineGCSFuseTmpVolumePath(stagingPath string, podUID string) string {
```
uriel-guzman: The return statement of this string is wrong, it's going to be: `pods/1234-abcd/volumes/kubernetes.io~empty-dir/gcs-fuse-tmp/.volumes/volumeID`, which was right for Parallelstore, but in the GCSFuse design doc, it says that it should be: `/var/lib/kubelet/pods/<podUID>/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp`, because there's only 1 volume per mounter pod, so the .volumes directory isn't required.

I think it'd just be simpler to make a function that returns `/var/lib/kubelet/pods/<podUID>/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp`, instead of all this regex parsing?

uriel-guzman (changes requested):

pkg/csi_driver/node.go
chrisThePattyEater: this was to use to validate the staging path contains volume id. I have moved it to NSV validations and removed this function

pkg/csi_driver/node_test.go
chrisThePattyEater: I have modified the tests so we always expect the staging path and empty dir to exist by the end of the test in the happy path. If its already mounted then it should exist and if it wasn't mounted it should have been created. This simplifies the logic a lot. Great suggestion!

pkg/csi_driver/node_test.go
chrisThePattyEater: I found a workaround just manually set it after the pod has been created by clientset

pkg/util/util.go
chrisThePattyEater: I set it back

pkg/util/util.go
chrisThePattyEater: this was to use to validate the staging path contains volume id. I have moved it to NSV validations and removed this function

pkg/util/util.go
chrisThePattyEater: pods/1234-abcd/volumes/kubernetes.io~empty-dir/gcs-fuse-tmp/.volumes/volumeID is wrong
I have replaced this function with 

// MounterPodEmptyDirPath constructs the path for the empty directory of a given pod.
func MounterPodEmptyDirPath(stagingPath string, podUID string) string {
	return stagingPathEmptyReplacementRegexp.ReplaceAllString(stagingPath, fmt.Sprintf("pods/%s/volumes/kubernetes.io~empty-dir/%v", podUID, SidecarContainerTmpVolumeName))
}

this should correctly create /var/lib/kubelet/pods/<podUID>/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp

ex from test - /tmp/TestExecuteNodeStageVolume2774958846/001/var/lib/kubelet/pods/test-pod-uid/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp

We need to keep the regex to allow easy substitution of the rooth path. If we don't use regex the unit test will attempt to create a dir in var/lib/kubelet for which it does not have access. In realitiy the regex will allow us to to create in a temp test dir

pkg/csi_driver/node.go
```diff
@@ -662,10 +662,15 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 	if err := s.driver.validateVolumeCapabilities([]*csi.VolumeCapability{req.GetVolumeCapability()}); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	// Validate the staging path.
 	stagingPath := req.GetStagingTargetPath()
 	if len(stagingPath) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Staging Target Path must be provided")
 	}
+	if matched := util.StagingPathRegexp.FindStringSubmatch(stagingPath); len(matched) < 2 {
```
uriel-guzman, line 670: Why do we care about the format of the staging path? If Kubelet one day decided to change the staging path for some reason, would we care? If not, then I think this validation isn't required.

uriel-guzman: Also, it shouldn't be InvalidArgument. That would be Internal, because the customer doesn't input the staging path, it's a system component (Kubelet).

uriel-guzman:
I don't think this PR is required. The emptyDir path `/var/lib/kubelet/pods/podUID/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp` already exists. It's created by Kubelet in the hostpath to mount the emptyDir. In Parallelstore, it was required because it still used the `.volumes/volume` convention, but the GCSFuse shared mount remove this redundant directory since there's only 1 volume per mounter pod. So, I think we don't need this PR.

PR #1453, opened 2026-06-12
Implement ControllerPublishVolume PodTemplate logic

uriel-guzman:
**What type of PR is this?**
/kind feature

Implement the logic for ControllerPublishVolume so that it can find the PodTemplate from the PVC, and then allow the mounter pod to be overridden with supported values from there. In this PR, we start simply with `serviceAccountName`. More will be added later.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/cloud_provider/clientset/fake.go
+++ b/pkg/cloud_provider/clientset/fake.go
@@ -55,9 +55,10 @@ type FakePVConfig struct {
 }
 
 type FakePVCConfig struct {
-	Name       string
-	VolumeName string
-	Namespace  string
+	Name        string
+	VolumeName  string
+	Namespace   string
+	Annotations map[string]string
 }
 
 type FakeSCConfig struct {
@@ -66,24 +67,32 @@ type FakeSCConfig struct {
 	MountOptions []string
 }
 
+type FakePodTemplateConfig struct {
+	Name      string
+	Namespace string
+}
+
 type FakeClientset struct {
-	Client    kubernetes.Interface
-	fakePod   *corev1.Pod
-	fakeNode  *corev1.Node
-	fakePVs   map[string]*corev1.PersistentVolume
-	fakePVCs  map[string]*corev1.PersistentVolumeClaim
-	fakeSCs   map[string]*storagev1.StorageClass
-	ListPVErr error
-	GetPodErr error
+	Client            kubernetes.Interface
+	fakePod           *corev1.Pod
+	fakeNode          *corev1.Node
+	fakePVs           map[string]*corev1.PersistentVolume
+	fakePVCs          map[string]*corev1.PersistentVolumeClaim
+	fakeSCs           map[string]*storagev1.StorageClass
+	fakePodTemplates  map[string]*corev1.PodTemplate
+	ListPVErr         error
+	GetPodErr         error
+	GetPodTemplateErr error
 }
 
 func NewFakeClientset(objects ...runtime.Object) *FakeClientset {
 	fakeK8sClient := fake.NewSimpleClientset(objects...)
 	fakeClientSet := &FakeClientset{
-		Client:   fakeK8sClient,
-		fakePVs:  make(map[string]*corev1.PersistentVolume),
-		fakePVCs: make(map[string]*corev1.PersistentVolumeClaim),
-		fakeSCs:  make(map[string]*storagev1.StorageClass),
+		Client:           fakeK8sClient,
+		fakePVs:          make(map[string]*corev1.PersistentVolume),
+		fakePVCs:         make(map[string]*corev1.PersistentVolumeClaim),
+		fakeSCs:          make(map[string]*storagev1.StorageClass),
+		fakePodTemplates: make(map[string]*corev1.PodTemplate),
 	}
 	// Default setting for most unit tests is pod doesn't use host network & workload identity is enabled on the node
 	fakeClientSet.CreatePod(FakePodConfig{HostNetworkEnabled: false})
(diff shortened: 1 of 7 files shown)


pkg/cloud_provider/clientset/fake.go
```diff
@@ -280,6 +304,17 @@ func (c *FakeClientset) GetSC(name string) (*storagev1.StorageClass, error) {
 	return c.fakeSCs[""], nil
 }
 
+func (c *FakeClientset) GetPodTemplate(namespace, name string) (*corev1.PodTemplate, error) {
```
chrisThePattyEater, line 307: do we need ns?

pkg/csi_driver/controller.go
chrisThePattyEater, line 171: For my own learning/understanding - I know this was not introduced in this PR but should the error code be Internal if pv is not bound to pvc. Wouldn't this trigger our slo as if it were a bug on our end?

pkg/csi_driver/shared_mount.go
```diff
@@ -197,6 +199,9 @@ func createMounterPodSpec(config *mounterPodConfig) *corev1.Pod {
 			},
 		},
 	}
+	if config.serviceAccountName != "" {
```
chrisThePattyEater, line 202: Why do we need this, isn't the default value already "". IIUC it gets set to "default" by a controller https://screenshot.googleplex.com/9XhLKqbiFyzaXxr

pkg/csi_driver/controller.go
```diff
@@ -173,12 +177,29 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
...
+		if apierrors.IsNotFound(err) {
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	podTemplate, err := mounterPodTemplate(clientset, pvc)
+	if err != nil {
+		return nil, err
+	}
+	if podTemplate == nil {
```
chrisThePattyEater, line 192: I don't think both the pod template and err can be nil? It seems to not be a case we handle for sc https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/57c552f20afd0f5ba006dedf4497597187e96bc6/pkg/webhook/client.go#L91-L92.

pkg/csi_driver/controller_test.go
```diff
@@ -314,12 +318,117 @@ func TestControllerPublishVolume(t *testing.T) {
...
+			name: "sharedMount true - missing mounter pod template annotation - should return error",
+			req: &csi.ControllerPublishVolumeRequest{
+				VolumeId:         testVolumeID,
+				NodeId:           testNodeID,
+				VolumeCapability: testVolumeCapability,
+				VolumeContext: map[string]string{
+					"sharedMount": "true",
+				},
+			},
+			setupFake: func() *clientset.FakeClientset {
```
chrisThePattyEater, line 381: these set up fakes seem repetetive aside from the annotations can we just make a helper function to shorten the test file?

chrisThePattyEater: Maybe 	
defaultPVCAnnotations := map[string]string{
		"gke-gcsfuse/mounter-pod-template": testMounterPodTemplate,
	}

	setupBaseFake := func(pvcAnnotations map[string]string, existingPod *corev1.Pod) *clientset.FakeClientset {
		var objs []runtime.Object
		if existingPod != nil {
			objs = append(objs, existingPod)
		}
		fc := clientset.NewFakeClientset(objs...)
		fc.CreatePV(clientset.FakePVConfig{Name: testPV, VolumeHandle: testVolumeID, ClaimRef: &corev1.ObjectReference{
			Name:      testPVC,
			Namespace: testNamespace,
		}})
		fc.CreatePVC(clientset.FakePVCConfig{
			Name:        testPVC,
			Namespace:   testNamespace,
			Annotations: pvcAnnotations,
		})
		fc.CreatePodTemplate(clientset.FakePodTemplateConfig{
			Name:      testMounterPodTemplate,
			Namespace: testNamespace,
		})
		return fc
	}

and then do 
setupFake: func() *clientset.FakeClientset {
return setupBaseFake(defaultPVCAnnotations, nil)
}
or 
setupFake: func() *clientset.FakeClientset {
return setupBaseFake(defaultPVCAnnotations, existingPod)
}

you get the point

pkg/cloud_provider/clientset/fake.go
uriel-guzman: Yes, it's an interface.

pkg/csi_driver/controller.go
uriel-guzman: CPV depends on Pod being scheduled, which depends on PV being bound to PVC. PV should be bound to PVC by the PV controller, which is a system component. Since a system controller is the one who should have bound the PV and the PVC together, I think it makes sense to return Internal, since the system is the one who's failing

pkg/csi_driver/controller.go
uriel-guzman: You also did the same thing for the mounter pod: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L729. I agree it shouldn't happen, but it doesn't really hurt to put it either, because if it never happens, the line of code won't be executed, and if it does, it returns Internal, which means it's unexpected, so I think it's fine either way.

pkg/csi_driver/controller_test.go
uriel-guzman: Done.

chrisThePattyEater (approved):

PR #1457, opened 2026-06-16
[Shared Mount] NSV - Wait for Mounter Pod GRPC Server

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature
> /kind flake

This feature continues the implementation of NSV by checking to see if the Mounter Pod GRPC Server is ready by checking the existence of the socket file over which it is hosted. We first check if the mounter pod is in a running state because if it is not then even though the socket file may exist it may be stale or the socket was created and then the mounter pod crashed.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/gcs_fuse_driver.go
+++ b/pkg/csi_driver/gcs_fuse_driver.go
@@ -21,12 +21,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
 
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/auth"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/storage"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/metrics"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/profiles"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -75,6 +77,7 @@ type GCSDriverConfig struct {
 	FeatureOptions                 *GCSDriverFeatureOptions
 	AssumeGoodSidecarVersion       bool
 	UniverseDomain                 string
+	EmptyDirBasePath               func(podUID string) string
 }
 
 type GCSDriver struct {
@@ -103,6 +106,11 @@ func NewGCSDriver(config *GCSDriverConfig, recorder record.EventRecorder) (*GCSD
 	}
 	if !config.RunController && !config.RunNode {
 		return nil, errors.New("must run at least one controller or node service")
+	}
+	if config.EmptyDirBasePath == nil {
+		config.EmptyDirBasePath = func(podUID string) string {
+			return filepath.Join(util.KubeletDir, "pods", podUID, "volumes", "kubernetes.io~empty-dir", util.SidecarContainerTmpVolumeName)
+		}
 	}
 
 	driver := &GCSDriver{
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -726,8 +726,11 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		return nil, status.Errorf(codes.Internal, "mkdir failed for path %q: %v", stagingPath, err)
 	}
 
-	// TODO(FUECHR) Add empty dir creation flow.
-	// TODO(FUECHR) Wait for mounter pod running flow.
+	// Wait for the mounter pod grpc server to be ready.
+	if err := waitForMounterServer(ctx, clientset, podNamespace, podName, string(pod.UID), s.driver.config.EmptyDirBasePath); err != nil {
+		return nil, err
+	}
+
 	// TODO(FUECHR) Add start gcsfuse flow.
 	klog.Infof("Mounter pod %s/%s is running and staging path %s is mounted", podNamespace, podName, stagingPath)
 
--- a/pkg/csi_driver/shared_mount.go
+++ b/pkg/csi_driver/shared_mount.go
@@ -21,6 +21,8 @@ import (
 	"crypto/sha1"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"time"
 
 	"context"
@@ -42,6 +44,7 @@ const (
 	mounterPodNamePrefix    = "gcsfusecsi-mount"
 	mounterPodPriorityClass = "gcsfusecsi-mount-priority"
 	mounterPodMountDir      = "mount-dir"
+	mounterSocketFile       = "mounter.sock"
 )
 
 var (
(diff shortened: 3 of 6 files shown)


pkg/csi_driver/shared_mount.go
```diff
@@ -320,3 +322,69 @@ func deleteMounterPod(ctx context.Context, clientset clientset.Interface, podNam
...
+		// Get the mounter pod status.
+		if pod, err = clientset.GetPod(mounterPodNamespace, mounterPodName); err != nil {
+			// Mounter Pod should exist by this point.
+			return false, status.Errorf(codes.Internal, "failed to get mounter pod %s/%s: %v", mounterPodNamespace, mounterPodName, err)
+		}
+
+		// If the pod is not pending or running, it's in an unexpected state.
+		if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodPending {
+			return false, status.Errorf(codes.Internal, "mounter pod %s/%s found with an unexpected status: %s", mounterPodNamespace, mounterPodName, pod.Status.Phase)
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: this will get caught by NPV

pkg/csi_driver/node.go
```diff
@@ -47,6 +47,7 @@ const (
 	GCSFuseKernelParamsFilePollInterval = time.Second * 5
 	FuseMountType                       = "fuse"
 	maxGCSFuseVolumesForMetrics         = 10
+	mounterSocketFile                   = "mounter.sock"
```
uriel-guzman: Put this in shared_mount.go, where every other mounter pod specific const goes: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/shared_mount.go#L43

pkg/csi_driver/shared_mount.go
```diff
@@ -320,3 +322,70 @@ func deleteMounterPod(ctx context.Context, clientset clientset.Interface, podNam
...
+// waitForMounterServer waits for the mounter pod's gRPC server to be ready.
+func waitForMounterServer(ctx context.Context, clientset clientset.Interface, mounterPodNamespace, mounterPodName, podUID, kubeletDir string) error {
+	// Construct the mounter socket file path.
+	mounterSocketFilePath := filepath.Join(kubeletDir, "pods", podUID, "volumes", "kubernetes.io~empty-dir", util.SidecarContainerTmpVolumeName, mounterSocketFile)
+	var pod *corev1.Pod
+	var err error
+
+	checkIfReady := func() (bool, error) {
+		// Get the mounter pod status.
+		if pod, err = clientset.GetPod(mounterPodNamespace, mounterPodName); err != nil {
```
uriel-guzman, line 401: If the mounter pod doesn't exist, this is inconsistent with: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L725, which returns "FailedPrecondition", while this returns internal.

pkg/csi_driver/shared_mount.go
```diff
@@ -320,3 +322,70 @@ func deleteMounterPod(ctx context.Context, clientset clientset.Interface, podNam
...
+			// Mounter Pod should exist by this point.
+			return false, status.Errorf(codes.Internal, "failed to get mounter pod %s/%s: %v", mounterPodNamespace, mounterPodName, err)
+		}
+
+		// If the pod is not pending or running, it's in an unexpected state.
+		if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodPending {
+			return false, status.Errorf(codes.Internal, "mounter pod %s/%s found with an unexpected status: %s", mounterPodNamespace, mounterPodName, pod.Status.Phase)
+		}
+
+		// If the pod is pending, it's still being scheduled or images are being pulled.
```
uriel-guzman: This comment is misleading. The mounter pod can't be waiting to be scheduled, because NodeStageVolume only executes when ControllerPublishVolume succeeds, which verified the pod being scheduled here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/controller.go#L219

pkg/csi_driver/shared_mount.go
```diff
@@ -320,3 +322,70 @@ func deleteMounterPod(ctx context.Context, clientset clientset.Interface, podNam
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			if ready, err := checkIfReady(); err != nil || ready {
+				return err
+			}
+		case <-ctx.Done():
+			code := status.FromContextError(ctx.Err()).Code()
```
uriel-guzman: To further improve the error code detection, can we distinguish between DeadlineExceeded and Canceled? Just copy this logic: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/shared_mount.go#L252

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -75,6 +76,7 @@ type GCSDriverConfig struct {
 	FeatureOptions                 *GCSDriverFeatureOptions
 	AssumeGoodSidecarVersion       bool
 	UniverseDomain                 string
+	KubeletDir                     string
```
uriel-guzman: KubeletDir is a critical path in Kubernetes. It could be misleading to future developers to think they can / should use this instead of using `util.KubeletDir`. I think we should reduce the scope and responsability of this field as much as possible, since its only purpose is to change the value of the emptyDir for testing.

You can create a field method of type `func(string) string` called `emptyDirPath`.

In node.go, you can define a function:

```
func emptyDirPathBase(podUID string) string {
  return fmt.Sprintf(/var/lib/kubelet/pods/%s/volumes/kubernetes.io~empty-dir/%s, podUID, SidecarContainerTmpVolumeName)
}
```

And in node_test.go, you can define:
```
func emptyDirPathTest(podUID string) string {
  return fmt.Sprintf(some/fake/tmp/dir/var/lib/kubelet/pods/%s/volumes/kubernetes.io~empty-dir/%s, podUID, SidecarContainerTmpVolumeName)
}
```

Then, when you initialize the CSI Driver, you can set `config.emptyDirPath = emptyDirPathBase`, and in the node_test.go, you can set `config.emptyDirPath = emptyDirPathTest`.

pkg/csi_driver/shared_mount.go
```diff
@@ -320,3 +322,70 @@ func deleteMounterPod(ctx context.Context, clientset clientset.Interface, podNam
...
+			klog.Infof("Mounter pod %s/%s is still Pending (scheduling or pulling images). Retrying...", mounterPodNamespace, mounterPodName)
+			return false, nil
+		}
+
+		// Mounter pod is running, check if the socket file is available.
+		// TODO(FUECHR): Investigate symlink traversal vulnerabilities for os.Stat vs os.Lstat.
+		if _, err = os.Stat(mounterSocketFilePath); err == nil {
+			klog.Infof("Mounter pod %s/%s is running and socket file %q is available.", mounterPodNamespace, mounterPodName, mounterSocketFilePath)
+			return true, nil
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: Added a todo

pkg/csi_driver/shared_mount.go
uriel-guzman: This is still wrong... you need to check the NotFound error code, if it's that, then return FailedPrecondition, if not, return Internal, otherwise it's always going to return FailedPrecondition, even if it's an internal error.

uriel-guzman (approved):

PR #1463, opened 2026-06-17
[Shared Mount] NSV - Add start gcsfuse

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds the logic to send a grpc request to the mounter pods grpc server and start the GCSFuse process.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -24,6 +24,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 	"time"
 
@@ -246,6 +247,10 @@ func main() {
 		},
 		SharedMountOptions: &driver.SharedMountOptions{
 			MounterPodImage: *mounterPodImage,
+			FuseSocketDir:   *fuseSocketDir,
+			EmptyDirBasePath: func(podUID string) string {
+				return filepath.Join(util.KubeletDir, "pods", podUID, "volumes", "kubernetes.io~empty-dir", util.SidecarContainerTmpVolumeName)
+			},
 		},
 	}
 
--- a/pkg/csi_driver/gcs_fuse_driver.go
+++ b/pkg/csi_driver/gcs_fuse_driver.go
@@ -21,14 +21,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"path/filepath"
 
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/auth"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/storage"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/metrics"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/profiles"
-	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -54,6 +52,9 @@ type GoMemLimitOptions struct {
 
 type SharedMountOptions struct {
 	MounterPodImage string
+	FuseSocketDir   string
+	// Needed to override the mounter pods emptydir base path, otherwise tests will try to write to var/lib/kubelet which it won't have access to.
+	EmptyDirBasePath func(podUID string) string
 }
 
 type GCSDriverFeatureOptions struct {
@@ -82,7 +83,6 @@ type GCSDriverConfig struct {
 	FeatureOptions                 *GCSDriverFeatureOptions
 	AssumeGoodSidecarVersion       bool
 	UniverseDomain                 string
-	EmptyDirBasePath               func(podUID string) string
 }
 
 type GCSDriver struct {
@@ -111,11 +111,6 @@ func NewGCSDriver(config *GCSDriverConfig, recorder record.EventRecorder) (*GCSD
 	}
 	if !config.RunController && !config.RunNode {
 		return nil, errors.New("must run at least one controller or node service")
-	}
-	if config.EmptyDirBasePath == nil {
-		config.EmptyDirBasePath = func(podUID string) string {
-			return filepath.Join(util.KubeletDir, "pods", podUID, "volumes", "kubernetes.io~empty-dir", util.SidecarContainerTmpVolumeName)
-		}
 	}
 
 	driver := &GCSDriver{
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -20,6 +20,7 @@ package driver
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -32,7 +33,9 @@ import (
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 	"golang.org/x/net/context"
 	"golang.org/x/time/rate"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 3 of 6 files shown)


pkg/csi_driver/node.go
```diff
@@ -731,7 +731,18 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
 	}
 
-	// TODO(FUECHR) Add start gcsfuse flow.
+	podUID := string(pod.UID)
+	emptyDirPath := s.driver.config.EmptyDirBasePath(podUID)
+
+	// Send GRPC to mounter pod to start GCSFuse.
+	if err := mountToNode(ctx, emptyDirPath, podUID, stagingPath, req.GetVolumeId()); err != nil {
+		// Surface GCSFuse error to the user, if it exists.
+		if gcsfuseErr := maybeReadGCSFuseErr(emptyDirPath); gcsfuseErr != nil {
```
uriel-guzman: I know this is copied from Parallelstore, but we don't really need this. If the Mount gRPC call fails, it's immediately going to return an error, which is returned by L738 anyway. If Mount succeeds, and GCSFuse returns an error later (i.e. mount succeeds but GCSFuse crashes), it won't be caught here anyway, it would be caught later in NodePublishVolume, so this check is redundant and doesn't actually cover any scenario. Let's just remove it.

pkg/csi_driver/node.go
```diff
@@ -731,7 +731,18 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+			return nil, status.Error(codes.Internal, gcsfuseErr.Error())
+		}
+		return nil, status.Error(codes.Internal, err.Error())
```
uriel-guzman: This is returning any Mount error as Internal, which will alert us. I don't think this is correct, because if Mount fails because of PermissionDenied (e.g. KSA doens't have bucket access), it shouldn't alert us. Instead, just return whatever Mount error returns, and we need to ensure we return Mount error codes correctly. Also, Parallelstore just returns the error correctly, so I'm not sure why it was copied wrong here.

pkg/csi_driver/shared_mount.go
```diff
@@ -49,6 +50,7 @@ const (
 
 var (
 	mounterPodPollInterval = 1 * time.Second
+	symlinkBaseDir         = filepath.Join(util.KubeletDir, "mount-socket")
```
uriel-guzman: This is wrong and dangerous. We should definetely not be creating, deleting, and executing arbitrary files in `/var/lib/kubelet`, which is a critical hostPath shared by many things in the node. Something goes wrong here can be a node-wide OMG even for non-GCSFuse components. I'm not sure how the Kubelet Dir made its way here.

In Parallelstore, we created it in the Parallelstore CSI Driver's emptyDir: `gke-parallelstore-tmp`. This ensures the file stays within the scope of the driver only.

In GCSFuse, there already seems to be an emptyDir dedicated to storing socket for the CSI Driver, so we should just use that: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L259

pkg/csi_driver/shared_mount.go
```diff
@@ -461,3 +463,66 @@ func waitForMounterServer(ctx context.Context, clientset clientset.Interface, mo
...
+	// Remove any stale symlink if it exists from a previous attempt.
+	if err := os.Remove(symlink); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to remove stale symlink %q: %w", symlink, err)
+	}
+
+	// Verify that the socket file is not a symbolic link to prevent symlink traversal attacks.
+	// TODO(FUECHR): Further investigate symlink traversal vulnerabilities.
+	if err := util.CheckNotSymlink(socketFile); err != nil {
+		return err
+	}
```
uriel-guzman: Why do we need this in this PR?

uriel-guzman (changes requested):

cmd/csi_driver/main.go
```diff
@@ -291,6 +291,7 @@ func main() {
 		FeatureOptions:                 featureOptions,
 		AssumeGoodSidecarVersion:       *assumeGoodSidecarVersion,
 		UniverseDomain:                 *universeDomain,
+		FuseSocketDir:                  *fuseSocketDir,
```
uriel-guzman: Scope your feature-specific changes to the feature-specific config: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/cmd/csi_driver/main.go#L214.

You can check my unmerged PR and copy what I did. Ideally, we should put all the shared-mount-specific stuff in here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1466/changes, see the `SharedMountOptions` change.

chrisThePattyEater: but this socket dir is where sidecar sockets also go

pkg/csi_driver/node.go
```diff
@@ -731,9 +733,61 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
 }
+
+// mountToNode connects to the mounter server, at which point it initializes the GCSFuse process.
+func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath, volumeID string) error {
+	emptyDirBasePath := s.driver.config.EmptyDirBasePath(podUID)
+	socketFile := filepath.Join(emptyDirBasePath, mounterSocketFile)
+
+	// Create a symlink to bypass the 108-character limit for Unix domain sockets
+	// when dialing the connection from the Node Server.
+	symlink := filepath.Join(s.driver.config.FuseSocketDir, podUID)
```
uriel-guzman: This is just going to create a file that looks like `/fuse-socket-dir/<podUID>`. At least, give it back the original `mount-socket` dir you had before this change so it looks like `/fuse-socket-dir/mount-socket/<podUID>`, so that the mounter pod-specific symlinks are contained in a directory isolated from the rest of the stuff that already exists in `/fuse-socket-dir`, minimizing the risk of messing with other files from the CSI Driver and vice-versa.

pkg/csi_driver/node.go
```diff
@@ -731,9 +733,61 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+	if err := os.MkdirAll(filepath.Dir(symlink), 0750); err != nil {
```
uriel-guzman, line 768: IIUC, this is attempting to create the directory for the path `/fuse-socket-dir/<podUID>`, which would be `/fuse-socket-dir`, but that already exists because of it's an emptyDir? So why need to create it?

pkg/csi_driver/node.go
```diff
@@ -731,9 +733,61 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
 }
+
+// mountToNode connects to the mounter server, at which point it initializes the GCSFuse process.
+func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath, volumeID string) error {
+	emptyDirBasePath := s.driver.config.EmptyDirBasePath(podUID)
+	socketFile := filepath.Join(emptyDirBasePath, mounterPodSocketFile)
+
+	// Create a symlink to bypass the 108-character limit for Unix domain sockets
+	// when dialing the connection from the Node Server.
+	symlink := filepath.Join(s.driver.config.FeatureOptions.SharedMountOptions.FuseSocketDir, mountSocketDir, podUID)
```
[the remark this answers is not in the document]
chrisThePattyEater: this poduid is for the mounter pod it should be specific to the volume

cmd/csi_driver/main.go
chrisThePattyEater: Moved it to the shared mount config

pkg/csi_driver/node_test.go
```diff
@@ -1001,6 +1003,56 @@ func TestNodeStageVolume(t *testing.T) {
...
+			emptyDirBasePath, err := util.PrepareEmptyDir(stagingPath, true)
+			if err != nil {
+				t.Fatalf("failed to prepare emptyDir path: %v", err)
+			}
+
+			// Create a dummy socket file for the client to connect to
+			socketFile := filepath.Join(emptyDirBasePath, mounterSocketFile)
+			if file, err := os.Create(socketFile); err == nil {
+				file.Close()
+			}
```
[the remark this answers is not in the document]
chrisThePattyEater: this will be fixed when the real server is added

pkg/csi_driver/node.go
chrisThePattyEater: added mount-socket

pkg/csi_driver/node.go
chrisThePattyEater: it now creates mount-socket if it doesnt already exist

cmd/csi_driver/main.go
```diff
@@ -246,6 +246,7 @@ func main() {
 		},
 		SharedMountOptions: &driver.SharedMountOptions{
 			MounterPodImage: *mounterPodImage,
+			FuseSocketDir:   *fuseSocketDir,
```
uriel-guzman, line 250: Can you also put the emptyDirBasePath function here, as we discussed via chat?

pkg/csi_driver/node.go
```diff
@@ -731,9 +734,64 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+
+// mountToNode connects to the mounter server, at which point it initializes the GCSFuse process.
+func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath, volumeID string) error {
+	emptyDirBasePath := s.driver.config.EmptyDirBasePath(podUID)
+	socketFile := filepath.Join(emptyDirBasePath, mounterPodSocketFile)
+
+	// Create a symlink to bypass the 108-character limit for Unix domain sockets
+	// when dialing the connection from the Node Server.
+	if s.driver.config.FeatureOptions == nil || s.driver.config.FeatureOptions.SharedMountOptions == nil || s.driver.config.FeatureOptions.SharedMountOptions.FuseSocketDir == "" {
+		return fmt.Errorf("FuseSocketDir is not configured")
```
uriel-guzman: This error returns no error code, and L741 doesn't return anything either, so this is going to be "Unknown". Please make this an internal error in exactly this line. We should be notified via SLO if this is missing. Also, change message to "fuse socket dir is not configured" to avoid returning implementation details to the user.

pkg/csi_driver/node.go
```diff
@@ -731,9 +734,64 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+	}
+	symlink := filepath.Join(s.driver.config.FeatureOptions.SharedMountOptions.FuseSocketDir, mountSocketDir, podUID)
+	if err := os.MkdirAll(filepath.Dir(symlink), 0750); err != nil {
+		return fmt.Errorf("failed to create dir for symlink %q: %w", symlink, err)
```
uriel-guzman: Again, no error code. This should be Internal.

pkg/csi_driver/shared_mount.go
```diff
@@ -46,6 +46,7 @@ const (
 	mounterPodMountDir            = "mount-dir"
 	mounterPodSocketFile          = "mounter.sock"
 	mounterPodManagedImageKeyword = "managed"
+	mountSocketDir                = "mount-socket"
```
uriel-guzman: mounterPodSocketDir, to keep consistency with every other const above.

pkg/csi_driver/node.go
```diff
@@ -731,9 +734,64 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+	}
+
+	if err := os.Remove(symlink); err != nil && !os.IsNotExist(err) {
+		klog.Errorf("failed to remove stale symlink %q: %v", symlink, err)
+	}
+
+	if err := os.Symlink(socketFile, symlink); err != nil {
+		return fmt.Errorf("failed to create symlink to %q: %w", socketFile, err)
```
uriel-guzman: Return internal error.

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -53,7 +53,9 @@ type GoMemLimitOptions struct {
 }
 
 type SharedMountOptions struct {
-	MounterPodImage string
+	MounterPodImage  string
+	FuseSocketDir    string
+	EmptyDirBasePath func(podUID string) string
```
uriel-guzman, line 57: Add comment explaining why this needs to be here. Just say that it's for overriding the directory with something other than /var/lib/kubelet, otherwise you get a an error during unit test execution.

pkg/csi_driver/gcs_fuse_driver.go
```diff
@@ -112,8 +113,14 @@ func NewGCSDriver(config *GCSDriverConfig, recorder record.EventRecorder) (*GCSD
 	if !config.RunController && !config.RunNode {
 		return nil, errors.New("must run at least one controller or node service")
 	}
-	if config.EmptyDirBasePath == nil {
-		config.EmptyDirBasePath = func(podUID string) string {
+	if config.FeatureOptions == nil {
```
uriel-guzman: I think the initialization here doesn't align with how things are initialized for the other values of SharedMountOptions. Can't you just put this in main.go like this?
```
SharedMountOptions: &driver.SharedMountOptions{
			MounterPodImage: *mounterPodImage,
			FuseSocketDir:   *fuseSocketDir,
            EmptyDirBasePath: func(podUID string) string {
			return filepath.Join(util.KubeletDir, "pods", podUID, "volumes", "kubernetes.io~empty-dir", util.SidecarContainerTmpVolumeName)
```
Then, you can avoid all that nil checking stuff you have here and make the code cleaner.

pkg/csi_driver/node.go
```diff
@@ -727,13 +730,69 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
 	klog.Infof("Mounter pod %s/%s is running and staging path %s is mounted", podNamespace, podName, stagingPath)
 
 	klog.Infof("NodeStageVolume succeeded on staging path %q for volume %q", stagingPath, req.GetVolumeId())
 	return &csi.NodeStageVolumeResponse{}, nil
 }
+
+// mountToNode connects to the mounter server, at which point it initializes the GCSFuse process.
+func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath, volumeID string) error {
+	if s.driver.config.FeatureOptions == nil || s.driver.config.FeatureOptions.SharedMountOptions == nil || s.driver.config.FeatureOptions.SharedMountOptions.FuseSocketDir == "" || s.driver.config.FeatureOptions.SharedMountOptions.EmptyDirBasePath == nil {
+		return status.Errorf(codes.Internal, "shared mount options are not fully configured")
```
uriel-guzman, line 753: This seems like it just bundles a bunch of unrelated things and returns an unhelpful and mysterious error message.

Instead, you should validate these things until you need them. For example, `FuseSocketDir` can be validated in L761 right before you're going to use it and return something helpful like "fuse socket dir can't be empty". Then, do the same with `EmptyDirBasePath` in L756 and return something useful like "empty dir base path method can't be nil". This way, if we ever get alerted, we know exactly which field is missing / nil / not initialized.

uriel-guzman (approved):

PR #1474, opened 2026-06-22
[Shared Node Mount] Mounter Pod grpc server skeleton

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds the following:
*Adds the mounter pods proto buff, defining the mounter pods server requests and response
* Adds the kublet plugs dir "kubelet-plugins-dir" to the node, allowing it to create the staging path during node stage volume
* Corrects the mounter pods expected staging path to specify the one we added "kubelet-plugins-dir"
* Registers the server in server.go
* Starts the server in sidecarmounter/main.go

Fixes #

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -39,6 +39,7 @@ import (
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	mount "k8s.io/mount-utils"
 
@@ -52,6 +53,18 @@ const (
 	GCSFuseKernelParamsFilePollInterval = time.Second * 5
 	FuseMountType                       = "fuse"
 	maxGCSFuseVolumesForMetrics         = 10
+	unmountRetryInterval                = 100 * time.Millisecond
+	unmountRetryBackoffFactor           = 2.0
+	unmountRetryJitter                  = 0.1
+	// Standard unmount has no built-in timeout, so we use a short timeout to fail fast.
+	standardUnmountRetryTimeout = 2 * time.Second
+	standardUnmountRetrySteps   = 5
+	// Force unmount attempts include an initial non-force unmount with a
+	// 5 second timeout (as defined in vendor/k8s.io/mount-utils/mount_linux.go).
+	// We then retry force unmounting for 2 seconds.
+	// Thus the full timeout is 7 seconds.
+	forceUnmountRetryTimeout = 7 * time.Second
+	forceUnmountRetrySteps   = 6
 )
 
 // nodeServer handles mounting and unmounting of GCS FUSE volumes on a node.
@@ -567,7 +580,7 @@ func (s *nodeServer) isGcsFuseKernelParamsFeatureSupported(gcsFuseContainerImage
 		s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseContainerImage, GCSFuseKernelParamsMinVersion)
 }
 
-func (s *nodeServer) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
+func (s *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
 	// Validate arguments
 	targetPath := req.GetTargetPath()
 	if len(targetPath) == 0 {
@@ -604,12 +617,16 @@ func (s *nodeServer) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpubli
 		// mount.CleanupMountPoint() call will hang.
 		forceUnmounter, ok := s.mounter.(mount.MounterForceUnmounter)
 		if ok {
-			if err = forceUnmounter.UnmountWithForce(targetPath, UmountTimeout); err != nil {
+			if err = s.unmountWithRetry(ctx, targetPath, forceUnmountRetryTimeout, forceUnmountRetrySteps, func() error {
+				return forceUnmounter.UnmountWithForce(targetPath, UmountTimeout)
+			}); err != nil {
 				return nil, status.Errorf(codes.Internal, "failed to force unmount target path %q: %v", targetPath, err)
 			}
 		} else {
 			klog.Warningf("failed to cast the mounter to a forceUnmounter, proceed with the default mounter Unmount")
-			if err = s.mounter.Unmount(targetPath); err != nil {
+			if err = s.unmountWithRetry(ctx, targetPath, standardUnmountRetryTimeout, standardUnmountRetrySteps, func() error {
+				return s.mounter.Unmount(targetPath)
+			}); err != nil {
 				return nil, status.Errorf(codes.Internal, "failed to unmount target path %q: %v", targetPath, err)
 			}
 		}
@@ -638,6 +655,43 @@ func (s *nodeServer) isDirMounted(targetPath string) (bool, error) {
 	}
 
 	return false, nil
+}
+
+// unmountWithRetry retries the unmount operation using exponential backoff.
+// It uses a child context with a timeout to limit the total retry duration,
+// in addition to being limited to the defined number of steps.
+//
+// If the OS unmount system call itself hangs, this function will block
+// on the hung call. It won't be able to interrupt the hung call until the Kubelet cancels the parent
+// context (NodeUnpublishVolume context) and retries the entire operation.
+func (s *nodeServer) unmountWithRetry(ctx context.Context, targetPath string, timeout time.Duration, steps int, unmountFn func() error) error {
+	var lastErr error
+	childCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	backoff := wait.Backoff{
+		Duration: unmountRetryInterval,
+		Factor:   unmountRetryBackoffFactor,
+		Jitter:   unmountRetryJitter,
+		Steps:    steps,
+	}
+
+	pollErr := wait.ExponentialBackoffWithContext(childCtx, backoff, func(ctx context.Context) (bool, error) {
+		lastErr = unmountFn()
+		if lastErr == nil {
+			return true, nil
+		}
+		klog.Warningf("Failed to unmount %q: %v. Retrying...", targetPath, lastErr)
+		return false, nil
+	})
+
+	if pollErr != nil {
+		if lastErr != nil {
+			return lastErr
+		}
+		return pollErr
+	}
+	return nil
 }
 
 // setupMultiNIC updates args with options for multi NIC configuration.
--- a/pkg/csi_driver/node_test.go
+++ b/pkg/csi_driver/node_test.go
@@ -19,6 +19,7 @@ package driver
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
@@ -27,6 +28,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/cloud_provider/clientset"
@@ -101,8 +103,47 @@ type nodeServerTestEnv struct {
 }
 
 func initTestNodeServer(t *testing.T) *nodeServerTestEnv {
+	return initTestNodeServerWithMounter(t, mount.NewFakeMounter([]mount.MountPoint{}))
+}
+
+func initTestNodeServerWithCustomClientset(t *testing.T, clientSet *clientset.FakeClientset, wiNodeLabelCheck bool) *nodeServerTestEnv {
 	t.Helper()
 	mounter := mount.NewFakeMounter([]mount.MountPoint{})
+	driver := initTestDriverWithCustomNodeServer(t, mounter, clientSet, wiNodeLabelCheck)
+	s, _ := driver.config.StorageServiceManager.SetupService(context.TODO(), nil, "")
+	if _, err := s.CreateBucket(context.Background(), &storage.ServiceBucket{Name: testVolumeID}); err != nil {
+		t.Fatalf("failed to create the fake bucket: %v", err)
+	}
+
+	return &nodeServerTestEnv{
+		ns:    newNodeServer(driver, mounter),
+		fm:    mounter,
+		nwMgr: driver.config.NetworkManager.(*fakeNetworkManager),
+	}
+}
+
+type fakeForceUnmounter struct {
+	*mount.FakeMounter
+	unmountWithForceFunc func(target string, umountTimeout time.Duration) error
+}
+
+func (f *fakeForceUnmounter) UnmountWithForce(target string, umountTimeout time.Duration) error {
+	if f.unmountWithForceFunc != nil {
+		return f.unmountWithForceFunc(target, umountTimeout)
+	}
+	return f.FakeMounter.Unmount(target)
+}
+
+func initTestNodeServerWithForceMounter(t *testing.T) (*nodeServerTestEnv, *fakeForceUnmounter) {
+	fakeMounter := mount.NewFakeMounter([]mount.MountPoint{})
+	mounter := &fakeForceUnmounter{
+		FakeMounter: fakeMounter,
+	}
+	return initTestNodeServerWithMounter(t, mounter), mounter
+}
+
+func initTestNodeServerWithMounter(t *testing.T, mounter mount.Interface) *nodeServerTestEnv {
+	t.Helper()
 	podName := createMounterPodName("test-node", testVolumeID)
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
@@ -119,7 +160,14 @@ func initTestNodeServer(t *testing.T) *nodeServerTestEnv {
 			Name:      "test-pvc",
 		},
 	})
-	driver := initTestDriver(t, mounter, fakeClient)
+	var fakeMounter *mount.FakeMounter
+	if fm, ok := mounter.(*mount.FakeMounter); ok {
+		fakeMounter = fm
+	} else if ffm, ok := mounter.(*fakeForceUnmounter); ok {
+		fakeMounter = ffm.FakeMounter
+	}
+
+	driver := initTestDriver(t, fakeMounter, fakeClient)
 	s, _ := driver.config.StorageServiceManager.SetupService(context.TODO(), nil, "")
 	if _, err := s.CreateBucket(context.Background(), &storage.ServiceBucket{Name: testVolumeID}); err != nil {
 		t.Fatalf("failed to create the fake bucket: %v", err)
@@ -127,23 +175,7 @@ func initTestNodeServer(t *testing.T) *nodeServerTestEnv {
 
 	return &nodeServerTestEnv{
 		ns:    newNodeServer(driver, mounter),
-		fm:    mounter,
-		nwMgr: driver.config.NetworkManager.(*fakeNetworkManager),
-	}
-}
-
-func initTestNodeServerWithCustomClientset(t *testing.T, clientSet *clientset.FakeClientset, wiNodeLabelCheck bool) *nodeServerTestEnv {
-	t.Helper()
-	mounter := mount.NewFakeMounter([]mount.MountPoint{})
-	driver := initTestDriverWithCustomNodeServer(t, mounter, clientSet, wiNodeLabelCheck)
-	s, _ := driver.config.StorageServiceManager.SetupService(context.TODO(), nil, "")
-	if _, err := s.CreateBucket(context.Background(), &storage.ServiceBucket{Name: testVolumeID}); err != nil {
-		t.Fatalf("failed to create the fake bucket: %v", err)
-	}
-
-	return &nodeServerTestEnv{
-		ns:    newNodeServer(driver, mounter),
-		fm:    mounter,
+		fm:    fakeMounter,
 		nwMgr: driver.config.NetworkManager.(*fakeNetworkManager),
 	}
 }
@@ -800,7 +832,9 @@ func TestNodeUnpublishVolume(t *testing.T) {
 		mounts        []mount.MountPoint // already existing mounts
 		req           *csi.NodeUnpublishVolumeRequest
 		actions       []mount.FakeAction
+		unmountFunc   func(path string) error
 		expectedMount *mount.MountPoint
+		ctx           context.Context
 		expectErr     error
 	}{
 		{
@@ -832,6 +866,41 @@ func TestNodeUnpublishVolume(t *testing.T) {
 				TargetPath: testTargetPath,
 			},
 		},
+		{
+			name:   "retry unmount succeeds eventually",
+			mounts: []mount.MountPoint{{Device: testVolumeID, Path: testTargetPath}},
+			req: &csi.NodeUnpublishVolumeRequest{
+				VolumeId:   testVolumeID,
+				TargetPath: testTargetPath,
+			},
+			unmountFunc: func() func(path string) error {
+				calls := 0
+				return func(path string) error {
+					calls++
+					if calls < 3 {
+						return fmt.Errorf("umount: %s: target is busy", path)
+					}
+					return nil
+				}
+			}(),
+		},
+		{
+			name:   "retry unmount fails after max retries",
+			mounts: []mount.MountPoint{{Device: testVolumeID, Path: testTargetPath}},
+			req: &csi.NodeUnpublishVolumeRequest{
+				VolumeId:   testVolumeID,
+				TargetPath: testTargetPath,
+			},
+			unmountFunc: func(path string) error {
+				return fmt.Errorf("umount: %s: target is busy", path)
+			},
+			expectedMount: &mount.MountPoint{Device: testVolumeID, Path: testTargetPath},
+			ctx: func() context.Context {
+				ctx, _ := context.WithTimeout(context.Background(), 500*time.Millisecond)
+				return ctx
+			}(),
+			expectErr: status.Errorf(codes.Internal, "failed to unmount target path %q: umount: %s: target is busy", testTargetPath, testTargetPath),
+		},
 	}
 
 	for _, test := range cases {
@@ -840,8 +909,95 @@ func TestNodeUnpublishVolume(t *testing.T) {
 			if test.mounts != nil {
 				testEnv.fm.MountPoints = test.mounts
 			}
+			if test.unmountFunc != nil {
+				testEnv.fm.UnmountFunc = test.unmountFunc
+			}
 
-			_, err := testEnv.ns.NodeUnpublishVolume(context.TODO(), test.req)
+			ctx := context.TODO()
+			if test.ctx != nil {
+				ctx = test.ctx
+			}
+
+			_, err := testEnv.ns.NodeUnpublishVolume(ctx, test.req)
+			if test.expectErr == nil && err != nil {
+				t.Errorf("got error %q, expected error nil", err)
+			}
+			if test.expectErr != nil && !errors.Is(err, test.expectErr) {
+				t.Errorf("got error %q, expected error %q", err, test.expectErr)
+			}
+
+			validateMountPoint(t, testEnv.fm, test.expectedMount, nil)
+		})
+	}
+}
+
+func TestNodeUnpublishVolumeForceUnmount(t *testing.T) {
+	t.Parallel()
+	testTargetPath, cleanup := setupTestTargetPath(t)
+	defer cleanup()
+
+	cases := []struct {
+		name                 string
+		mounts               []mount.MountPoint
+		req                  *csi.NodeUnpublishVolumeRequest
+		unmountWithForceFunc func(target string, umountTimeout time.Duration) error
+		expectedMount        *mount.MountPoint
+		ctx                  context.Context
+		expectErr            error
+	}{
+		{
+			name:   "retry force unmount succeeds eventually",
+			mounts: []mount.MountPoint{{Device: testVolumeID, Path: testTargetPath}},
+			req: &csi.NodeUnpublishVolumeRequest{
+				VolumeId:   testVolumeID,
+				TargetPath: testTargetPath,
+			},
+			unmountWithForceFunc: func() func(target string, umountTimeout time.Duration) error {
+				calls := 0
+				return func(target string, umountTimeout time.Duration) error {
+					calls++
+					if calls < 3 {
+						return fmt.Errorf("umount: %s: target is busy", target)
+					}
+					return nil
+				}
+			}(),
+		},
+		{
+			name:   "retry force unmount fails after max retries",
+			mounts: []mount.MountPoint{{Device: testVolumeID, Path: testTargetPath}},
+			req: &csi.NodeUnpublishVolumeRequest{
+				VolumeId:   testVolumeID,
+				TargetPath: testTargetPath,
+			},
+			unmountWithForceFunc: func(target string, umountTimeout time.Duration) error {
+				return fmt.Errorf("umount: %s: target is busy", target)
+			},
+			expectedMount: &mount.MountPoint{Device: testVolumeID, Path: testTargetPath},
+			ctx: func() context.Context {
+				ctx, _ := context.WithTimeout(context.Background(), 500*time.Millisecond)
+				return ctx
+			}(),
+			expectErr: status.Errorf(codes.Internal, "failed to force unmount target path %q: umount: %s: target is busy", testTargetPath, testTargetPath),
+		},
+	}
+
+	for _, test := range cases {
+		t.Run(test.name, func(t *testing.T) {
+			testEnv, forceMounter := initTestNodeServerWithForceMounter(t)
+			if test.mounts != nil {
+				testEnv.fm.MountPoints = test.mounts
+			}
+			if test.unmountWithForceFunc != nil {
+				forceMounter.unmountWithForceFunc = test.unmountWithForceFunc
+			}
+
+			ctx := context.TODO()
+			if test.ctx != nil {
+				ctx = test.ctx
+			}
+
+			_, err := testEnv.ns.NodeUnpublishVolume(ctx, test.req)
 			if test.expectErr == nil && err != nil {
 				t.Errorf("got error %q, expected error nil", err)
 			}


pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,125 @@
...
+		mc.VolumeName = mc.BucketName
+		if mc.VolumeName == "" {
+			mc.VolumeName = filepath.Base(filepath.Dir(mountPoint))
+		}
+	}
+
+	tmpVolDir := filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, ".volumes", mc.VolumeName)
+	if err := os.MkdirAll(tmpVolDir, 0750); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create volume temp dir: %v", err)
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: those inputs are set during node stage volume, taken from the request created by kubelet. I think we can trust them

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,121 @@
...
+	mc.BucketName = req.GetBucketName()
+	mc.Options = req.GetMountOptions()
+	mc.MountPoint = mountPoint
+
+	if mc.VolumeName == "" {
+		mc.VolumeName = mc.BucketName
+		if mc.VolumeName == "" {
+			mc.VolumeName = filepath.Base(filepath.Dir(mountPoint))
+		}
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: mount will only get called once

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,57 @@ var (
...
 
 	// This is set at compile time.
 	version = "unknown"
 )
 
+func runNodeMounter(ctx context.Context, cancel context.CancelFunc, mounter *sidecarmounter.Mounter) {
+	s := driver.NewNonBlockingGRPCServer()
+	ms := sidecarmounter.NewMounterServer(ctx, mounter)
+
+	socketFile := filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, "mounter.sock")
```
uriel-guzman: Use the mounterPodSocketFile instead of hardcoding it: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e96e22880ff0b790dd50c2140a30d29512841d64/pkg/csi_driver/shared_mount.go#L47

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,57 @@ var (
...
+	}
+	socketPath := fmt.Sprintf("unix://%s", socketFile)
+
+	s.Start(socketPath, nil, nil, nil, ms)
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
+	klog.Info("Waiting for SIGTERM signal...")
+	<-c
+	klog.Info("Received SIGTERM signal, waiting for all gcsfuse processes to exit...")
```
uriel-guzman: There's only 1 gcsfuse process

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,57 @@ var (
 	storageServiceAndBucketAccessFactor   = flag.Float64("storage-service-check-retry-factor", 2.0, "storage service creation and bucket access check exponential retry factor")
 	storageServiceAndBucketAccessJitter   = flag.Float64("storage-service-check-retry-jitter", 0.1, "storage service creation and bucket access check exponential retry jitter")
 	storageServiceAndBucketAccessDuration = flag.Duration("storage-service-check-retry-duration", 5*time.Second, "storage service creation and bucket access check exponential retry initial duration")
+	sharedMount                           = flag.Bool("shared-mount", false, "whether to run gcsfuse in a mounter pod")
 
 	// This is set at compile time.
 	version = "unknown"
 )
 
+func runNodeMounter(ctx context.Context, cancel context.CancelFunc, mounter *sidecarmounter.Mounter) {
```
uriel-guzman: Add a log at the beginning here: `klog.Infof("Running node mounter...")` so we can easily see when node mounter is enabled.

pkg/csi_driver/node.go
```diff
@@ -735,9 +736,22 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	}
 
 	podUID := string(pod.UID)
+	bucketName := util.ParseVolumeID(req.GetVolumeId())
+	fuseMountOptions := []string{}
+
+	if capMount := req.GetVolumeCapability().GetMount(); capMount != nil {
```
uriel-guzman: It seems like this was copy pasted from here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e96e22880ff0b790dd50c2140a30d29512841d64/pkg/csi_driver/utils.go#L368-L376

I don't think we should have duplicate code. Can we just either modify `parseRequestArguments` or create new helper functions so this type of logic is only defined in one single place? 

Also, `parseRequestArguments` already parses the bucketName and puts it into `args`: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/e96e22880ff0b790dd50c2140a30d29512841d64/pkg/csi_driver/utils.go#L379, so this entire block is duplicated.

pkg/csi_driver/node.go
```diff
@@ -735,9 +736,22 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+		fuseMountOptions = joinMountOptions(fuseMountOptions, capMount.GetMountFlags())
+	}
+	args, err := parseVolumeAttributes(fuseMountOptions, req.GetVolumeContext())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse stage request arguments: %v", err)
+	}
 
 	// Send GRPC to mounter pod to start GCSFuse.
-	if err := s.mountToNode(ctx, podUID, stagingPath, req.GetVolumeId()); err != nil {
+	if err := s.mountToNode(ctx, podUID, stagingPath, bucketName, args.fuseMountOptions); err != nil {
```
uriel-guzman: When you refactor like I mentioned above, you'll only needs to pass `args`, since `bucketName` is already contained within `args`.

pkg/csi_driver/node.go
```diff
@@ -788,16 +802,14 @@ func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath, volum
...
-			return status.Errorf(codes.Internal, "failed to mount: %v", err)
-		}
-	*/
+	c := mounter.NewMounterClient(conn)
+	if _, err := c.Mount(ctx, &mounter.MountRequest{
+		TargetPath:   stagingPath,
+		BucketName:   bucketName,
+		MountOptions: fuseMountOptions,
+	}); err != nil {
+		return status.Errorf(codes.Internal, "failed to mount: %v", err)
```
uriel-guzman: This is going to transform any error code coming from the Mount gRPC call as Internal, which is incorrect, since Mount could return PermissionDenied errors if KSA doesn't have access. I think we've gone over this quite a few times.

pkg/csi_driver/shared_mount.go
```diff
@@ -150,18 +150,19 @@ func createMounterPodSpec(config *mounterPodConfig) *corev1.Pod {
 					Name:            mounterPodNamePrefix,
 					Image:           config.image,
 					ImagePullPolicy: corev1.PullAlways,
+					Args:            []string{"--shared-mount"},
```
uriel-guzman: Make the name of the flag `enable-shared-mount` for consistency. This is the feature flag we've been using: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fgcs-fuse-csi-driver%20enable-shared-mount&type=code.

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,57 @@ var (
 	storageServiceAndBucketAccessFactor   = flag.Float64("storage-service-check-retry-factor", 2.0, "storage service creation and bucket access check exponential retry factor")
 	storageServiceAndBucketAccessJitter   = flag.Float64("storage-service-check-retry-jitter", 0.1, "storage service creation and bucket access check exponential retry jitter")
 	storageServiceAndBucketAccessDuration = flag.Duration("storage-service-check-retry-duration", 5*time.Second, "storage service creation and bucket access check exponential retry initial duration")
+	sharedMount                           = flag.Bool("shared-mount", false, "whether to run gcsfuse in a mounter pod")
```
uriel-guzman: Make the name of the flag enable-shared-mount for consistency. This is the feature flag we've been using: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fgcs-fuse-csi-driver%20enable-shared-mount&type=code.

Also, change the var to `enableSharedMount`.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/proto/mounter"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/klog/v2"
+)
+
+const (
+	mounterPodMountTimeout = 30 * time.Second
```
uriel-guzman: This is vague, rename the variable to what this timeout is controlling. In this case, it's controlling how much time to wait until it gives up attempting to mount, so it could be called `mountTimeout`.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type MounterServer struct {
+	mounter.UnimplementedMounterServer
+	Mounter *Mounter
+	Config  MountConfig
```
uriel-guzman: Where is this being used? It's not even being set by NewMounterServer?

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,113 @@
...
+
+func NewMounterServer(ctx context.Context, config MountConfig, mounter *Mounter) *MounterServer {
+	return &MounterServer{
+		ctx:     ctx,
+		Config:  config,
+		Mounter: mounter,
+	}
+}
+
+func (ms *MounterServer) Mount(_ context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
```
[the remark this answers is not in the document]
uriel-guzman: Why isn't this addressed?

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+
+func NewMounterServer(ctx context.Context, mounter *Mounter) *MounterServer {
+	return &MounterServer{
+		ctx:     ctx,
+		Mounter: mounter,
+	}
+}
+
+func (ms *MounterServer) Mount(_ context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
+	mountPoint := filepath.Clean(req.GetTargetPath())
```
uriel-guzman: TargetPath can be confused with the CSI Driver's NodePublishVolume TargetPath (in CSI terminology, target path is the bind-mount, and staging path is well the staging path). Instead, make this name generic and just call it `Mountpoint`. You're renaming TargetPath to mountPoint anyway, so it's already kind of weird.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+	if mountPoint == "" || mountPoint == "." || mountPoint == "/" {
+		return nil, status.Error(codes.InvalidArgument, "invalid target path")
```
uriel-guzman: This should be Internal, not InvalidArgument, because if it's missing it's our fault, not the customer's.

uriel-guzman: Also, why would mountPoint be "."?  Or why would it be "/"? These are odly specific and I can't think of a reason why they would be named like that. I'd just leave the empty mount point validation and change the error message to say "mount point can't be empty".

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	bucketName := req.GetBucketName()
+	if bucketName == "" {
+		return nil, status.Error(codes.InvalidArgument, "bucket name cannot be empty")
```
uriel-guzman: Again, this should be Internal, and it should say "bucket name can't be empty". Also, these validations are very weird.... `strings.Contains(bucketName, "..") || bucketName == "." `. I don't think they make any sense at all...

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if strings.Contains(bucketName, "/") || strings.Contains(bucketName, "..") || bucketName == "." {
```
uriel-guzman: Same comment as above. These validaitons don't make any sense, error code should be Internal.

uriel-guzman (changes requested):

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, status.Error(codes.InvalidArgument, "invalid bucket name")
+	}
+
+	// Verify that the mount point is nested under the expected GCSFuse plugins directory to prevent directory traversal.
```
uriel-guzman: Why would this help with preventing directory traversal? If someone has the ability to change the mounter pod binary (e.g. private sidecar) to access the hostPath, they can trivially change this too, in which this validation does absolutely nothing, so I'd remove this validation entirely.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+	expectedPrefix := util.KubeletPluginsGCSFuseDir + "/"
+	if !strings.HasPrefix(mountPoint, expectedPrefix) {
+		return nil, status.Errorf(codes.InvalidArgument, "target path %q is not within the valid directory (expected prefix %q)", mountPoint, expectedPrefix)
+	}
+
+	volumeName := filepath.Base(filepath.Dir(mountPoint))
```
uriel-guzman: You're trying to guess the volumeName by looking at the directory before the mount point: `/var/lib/kubelet/plugins/kubernetes.io/csi/gcsfuse.csi.storage.gke.io/<sha256(volumeID)>/globalmount`, which has a sha256 hash applied to the volumeID, so this doesn't make sense. Also, you could just get volumeID by adding it into the Mounter server's gRPC proto instead of bucketName, and then deriving the bucketName from volumeID trivially from the mounter pod here.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,122 @@
...
+
+	tmpVolDir := filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, ".volumes", volumeName)
```
uriel-guzman: Why is this using `.volumes`? Didn't we say there's only 1 volume per mounter pod so you don't need this?

uriel-guzman (changes requested):

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: Sure ill update the design doc to reflect this

pkg/sidecar_mounter/mounter_server.go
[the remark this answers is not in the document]
chrisThePattyEater: // Use an independent context, as using the CSI Driver's ctx can cause
// a race condition where the driver attempts to remove the staging path
// while gcsfuse has not exited yet.

I was following the comment that warned against using this ctx from parallelstore, therefore the ctx was not needed.

pkg/csi_driver/node.go
```diff
@@ -735,9 +739,20 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	}
 
 	podUID := string(pod.UID)
+	volumeID := req.GetVolumeId()
+	readOnly, err := strconv.ParseBool(publishContext[PublishContextReadOnly])
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to parse read-only flag from publish context: %v", err)
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: controller publish volume is introduced in this change therefore if there is any publish context at all we can assume that the controller sent it

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/proto/mounter"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/klog/v2"
+)
+
+const (
+	mountTimeout = 30 * time.Second
+)
```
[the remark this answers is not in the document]
chrisThePattyEater: removed

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,100 @@
...
+	mountPoint := req.GetMountPoint()
+	volumeID := req.GetVolumeId()
+	bucketName := util.ParseVolumeID(volumeID)
+
+	tmpVolDir := filepath.Join(webhook.SidecarContainerTmpVolumeMountPath)
+	if err := os.MkdirAll(tmpVolDir, 0750); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create volume temp dir: %v", err)
+	}
+
+	mc := ms.Config
```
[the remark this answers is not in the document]
chrisThePattyEater: mounter pod only handles one mount

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: fixed

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: fixed

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: fixed

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: fixed

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: fixed

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: fixed

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,58 @@ var (
...
+	enableSharedMount                     = flag.Bool("enable-shared-mount", false, "whether to run gcsfuse in a mounter pod")
 
 	// This is set at compile time.
 	version = "unknown"
 )
 
+func runNodeMounter(ctx context.Context, cancel context.CancelFunc, mounter *sidecarmounter.Mounter) {
+	klog.Infof("Running node mounter...")
+	s := driver.NewNonBlockingGRPCServer()
+	ms := sidecarmounter.NewMounterServer(ctx, mounter, sidecarmounter.MountConfig{})
```
uriel-guzman: Why are we passing an empty `sidecarmounter.MountConfig{}`? If you don't need the paramter, then delete it from the function definition?

pkg/csi_driver/controller.go
```diff
@@ -254,6 +255,7 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
 			// NodePublishVolume calls to avoid re-computation.
 			PublishContextKeyMounterPodNamespace: podNamespace,
 			PublishContextKeyMounterPodName:      podName,
+			PublishContextReadOnly:               strconv.FormatBool(req.GetReadonly()),
```
uriel-guzman: Why do we need to pass ReadOnly from ControllerPublish to NodeStage? 

According to LLM:

> Why readonly is in ControllerPublishVolume & NodePublishVolume but NOT NodeStageVolume:
ControllerPublishVolume: The readonly option here applies to the device attachment to the node. Some block storage systems allow a device to be attached as read-only, preventing any writes from the node's OS.
NodeStageVolume: The NodeStageVolumeRequest in the CSI specification does not have a readonly field. The staging path is an intermediate, system-managed mount point on the node. It's generally mounted with read/write permissions to allow the node's system to manage it. The fine-grained read-only control for workloads is not typically enforced at this stage.
NodePublishVolume: This is where the volume is made visible to the actual container. The readonly field here directly corresponds to how the pod requests the volume (e.g., via PersistentVolumeClaim or VolumeMounts in Kubernetes). The CSI driver uses this flag to perform a bind mount with read-only options (e.g., mount -o bind,ro ...).

So, it doesn't seem like NodeStageVolume mounting necessarily matters here, because during NodePublishVolume, each bind mount will have the `ro` mount option anyway, so I think this is not needed? Can you please confirm?

pkg/csi_driver/node.go
```diff
@@ -135,13 +136,16 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	// Validate arguments
-	args, err := parseRequestArguments(req, vc)
```
uriel-guzman: This seems to be done because you're trying to parse read only from both controller publish and node publish? But as discussed above, you don't actually need it, so can we revert this node publish volume to how it was before? Or is there another reason why you need this refactor aside from parsing readonly (which we don't seem to need?)

pkg/csi_driver/node.go
```diff
@@ -698,6 +702,10 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	publishContext := req.GetPublishContext()
 	podName := publishContext[PublishContextKeyMounterPodName]
 	podNamespace := publishContext[PublishContextKeyMounterPodNamespace]
+	readOnly, err := util.ParseBool(publishContext[PublishContextReadOnly])
+	if err != nil {
+		return nil, status.Error(codes.Internal, "publishContext must contain a valid read-only flag")
+	}
```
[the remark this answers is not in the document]
uriel-guzman: We don't need readonly here anyway, so you acn remove that and resolve this comment.

pkg/csi_driver/node.go
```diff
@@ -735,9 +743,16 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	}
 
 	podUID := string(pod.UID)
+	volumeID := req.GetVolumeId()
+
+	volumeCapability := req.GetVolumeCapability()
+	args, err := parseRequestArguments(volumeID, readOnly, volumeCapability, req.GetVolumeContext())
```
uriel-guzman: Since you don't need to parse readonly, maybe just set it to false?

pkg/csi_driver/node.go
```diff
@@ -748,7 +763,7 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 }
 
 // mountToNode connects to the mounter server, at which point it initializes the GCSFuse process.
-func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath, volumeID string) error {
+func (s *nodeServer) mountToNode(ctx context.Context, podUID, stagingPath string, volumeID string, fuseMountOptions []string) error {
```
uriel-guzman: Nit: `stagingPath, volumeID string`

pkg/sidecar_mounter/mounter_server.go
uriel-guzman: This comment isn't resolved yet.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,107 @@
...
+		ctx:     ctx,
+		Mounter: mounter,
+		Config:  config,
+	}
+}
+
+func (ms *MounterServer) Mount(ctx context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.Internal, "mount request cannot be nil")
+	}
```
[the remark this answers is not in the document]
uriel-guzman: +1, can you please just check that these fields aren't empty? We should have validated them in NodeStage, so they should just be Internal.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,107 @@
...
+
+	mc, err := ms.prepareMountConfig(req)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "%v", err)
+	}
+
+	// TODO(FUECHR): Implement Cloud Profiler hook (profiler.Start).
```
uriel-guzman: Add TODO to implement ErrWriter.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,107 @@
...
+
+func NewMounterServer(ctx context.Context, mounter *Mounter, config MountConfig) *MounterServer {
+	return &MounterServer{
+		ctx:     ctx,
+		Mounter: mounter,
+		Config:  config,
+	}
+}
+
+func (ms *MounterServer) Mount(ctx context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
```
uriel-guzman: Isn't `ctx` unused? If so, make it `_`.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// TODO(FUECHR): Implement SetupTokenAndStorageManager for sidecar bucket access check.
+
+	if err := mc.prepareConfigFile(); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to prepare config file: %v", err)
+	}
+
+	klog.Infof("Start mounting bucket %q to %q for volume %q", mc.BucketName, mc.MountPoint, mc.VolumeName)
+
+	// Clean up any errors raised until this point due to transient metadata and storage service creation issues.
```
uriel-guzman: what metadata and storage service creation issues are you referring to? Those are not even implemented yet, maybe delete this block of code until they are implemented and add a TODO so we don't forget?

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := ms.Mounter.Mount(ms.ctx, &mc); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to mount: %v", err)
+	}
+
+	return &mounter.MountResponse{}, nil
+}
+
+// prepareMountConfig handles populating all required directory paths, error writers, and fields for MountConfig.
+func (ms *MounterServer) prepareMountConfig(req *mounter.MountRequest) (MountConfig, error) {
+	mc := ms.Config
```
uriel-guzman: Is `ms.Config` actually used at all?

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,107 @@
...
+	mc.VolumeName = req.GetVolumeId()
```
uriel-guzman: `mc.VolumeName` in the current CSI Driver is the `<volume-name>` component of the emptyDir https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/sidecar_mounter/sidecar_mounter_config.go#L136

That value is derived from the target path here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/39f676bb19b38a4d887f7ad6a8263b348749b790/pkg/util/util.go#L186, and the target path's volume component is actually the PV / ephemeral volume name, NOT the volume ID. The volume ID being passed to the `Mount` gRPC call is actually the volumeHandle in the PV spec (which is usually the bucket), so they are entirely two different things.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -73,6 +73,7 @@ func New(mounterPath string) *Mounter {
 }
 
 func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
```
uriel-guzman, line 75: This shouldn't be calling `Mounter.Mount`. There's too many differences and it has the potential of creating bugs on the existing sidecar architecture. Instead, you should create a new function `MountToNode` and separate the lagic. You can wrap helper functions if there's duplicate code as required, but the current way you have it I think is too risky.

uriel-guzman (changes requested):

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: Changes this to instantiate a new mount config in prepareMountConfig

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: I have replace this with a todo.

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: I have replaced this with an _ since it is not being used.

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: Added.

pkg/csi_driver/node.go
chrisThePattyEater: fixed

pkg/csi_driver/node.go
chrisThePattyEater: No longer passing readonly and now am just setting to false.

pkg/csi_driver/node.go
chrisThePattyEater: Correct, parseRequestArguments has been refactored to allow us to pass readOnly but this needs to be used in nodeStageVolume to retrieve mount options to be able to initiate gcsfuse. I think we should leave this to allow reuse. Let me know your thoughts.

pkg/csi_driver/controller.go
chrisThePattyEater: As discussed offline I have removed this and am hardcoding to false

cmd/sidecar_mounter/main.go
chrisThePattyEater: Fixed

pkg/sidecar_mounter/mounter_server.go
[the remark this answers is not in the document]
chrisThePattyEater: added

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: I am now passing the volume name from the volume context via the mount request

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := ms.Mounter.Mount(ms.ctx, &mc); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to mount: %v", err)
+	}
+
+	return &mounter.MountResponse{}, nil
+}
+
+// prepareMountConfig handles populating all required directory paths, error writers, and fields for MountConfig.
+func (ms *MounterServer) prepareMountConfig(req *mounter.MountRequest) (MountConfig, error) {
+	tmpVolDir := filepath.Join(webhook.SidecarContainerTmpVolumeMountPath)
```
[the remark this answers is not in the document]
chrisThePattyEater: each mounter pod has 1 volume so this should not be a scenario

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -131,68 +131,96 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	if err := mc.ErrWriter.Clean(); err != nil {
+		klog.Warningf("failed to cleanup error file: %v", err)
+	}
 
-		return cmd.Process.Signal(syscall.SIGTERM)
+	klog.Infof("start to mount bucket %q for volume %q", mc.BucketName, mc.VolumeName)
+
+	if err := os.MkdirAll(mc.BufferDir+TempDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create temp dir %q: %w", mc.BufferDir+TempDir, err)
 	}
```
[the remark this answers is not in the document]
chrisThePattyEater: ack

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,116 @@
...
+		Mounter: mounter,
+		ctx:     ctx,
+	}
+}
+
+func (ms *MounterServer) Mount(_ context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.Internal, "mount request cannot be nil")
+	}
+	if req.GetVolumeId() == "" {
```
uriel-guzman: Also validate VolumeName?

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,116 @@
...
+		mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to mount bucket %q for volume %q: %v\n", mc.BucketName, mc.VolumeName, err))
+	}
+
+	return &mounter.MountResponse{}, nil
+}
+
+// prepareMountConfig handles populating all required directory paths, error writers, and fields for MountConfig.
+func (ms *MounterServer) prepareMountConfig(req *mounter.MountRequest) (MountConfig, error) {
+	tmpVolDir := webhook.SidecarContainerTmpVolumeMountPath
+	if err := os.MkdirAll(tmpVolDir, 0750); err != nil {
```
uriel-guzman: `/gcsfuse-tmp` is a volume that is part of the gke-gcsfuse-sidecar container, so I don't see it being created explicitly here:  https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/240cf5703788ccdc8c5fc9d4aff44a527c2906c7/pkg/sidecar_mounter/sidecar_mounter_config.go#L135

It's also a default volume in the mounter pod spec: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/shared_mount.go#L144, so it doesn't need to be created explicitly here. This line is redundant.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,116 @@
...
+	mc := MountConfig{
+		VolumeName:       req.GetVolumeName(),
+		BucketName:       util.ParseVolumeID(req.GetVolumeId()),
+		Options:          req.GetMountOptions(),
+		SharedMountPoint: req.GetMountPoint(),
+		TempDir:          tmpVolDir,
+		ErrWriter:        NewErrorWriter(filepath.Join(tmpVolDir, util.ErrorFileName)),
+		BufferDir:        webhook.SidecarContainerBufferVolumeMountPath,
+		CacheDir:         webhook.SidecarContainerCacheVolumeMountPath,
+		ConfigFile:       filepath.Join(tmpVolDir, "config.yaml"),
```
uriel-guzman: We're missing `AutoGoMemLimitRatio`: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/240cf5703788ccdc8c5fc9d4aff44a527c2906c7/pkg/sidecar_mounter/sidecar_mounter_config.go#L146

This should be implemented in a separate PR. Add TODO to not forget.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return &mounter.MountResponse{}, nil
+}
+
+// prepareMountConfig handles populating all required directory paths, error writers, and fields for MountConfig.
+func (ms *MounterServer) prepareMountConfig(req *mounter.MountRequest) (MountConfig, error) {
+	tmpVolDir := webhook.SidecarContainerTmpVolumeMountPath
+	if err := os.MkdirAll(tmpVolDir, 0750); err != nil {
+		return MountConfig{}, fmt.Errorf("failed to create volume temp dir: %w", err)
```
uriel-guzman: Since you don't need to create the tmpDir here, you don't need to return an error, so make this function only return *MountConfig.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,116 @@
...
+	if req.GetVolumeId() == "" {
+		return nil, status.Error(codes.Internal, "volume id cannot be empty")
+	}
+	if req.GetMountPoint() == "" {
+		return nil, status.Error(codes.Internal, "mount point cannot be empty")
+	}
+
+	mc, err := ms.prepareMountConfig(req)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "%v", err)
```
uriel-guzman: Be more descriptive: "failed to prepare mount config: %v" rather than just "%v". This helps code search and pin down the errors when debugging.

proto/mounter/mounter.proto
```diff
@@ -12,6 +12,7 @@ message MountRequest {
   string mount_point = 1;
   string volume_id = 2;
   repeated string mount_options = 3;
+  string volume_name = 4;
```
yaozile123: Can you regenerate `mounter.pb.go`? The `mount_options` was 3 and `volume_name` was 4 in here , but in `mounter.pb.go` the `mount_options` become 4 and `volume_name` become 3

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,116 @@
...
+
+	klog.Infof("Start mounting bucket %q to %q for volume %q", mc.BucketName, mc.SharedMountPoint, mc.VolumeName)
+
+	// TODO(FUECHR): Implement Cleaning of existing errors in ErrorWriter before mounting.
+
+	// Use an independent context to prevent the gcsfuse process from hanging indefinitely.
+	mountCtx, mountCancel := context.WithTimeout(ms.ctx, mountTimeout)
+	defer mountCancel()
+	if err := ms.Mounter.MountToNode(mountCtx, &mc); err != nil {
+		mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to mount bucket %q for volume %q: %v\n", mc.BucketName, mc.VolumeName, err))
```
uriel-guzman: If you fail to mount the bucket, then return the error? Why are you writing it to the error file? Then in L93 you're just returning nil, causing NodeStageVolume to succeed, and then waiting until NodePublishVolume to see the error file. Instead, you could've just returned the error immediately, not written it to the error file, and made NodeStageVolume fail fast.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -235,6 +262,69 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	cmd := exec.CommandContext(ctx, cmdName, args...)
+
+	if mc.EnableAutoGoMemLimit {
+		appliedLimit, err := memlimit.SetGoMemLimitWithOpts(
+			memlimit.WithRatio(mc.AutoGoMemLimitRatio),
+			memlimit.WithProvider(memlimit.FromCgroup),
+		)
+		if err != nil {
+			klog.V(4).Infof("GOMEMLIMIT not set (expected if gke-gcsfuse-sidecar memory limit is unset): %v", err)
+		} else if appliedLimit > 0 {
```
uriel-guzman: Can we keep the original comments? I think they were helpful, not sure why you deleted them.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -235,6 +262,69 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	return nil
 }
 
+func startCleanupMonitor(ctx context.Context, cmd *exec.Cmd) {
```
uriel-guzman: The refactor into a separate helper function (startCleanupMonitor) is much cleaner to read, but it introduces a closure behavior and leaves an existing panic vulnerability unaddressed.

Explicit Argument Passing vs. Closure: In the new startCleanupMonitor, the goroutine implicitly captures cmd as a closure. If cmd is ever mutated or reused outside this function, it can cause data races. It is safer to pass cmd explicitly into the goroutine's signature (like the original code) to freeze the pointer address at execution time.

Nil Pointer Panic Risk (Critical): cmd.Process remains nil until cmd.Start() or cmd.Run() successfully executes. If ctx.Done() fires before the process fully kicks off, accessing cmd.Process.Pid will trigger a nil pointer dereference panic.

uriel-guzman (changes requested):

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,58 @@ var (
...
+	socketPath := fmt.Sprintf("unix:%s", socketFile)
+
+	s.Start(socketPath, nil, nil, nil, ms)
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
+	klog.Info("Waiting for SIGTERM signal...")
+	<-c
+	klog.Info("Received SIGTERM signal, waiting for the gcsfuse process to exit...")
+
```
yaozile123: I think we should call `s.Stop()` before `mounter.WaitGroup.Wait()` to stop accepting new gRPC requests. Otherwise any incoming RPC during Wait() will invoke WaitGroup.Add() and trigger a runtime panic.

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: Added

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: removed

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: I got rid of the whole function since now its just a one liner

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: added them back

cmd/sidecar_mounter/main.go
chrisThePattyEater: I don't think this is a valid case since the only one sending grpc requests would be the csi driver and there should only be one NSV called per volume. I will add since it doesn't hurt.

pkg/sidecar_mounter/mounter_server.go
chrisThePattyEater: added

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: I am now passing the cmd ptr as it was done before and performing a nil check to prevent panics

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if err := mc.prepareConfigFile(); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to prepare config file: %v", err)
+	}
+
+	klog.Infof("Start mounting bucket %q to %q for volume %q", mc.BucketName, mc.SharedMountPoint, mc.VolumeName)
+
+	// TODO(FUECHR): Implement Cleaning of existing errors in ErrorWriter before mounting.
+
+	// Use the mounter servers lon running ctx to prevent the one from NodeStageVolume from killing the gcsfuse process.
```
yaozile123: "lon" -> "long"

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -131,68 +131,98 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	// Clear the GCS Fuse error file after the  bucket access check completes and just before the GCS Fuse mount starts.
+	// This ensures the NodePublishVolume/NodeStageVolume will always fail if the error persists until the GCS Fuse mount actually starts.
+	if err := mc.ErrWriter.Clean(); err != nil {
+		klog.Warningf("failed to cleanup error file: %v", err)
+	}
 
-		return cmd.Process.Signal(syscall.SIGTERM)
+	klog.Infof("start to mount bucket %q for volume %q", mc.BucketName, mc.VolumeName)
+
+	if err := os.MkdirAll(filepath.Join(mc.BufferDir, TempDir), os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
uriel-guzman: We don't even need to create this directory in the first place @chrisThePattyEater. It's already part of the pod spec, same as the temp dir. Why are you creating it?

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, status.Errorf(codes.Internal, "failed to prepare config file: %v", err)
+	}
+
+	klog.Infof("Start mounting bucket %q to %q for volume %q", mc.BucketName, mc.MountPoint, mc.VolumeName)
+
+	// TODO(FUECHR): Implement Cleaning of existing errors in ErrorWriter before mounting.
+
+	// Use the long-lived server context (ms.ctx) so that the gcsfuse process
+	// is not killed when the short-lived mountCtx is cancelled.
+	if err := ms.Mounter.MountToNode(ms.ctx, &mc); err != nil {
```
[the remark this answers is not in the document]
uriel-guzman: I agree with this comment. It's important to not use the CSI Driver's ctx for the mount because it'd kill the gcsfuse process, but for other stuff that isn't to keep the gcsfuse mount alive, using the ctx provided by the CSI Driver is the correct approach, because if NPV times out, it should cancel all things related to the context.

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, status.Errorf(codes.Internal, "failed to mount: %v", err)
+	}
```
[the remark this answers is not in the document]
uriel-guzman: This comment is correct? Why did you close it?

pkg/csi_driver/node.go
```diff
@@ -890,10 +891,8 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		return nil, err
 	}
 
-	podUID := string(pod.UID)
-
 	// Send GRPC to mounter pod to start GCSFuse.
-	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, args.fuseMountOptions); err != nil {
+	if err := s.mountToNode(ctx, string(pod.UID), stagingPath, vc[util.VolumeContextKeyPVName], volumeID, args.fuseMountOptions); err != nil {
```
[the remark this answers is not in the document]
chrisThePattyEater: adding a check at the top of NSV to verify volume name has been passed. Its incorrect to use volume id instead. and should return an error

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,94 @@
...
+	if req.GetVolumeId() == "" {
+		return nil, status.Error(codes.Internal, "volume id cannot be empty")
+	}
+	if req.GetVolumeName() == "" {
+		return nil, status.Error(codes.Internal, "volume name cannot be empty")
+	}
+	if req.GetMountPoint() == "" {
+		return nil, status.Error(codes.Internal, "mount point cannot be empty")
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is wrong

pkg/sidecar_mounter/sidecar_mounter.go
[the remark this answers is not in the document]
chrisThePattyEater: removed this

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,58 @@ var (
...
+func runNodeMounter(ctx context.Context, cancel context.CancelFunc, mounter *sidecarmounter.Mounter) {
+	klog.Infof("Running node mounter...")
+	s := driver.NewNonBlockingGRPCServer()
+	ms := sidecarmounter.NewMounterServer(ctx, mounter)
+
+	socketFile := filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, driver.MounterPodSocketFile)
+
+	if err := os.Remove(socketFile); err != nil && !os.IsNotExist(err) {
+		klog.Warningf("failed to remove stale socket file %q: %v", socketFile, err)
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: no harm in adding this defensive change, added

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -131,68 +131,99 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-	}(cmd)
+		args = append(args, fmt.Sprintf("--%s", arg))
+	}
+
+	args = append(args, mc.BucketName)
+	args = append(args, mc.SharedMountPoint)
+
+	cmd := m.prepareMountCommand(ctx, mc, args)
+
+	startCleanupMonitor(ctx, cmd)
```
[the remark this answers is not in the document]
chrisThePattyEater: I am pretty sure startCleanupMonitor creates its own background goroutine that blocks on <-ctx.Done(). Because of this, it doesn't immediately check cmd.Process.

pkg/sidecar_mounter/mounter_server.go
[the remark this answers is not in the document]
chrisThePattyEater: yeah this is now using the sidecars ctx

pkg/sidecar_mounter/mounter_server.go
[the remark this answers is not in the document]
chrisThePattyEater: This has been corrected

pkg/csi_driver/node.go
```diff
@@ -849,6 +850,9 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	var profile *profiles.ProfileConfig
 	if profilesEnabled {
 		klog.V(4).Infof("NodeStageVolume gcsfuse profiles feature is enabled for mounter pod %s/%s", podNamespace, podName)
+		if req.GetVolumeContext()[util.VolumeContextKeyPVName] == "" {
+			return nil, status.Error(codes.InvalidArgument, "NodeStageVolume VolumeContext must contain PV name")
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: added a nil check

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,94 @@
...
+		ConfigFile:       filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, "config.yaml"),
+		// TODO(FUECHR): Implement Auto Go Mem Limit.
+	}
+
+	// TODO(FUECHR): Implement Cloud Profiler hook (profiler.Start).
+	// TODO(FUECHR): Implement ErrorWriter.
+	// TODO(FUECHR): Implement defaultingFlagFileParsing.
+
+	// TODO(FUECHR): Make prometheusPort increment inside prepareMountArgs thread-safe.
+	mc.prepareMountArgs()
```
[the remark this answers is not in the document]
chrisThePattyEater: There is a comment to "Make prometheusPort increment inside prepareMountArgs thread-safe."
Will fix this in follow up pr

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,94 @@
...
+		BucketName:       util.ParseVolumeID(req.GetVolumeId()),
+		Options:          req.GetMountOptions(),
+		SharedMountPoint: req.GetMountPoint(),
+		TempDir:          webhook.SidecarContainerTmpVolumeMountPath,
+		ErrWriter:        NewErrorWriter(filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, util.ErrorFileName)),
+		BufferDir:        webhook.SidecarContainerBufferVolumeMountPath,
+		CacheDir:         webhook.SidecarContainerCacheVolumeMountPath,
+		ConfigFile:       filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, "config.yaml"),
+		// TODO(FUECHR): Implement Auto Go Mem Limit.
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: mounter pods should only be handling a single volume so this should not be a concern. For dynamic mounts the multiple buckets is handled internally by gcsfuse so there is only 1 mount point and 1 directory associated with the mount

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -131,68 +131,99 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
 
 	m.WaitGroup.Add(1)
+
 	go func() {
 		defer m.WaitGroup.Done()
 		if err := cmd.Start(); err != nil {
```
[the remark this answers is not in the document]
chrisThePattyEater: start cleanup monitor should wait for the context to end . If we start it in the go func then if the cmd hangs we won't clean

pkg/csi_driver/node.go
```diff
@@ -865,6 +866,12 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	var profile *profiles.ProfileConfig
 	if profilesEnabled {
 		klog.V(4).Infof("NodeStageVolume gcsfuse profiles feature is enabled for mounter pod %s/%s", podNamespace, podName)
+		if vc == nil {
+			return nil, status.Error(codes.InvalidArgument, "NodeStageVolume VolumeContext must be provided")
+		}
+		if vc[util.VolumeContextKeyPVName] == "" {
+			return nil, status.Error(codes.InvalidArgument, "NodeStageVolume VolumeContext must contain PV name")
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: we only want to block this method if we use profiles

pkg/sidecar_mounter/mounter_server.go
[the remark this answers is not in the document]
uriel-guzman: The recommendation here is to use the CSI Driver's ctx, not the sidecar's ctx. The sidecar's ctx should be only for the gcsfuse mount itself, to keep it alive. It's already explained in the comment that I made above.

pkg/sidecar_mounter/mounter_server.go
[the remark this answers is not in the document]
uriel-guzman: The pattern in the codebase is to return err directly when the underlying function is handling the categorization of the error codes, so please just return it directly:

```
if err := ms.MountToNode(mountCtx, ms.Config); err != nil {
--
return nil, err
}
```

Most of the time, these answers are already present in the Parallelstore CSI Driver codebase for the shared mount, so not sure why we're re-inventing them.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -235,6 +263,79 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
 }
 
+func startCleanupMonitor(ctx context.Context, cmd *exec.Cmd) {
+	// when the ctx.Done() is closed,
+	// the main workload containers have exited,
+	// so it is safe to force kill the gcsfuse process.
+	go func(cmd *exec.Cmd) {
+		<-ctx.Done()
+
+		if cmd.Process == nil {
```
uriel-guzman: Why did you add this? It wasn't present before. This is a new thing you added. Let's please remove this, because this is modifying the current gke-gcsfuse-sidecar logic. Let's please not add any new functionality to the gke-gcsfuse-sidecar without any strong reason since we don't want to to risk making the existing architecture run into new unexpected new behavior.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -235,6 +262,73 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		time.Sleep(time.Second * 5)
+
+		if cmd.ProcessState == nil || !cmd.ProcessState.Exited() {
+			klog.Warningf("after 5 seconds, process with id %v has not exited, force kill the process", cmd.Process.Pid)
+			if err := cmd.Process.Kill(); err != nil {
+				klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
+			}
+		}
+	}()
+}
```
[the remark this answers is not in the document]
uriel-guzman: I don't understand why you added `cmd.Process == nil` either, this bug is going to affect the gke-gcsfuse-sidecar too, so please remove it

pkg/sidecar_mounter/sidecar_mounter.go
uriel-guzman: Gemini also thinks this new check is going to generate a bug for gke-gcsfuse-sidecar, so please remove it: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1474/changes#r3540228948

pkg/csi_driver/node.go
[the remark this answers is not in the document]
uriel-guzman: resolve

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: removed this

pkg/csi_driver/node.go
```diff
@@ -943,7 +950,7 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	}
 
 	// Send GRPC to mounter pod to start GCSFuse.
-	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, args.fuseMountOptions); err != nil {
+	if err := s.mountToNode(ctx, string(pod.UID), stagingPath, volumeID, args.fuseMountOptions); err != nil {
```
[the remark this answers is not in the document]
chrisThePattyEater: fixed

pkg/csi_driver/node.go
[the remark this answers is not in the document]
chrisThePattyEater: fixed

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,94 @@
...
+func (ms *MounterServer) Mount(ctx context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.Internal, "mount request cannot be nil")
+	}
+	if req.GetVolumeId() == "" {
+		return nil, status.Error(codes.Internal, "volume id cannot be empty")
+	}
+	if req.GetMountPoint() == "" {
+		return nil, status.Error(codes.Internal, "mount point cannot be empty")
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: wrong

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,94 @@
...
+		ConfigFile:       filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, "config.yaml"),
+		// TODO(FUECHR): Implement Auto Go Mem Limit.
+	}
+
+	// TODO(FUECHR): Implement Cloud Profiler hook (profiler.Start).
+	// TODO(FUECHR): Implement ErrorWriter.
+	// TODO(FUECHR): Implement defaultingFlagFileParsing.
+
+	// TODO(FUECHR): Make prometheusPort increment inside prepareMountArgs thread-safe.
+	mc.prepareMountArgs()
```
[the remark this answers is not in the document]
chrisThePattyEater: prometheus port will be added in a later pr, also this shouldn't be called concurrently since only the csi driver will call for a single mount operation

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,63 @@ var (
...
+		if err := os.Remove(socketFile); err != nil && !os.IsNotExist(err) {
+			klog.Warningf("failed to remove socket file %q on exit: %v", socketFile, err)
+		}
+	}()
+	socketPath := fmt.Sprintf("unix:%s", socketFile)
+
+	s.Start(socketPath, nil, nil, nil, ms)
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
```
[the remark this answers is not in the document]
chrisThePattyEater: I think we should leave this alone since its copied from the sidecar architecture, unless we plan to also add it to the sidecar

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,63 @@ var (
...
```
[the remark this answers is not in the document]
chrisThePattyEater: as already commented lets leave this alone to stay in sync with sidecar

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,39 @@
...
+type MounterServer struct {
+	mounter.UnimplementedMounterServer
+	Mounter *Mounter
+}
+
+func NewMounterServer(mounter *Mounter) *MounterServer {
+	return &MounterServer{
+		Mounter: mounter,
+	}
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: this will be added as needed

pkg/csi_driver/server.go
```diff
@@ -67,7 +68,7 @@ func (s *nonBlockingGRPCServer) ForceStop() {
 	s.server.Stop()
 }
 
-func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer) {
+func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer, ms mounter.MounterServer) {
```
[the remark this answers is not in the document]
chrisThePattyEater: this is existing code modifying it would modify existing functionality lets leave it be

pkg/sidecar_mounter/mounter_server.go
uriel-guzman: this wasn't resolved, I still see ctx instead of _?

uriel-guzman: same for req, it isn't used so should also be _

pkg/csi_driver/shared_mount.go
```diff
@@ -44,7 +44,7 @@ import (
 const (
 	mounterPodPriorityClass       = "gcsfusecsi-mount-priority"
 	mounterPodMountDir            = "mount-dir"
-	mounterPodSocketFile          = "mounter.sock"
+	MounterPodSocketFile          = "mounter.sock"
```
uriel-guzman: why is this exported?I don't see it being referenced outside of the package?

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,58 @@ var (
...
+
+	if err := os.Remove(socketFile); err != nil && !os.IsNotExist(err) {
+		klog.Warningf("failed to remove stale socket file %q: %v", socketFile, err)
+	}
+	socketPath := fmt.Sprintf("unix:%s", socketFile)
+
+	s.Start(socketPath, nil, nil, nil, ms)
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
```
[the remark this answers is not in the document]
uriel-guzman: I think this improvement is valid, please address.

cmd/sidecar_mounter/main.go
```diff
@@ -49,24 +49,58 @@ var (
...
+	}
+	socketPath := fmt.Sprintf("unix://%s", socketFile)
+
+	s.Start(socketPath, nil, nil, nil, ms)
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
+	klog.Info("Waiting for SIGTERM signal...")
+	<-c
+	klog.Info("Received SIGTERM signal, waiting for the gcsfuse process to exit...")
```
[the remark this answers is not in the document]
uriel-guzman: I think this improvement is valid, please address.

uriel-guzman (changes requested):

cmd/sidecar_mounter/main.go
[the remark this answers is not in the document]
uriel-guzman: I think this is a valid improvement please address.

PR #1477, opened 2026-06-23
Optimize PV lookup in shared node mount by injecting PV name in webhook

yaozile123:
**What type of PR is this?**

/kind feature

This PR optimizes PV lookup in the CSI controller for shared node mounts and integrates a new feature flag for node driver and controller.

### Key Changes
1. **Optimize PV Lookup in ControllerPublishVolume**: 
   - The mutating webhook now injects the PV name (`csi.storage.k8s.io/pv/name`) into the CSI `VolumeAttributes` during PV admission (`pkg/webhook/mutatingwebhook.go`).
   - The controller retrieves this name to perform a direct `GetPV()` lookup in `ControllerPublishVolume` (`pkg/csi_driver/controller.go`), replacing the previous $O(N)$ full-cluster PV scan.
2. **Add Feature Flag & Resource Management**: 
   - Introduced `--enable-shared-mount` flag to node and controller.
   - Informer listers for PV, PVC, Pod, and PodTemplate are only registered and started if the flag is enabled (`cmd/csi_driver/main.go`), reducing resource overhead.
   - Updated overlays (`deploy/overlays/shared-mount/`) and unit tests to align with the feature flag.
3. **Informer Memory Optimization**: 
   - Updated PV/PVC informer trim functions to only preserve fields required for shared mounts: `ClaimRef` on PV, and only the `mounter-pod-template` annotation on PVC (`pkg/cloud_provider/clientset/clientset.go`).

Fixes #

I did a poc locally and verified the optimization works

```release-note

### CODE REVIEW

File: pkg/cloud_provider/clientset/clientset.go [hunk 1]
```diff
@@ -187,10 +188,19 @@ func (c *Clientset) ConfigurePVCLister(ctx context.Context) {
...
 			return obj, nil
 		}
+		var annotations map[string]string
+		if pvcObj.ObjectMeta.Annotations != nil {
+			if val, ok := pvcObj.ObjectMeta.Annotations[webhook.MounterPodTemplateAnnotation]; ok {
+				annotations = map[string]string{
+					webhook.MounterPodTemplateAnnotation: val,
+				}
+			}
+		}
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -187,10 +188,17 @@ func (c *Clientset) ConfigurePVCLister(ctx context.Context) {
 		if !ok {
 			return obj, nil
 		}
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/csi_driver/node.go
```diff
@@ -653,6 +653,9 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 	if !sharedMount(req.GetVolumeContext()) {
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
+	if s.driver.config.FeatureOptions.SharedMountOptions == nil || !s.driver.config.FeatureOptions.SharedMountOptions.Enabled {
+		return nil, status.Error(codes.FailedPrecondition, "shared mount feature is disabled")
```
chrisThePattyEater: why not just add stage unstage to the same if block as csc = append(csc, csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME)

that way if the feature is not enabled then nodestagevolume will never get called

yaozile123: You are right, let me fix this as well as cpv

yaozile123: I have removed the check from CPV and NSV, and I've updated the driver to conditionally register the STAGE_UNSTAGE_VOLUME node capability based on the feature flag

deploy/overlays/shared-mount/enable_shared_mount_patch_controller.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "op": "add",
```
chrisThePattyEater, line 3: Can't this go into controller patch deploy/overlays/shared-mount/controller_patch.yaml

yaozile123: We use JSON patch here to avoid duplicating all base arguments in  controller_patch.yaml , as strategic merge replaces lists instead of appending.

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -193,13 +194,20 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 func (c *Clientset) ConfigurePVCLister(ctx context.Context) {
 	trim := func(obj any) (any, error) {
 		pvcObj, ok := obj.(*corev1.PersistentVolumeClaim)
-		if !ok {
+		if !ok || pvcObj == nil {
 			return obj, nil
 		}
+		var annotations map[string]string
+		if val, ok := pvcObj.ObjectMeta.Annotations[webhook.MounterPodTemplateAnnotation]; ok {
```
uriel-guzman, line 201: Can annotations map be nil? Otherwise, please check for nil existence before checking the value.

pkg/csi_driver/controller.go
```diff
@@ -153,17 +153,20 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
 
 	// Skip ControllerPublishVolume if the volume is not using the shared mount feature.
 	vc := req.GetVolumeContext()
+	klog.Infof("ControllerPublishVolume: checking shared mount for volume %q, context: %v", volumeID, vc)
```
uriel-guzman: I think we can get rid of this log to reduce log noise, or at least reduce verbosity to 6. If you decide to keep it, put it after L160, after we have return no-op success for non-shared-mount volumes.

pkg/csi_driver/controller.go
```diff
@@ -153,17 +153,20 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
...
 	// Find the workload namespace where the mounter pod should be created by identifying the PVC
 	// bound to this volume's PV.
 	clientset := s.driver.config.K8sClients
-	// TODO(urielguzman): This implementation requires PVs from the informer's cache, which can
-	// unbounded. When manually testing the feature, we should add a volumeHandle indexer,
-	// so that the lookup can be O(1).
-	pv, err := pvFromVolumeID(clientset, volumeID)
+	pvName := vc[util.VolumeContextKeyPVName]
+	if pvName == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "PV name not found in VolumeContext (key %q)", util.VolumeContextKeyPVName)
```
uriel-guzman: This should return Internal error, because if it's missing, it's out fault, not the customer's.

pkg/csi_driver/controller.go
```diff
@@ -153,17 +153,20 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
...
+	}
+	klog.Infof("ControllerPublishVolume: directly fetching PV %q from VolumeContext for volume %q, context: %v", pvName, volumeID, vc)
```
uriel-guzman: This log can also either be removed or reduced verbosity to 6.

pkg/webhook/mutatingwebhook.go
```diff
@@ -108,11 +108,12 @@ func (si *SidecarInjector) Handle(ctx context.Context, req admission.Request) ad
 
 		// Patch the label to PV if it uses shared node mount.
 		if si.Config.EnableSharedNodeMount && pv.Spec.CSI.VolumeAttributes != nil && pv.Spec.CSI.VolumeAttributes[SharedMountVolumeAttribute] == util.TrueStr {
-			if pv.Labels == nil {
-				pv.Labels = make(map[string]string)
-			}
-			if pv.Labels[SharedMountLabel] != util.TrueStr {
+			if pv.Labels[SharedMountLabel] != util.TrueStr || pv.Spec.CSI.VolumeAttributes[util.VolumeContextKeyPVName] != pv.Name {
```
uriel-guzman, line 111: You are accessing pv.Labels before checking if it's nil, this could lead to a panic.

pkg/cloud_provider/clientset/clientset.go
yaozile123: I remembered in Go we can still check the map even if it is nil, and it will just return false for `ok`

pkg/webhook/mutatingwebhook.go
yaozile123: I think it won't panic in go. It will simply returns the zero value of the value type which is an empty string in this case

pkg/csi_driver/controller.go
yaozile123: Done

pkg/csi_driver/controller.go
yaozile123: Done

pkg/csi_driver/controller.go
yaozile123: Done

uriel-guzman (approved):

PR #1478, opened 2026-06-23
Revert "e2e: add dual CSI volume tests for GCSFuse and PD CSI combination"

uriel-guzman:
Reverts GoogleCloudPlatform/gcs-fuse-csi-driver#1458 since tests started failing after this


amacaskill (approved):

mattcary:
Seems to be the logic around line 92 in e2e_test.go.

@samyukthapa

PR #1490, opened 2026-06-26
Add profiles support for shared mounts

yaozile123:
**What type of PR is this?**
/kind feature

This PR introduces Storage Profiles support for the Shared Mount architecture. Previously, profile recommendations were only supported in the Sidecar architecture. 

Key changes include:
- **NodeStageVolume Integration**: Invokes the `BuildProfileConfig` recommendation engine during `executeNodeStageVolume` to calculate and apply optimal `file-cache` and `stat-cache` configurations for the shared Mounter Pod.
- **Mounter Pod Provisioning**: Automatically mounts the required placeholder volumes to the Mounter Pod based on the profile engine's recommended file cache medium.
- **Shared Parsing Logic**: Refactored `parseRequestArguments` to decouple from `NodePublishVolumeRequest`. This allows both `NodeStageVolume` and `NodePublishVolume` to use the exact same logic for extracting mount options and volume attributes.
- **Recommendation Log Target**: Added a `containerName` field to the recommendation insights JSON payload (`GCSFuseCSIRecommendationLog`) to clearly distinguish resource tracking between `gcsfusecsi-mount` (shared mount) and `gke-gcsfuse-sidecar` (sidecar mode).

Fixes #

Verified end-to-end on local cluster by checking the logs in the pantheon for shared mount PVs.

```release-note
Add Storage Profiles recommendation support for shared mount volumes.

### CODE REVIEW

File: pkg/csi_driver/node.go [hunk 1]
```diff
@@ -104,6 +104,12 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Get the Pod object.
 	vc := req.GetVolumeContext()
+	isSharedMount := sharedMount(vc)
+	stagingPath := req.GetStagingTargetPath()
+	if isSharedMount && len(stagingPath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Staging Target Path must be provided for shared mount")
+	}
```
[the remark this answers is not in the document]
yaozile123: The bind-mount will be implemented in another PR

pkg/profiles/recommender.go
```diff
@@ -252,16 +262,25 @@ func BuildProfileConfig(params *BuildProfileConfigParams) (*ProfileConfig, error
...
+			return nil, status.Errorf(codes.Internal, "failed to get pod %s/%s: %v", params.PodNamespace, params.PodName, err)
 		}
-		return nil, status.Errorf(codes.Internal, "failed to get pod %s/%s: %v", params.PodNamespace, params.PodName, err)
+		pod = p
+	}
+
+	containerName := params.ContainerName
+	if containerName == "" {
+		containerName = webhook.GcsFuseSidecarName
 	}
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/csi_driver/shared_mount.go
```diff
@@ -361,8 +367,24 @@ func mounterPodVolumes(config *mounterPodConfig) []corev1.Volume {
...
+			corev1.Volume{
+				Name: webhook.SidecarContainerFileCacheRamDiskVolumeName,
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{
+						Medium: corev1.StorageMediumMemory,
+					},
+				},
+			},
+		)
+	}
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/csi_driver/controller.go
```diff
@@ -223,6 +224,19 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
...
 	}
 
+	var profilesEnabled bool
+	if s.features != nil && s.features.FeatureGCSFuseProfiles.Enabled && pv.Spec.StorageClassName != "" {
+		sc, err := clientset.GetSC(pv.Spec.StorageClassName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return nil, status.Errorf(codes.Internal, "failed to get StorageClass %q: %v", pv.Spec.StorageClassName, err)
+			}
+			klog.Warningf("failed to determine if StorageClass %q uses profiles: not found", pv.Spec.StorageClassName)
```
uriel-guzman: This will happen every time a customer isn't using profiles, because they can specify dummy storage classes. K8S doesn't enforce the storage classes to exist between PV and PVC, but it does enforce the PV spec's storageclass field isn't empty (it can be empty stirng though), and are the same.

Instead, I'd downgrade this log to a V6. If you got this code from the original profiles logic, it should also be V6 over there too. This will help preventing warning spams to customers that aren't relevant.

pkg/csi_driver/node.go
```diff
@@ -723,6 +727,42 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
+	// Validate arguments
+	args, err := parseRequestArguments(req.GetVolumeId(), false, req.GetVolumeCapability(), req.GetVolumeContext())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse volume attributes: %v", err)
```
uriel-guzman: This says "failed to parse volume attributes" but what's being validated are request arguments, not volume attributes.

pkg/csi_driver/node.go
```diff
@@ -723,6 +727,42 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+			ContainerName:       util.MounterPodNamePrefix,
+			IsInitContainer:     false,
+			VolumeAttributeKeys: transformKeysToSet(volumeAttributesToMountOptionsMapping),
+			PodNamespace:        podNamespace,
+			PodName:             podName,
+			NodeName:            s.driver.config.NodeID,
+			Pod:                 pod,
+		})
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to build profile config: %v", err)
```
uriel-guzman: This is going to transform all errors returned by `BuildProfileConfig` as internal an alert us every time. If you look at the implementation of `BuildProfileConfig`, it returns a mixture of error codes, so we don't want this: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/recommender.go#L230

Instead, directly return the error without wrapping it in an error code, like in: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L129

pkg/csi_driver/node.go
```diff
@@ -723,6 +727,42 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+		}
+		if profile != nil {
+			// Merge recommended mount options
+			// TODO(yaozile): pass recommended mount options into the gRPC MountRequest when the mounter client is implemented
+			args.fuseMountOptions, err = profile.MergeRecommendedMountOptionsOnMissingKeys(args.fuseMountOptions)
```
uriel-guzman: I think we're missing `MergeVolumeAttribtuesOnRecommendedMissingKeys?`: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L133

I think you accidentally used `MergeRecommendedMountOptionsOnMissingKeys`. This part should use `MergeVolumeAttribtuesOnRecommendedMissingKeys`, and the returned VC should be the one being passed to `parseRequestArguments`. 

Take a look at https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L138, the `parseRequestArguments` should be called after the VC was already mixed with the StorageClass's parameters, not before, like you have now.

Also, I assume tests pass because this isn't being validated, after you make that change, please add unit tests so that you confirm that it 1. Fails on the current version, then when you fix, 2. it should pass.

pkg/csi_driver/node.go
```diff
@@ -723,6 +727,42 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to recommend mount options: %v", err)
+			}
+		}
+	}
+
+	klog.V(6).Infof("NodeStageVolume mount options about to be passed for volume %q: %v", req.GetVolumeId(), args.fuseMountOptions)
```
uriel-guzman: Add a TODO here to remember to remove this once we actually pass them to the gRPC server.

pkg/csi_driver/shared_mount.go
```diff
@@ -361,8 +367,24 @@ func mounterPodVolumes(config *mounterPodConfig) []corev1.Volume {
 			},
 		}})
 
-	// TODO(urielguzman): Add profiles and host network volumes when those features are implemnented for
```
uriel-guzman, line 364: This TODO should stay, but should just say "host network" now, because we're not adding that yet.

pkg/csi_driver/node.go

pkg/csi_driver/controller.go
yaozile123: Done

pkg/csi_driver/shared_mount.go
yaozile123: Done

pkg/csi_driver/node.go

pkg/csi_driver/shared_mount.go
```diff
@@ -361,8 +367,24 @@ func mounterPodVolumes(config *mounterPodConfig) []corev1.Volume {
...
-	// shared mount.
+	if config.profilesEnabled {
+		if !webhook.VolumeExists(volumes, webhook.SidecarContainerFileCacheEphemeralDiskVolumeName) {
+			volumes = append(volumes, corev1.Volume{
```
uriel-guzman: These two volumes are the same as https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/mutatingwebhook.go#L454

Can we just make both of them into one single global variable like these: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/webhook/sidecar_spec.go#L68, and refactor so that both files reference the same var? This eliminates the chance of both volumes being different accidentally.

Alternatively, we could make a helper function in webhook package called "appendProfilesVolumes" and "appendProfilesVolumeMounts" and just call that from here so we remove duplciate code.

pkg/csi_driver/node.go

pkg/profiles/recommender.go
```diff
@@ -182,10 +186,16 @@ func (c *ProfileConfig) MergeVolumeAttributesOnRecommendedMissingKeys(input map[
 // volumeAttributeKeys is used to filter parameters from the StorageClass that should be
 // treated as volume attributes.
 func BuildProfileConfig(params *BuildProfileConfigParams) (*ProfileConfig, error) {
-	// Get the PV name from target path.
-	_, volumeName, err := util.ParsePodIDVolumeFromTargetpath(params.TargetPath)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get pv name from target path %q: %v", params.TargetPath, err)
+	var volumeName string
+	if params.VolumeName != "" {
+		volumeName = params.VolumeName
```
uriel-guzman: If `VolumeName` is empty and is called from `NodeStageVolume`, then L195 will return an Internal error which doesn't make a lot of sense to `NodeStageVolume` (i.e. failed to get pv name from target path), which would also be empty.

Instead, I think you should check:
1. If `params.TargetPath` is not empty, set `VolumeName` to the volume parsed from that.
2. `VolumeName` now either holds a) the volume populated from step #1, or b) a value inserted by `NodeStageVolume`. In either of these cases, if this is empty, it's an `Internal` error, so we can just fail with `volume name can't be empty`.

pkg/profiles/recommender.go
```diff
@@ -252,16 +262,28 @@ func BuildProfileConfig(params *BuildProfileConfigParams) (*ProfileConfig, error
 	}
 
 	// Get the Pod object using the Pod namespace/name.
-	pod, err := params.Clientset.GetPod(params.PodNamespace, params.PodName)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil, status.Errorf(codes.NotFound, "pod %s/%s not found: %v", params.PodNamespace, params.PodName, err)
+	pod := params.Pod
```
uriel-guzman: I am confused as to why we are passing pod here, if pod is obtained with `params.Clientset.GetPod` in L267?

pkg/profiles/recommender.go
```diff
@@ -252,16 +262,28 @@ func BuildProfileConfig(params *BuildProfileConfigParams) (*ProfileConfig, error
...
+			}
+			return nil, status.Errorf(codes.Internal, "failed to get pod %s/%s: %v", params.PodNamespace, params.PodName, err)
 		}
-		return nil, status.Errorf(codes.Internal, "failed to get pod %s/%s: %v", params.PodNamespace, params.PodName, err)
+		pod = p
+	}
+
+	containerName := params.ContainerName
+	if containerName == "" {
+		containerName = webhook.GcsFuseSidecarName
```
uriel-guzman: Aren't we passing `containerName` from `NodePublishVolume` and populating it with `GcsFuseSidecarName`, and with mounter pod prefix in `NodeStageVolume`? This check seems redundant. It also doesn't seem correct, because if `containerName` is missing, and we're using shared mount, it's going to set gke-gcsfuse-sidecar. I'd just remove this and use whatever value we populated before.

test/e2e/utils/utils_test.go
```diff
@@ -178,6 +178,7 @@ func TestParseConfigFlags(t *testing.T) {
 					FileCacheCapacity: "-1Mi",
 					ReadOnly:          false,
 					LogSeverity:       "info",
+					BillingProject:    testProject,
```
uriel-guzman, line 181: What's that for?

pkg/profiles/recommender.go
```diff
@@ -108,12 +108,13 @@ type nodeDetails struct {
...
 type podDetails struct {
-	sidecarLimits *parsedResourceList
-	namespace     string
-	name          string // The name of the Pod.
-	labels        map[string]string
+	containerLimits *parsedResourceList
+	namespace       string
+	name            string // The name of the Pod.
+	containerName   string // The name of the container (sidecar or mounter).
+	labels          map[string]string
```
uriel-guzman, line 117: I think we're not handling this. In https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/recommender.go#L283, if the cache is created by the user, then the idea is to disable smart caching, because customer already overrode it: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/recommender.go#L573

However, in `createMounterPodSpec`, the label isn't ever applied, so this will not work: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/shared_mount.go#L131

I think we should add logic there to detect a manual cache override and just apply the mounter pod with the label. That's the easiest way to fix this. 

Also please add unit tests to validate this, since I assume tests are also passing because it's not tested.

uriel-guzman (changes requested):

test/e2e/utils/utils_test.go
yaozile123: That's for the [change](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/commit/951a8a6bf78079daa49109ee009f843bb5b7441a) that GCSFuse team made for requester_pay bucket gcsfuse integration tests. We are now [parsing and exporting billing project](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/39f676bb19b38a4d887f7ad6a8263b348749b790/test/e2e/utils/utils.go#L357-L364) for that test package but we forgot to update the unit tests.

pkg/csi_driver/shared_mount.go
yaozile123: Done, I have make both of them as global variable

pkg/profiles/recommender.go
yaozile123: Done

pkg/profiles/recommender.go
yaozile123: We pass the pod directly saves an extra API call since it's already fetched by the caller. The Clientset.GetPod here is just a fallback, let me know if you prefer I remove the fallback entirely

pkg/profiles/recommender.go
yaozile123: I removed the redundant fallback and replace it with a validation error

pkg/profiles/recommender.go
yaozile123: Done

pkg/csi_driver/node.go
yaozile123: I have fixed fix the logic and added the unit tests by mocking the mounter server.

pkg/profiles/recommender.go
uriel-guzman: Can we just pass the pod from the NodePublish and return `Internal` if it's nil? I think that's simpler. We don't need the extra fallback if we fail early.

pkg/csi_driver/controller_test.go
```diff
@@ -662,11 +668,56 @@ func TestControllerPublishVolume(t *testing.T) {
...
+				}
+			},
+			enableProfiles: true,
+		},
 	}
 	for _, test := range cases {
 		t.Run(test.name, func(t *testing.T) {
 			fc := test.setupFake()
-			s := initTestController(t, fc)
+			s, driver := initTestController(t, fc)
```
uriel-guzman: Wht do we need to return driver here? It seems like it's just to mutate `FeatureGCSFuseProfiles.Enabled?`. I think we can just always set that to true and avoid returning the extra driver object to simplify. Usually in unit tests the feature flags are all always enabled.

pkg/csi_driver/node.go
```diff
@@ -114,16 +118,18 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
 		profile, err = profiles.BuildProfileConfig(&profiles.BuildProfileConfigParams{
 			TargetPath:          targetPath,
+			ContainerName:       webhook.GcsFuseSidecarName,
 			Clientset:           s.k8sClients,
 			VolumeAttributeKeys: transformKeysToSet(volumeAttributesToMountOptionsMapping),
 			PodNamespace:        pod.Namespace,
 			PodName:             pod.Name,
 			NodeName:            s.driver.config.NodeID,
 			IsInitContainer:     isInitContainer,
+			Pod:                 pod,
```
uriel-guzman, line 175: Either remove redundant pod object, or always pass it and stop calling GET in `BuildProfileConfig`. I think I'd prefer passing it from here and stop doing the GET over there, because we already have it here.

pkg/csi_driver/node.go
```diff
@@ -723,6 +729,47 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+	// Build profile config and merge recommended mount options if storage profiles are enabled.
+	profilesEnabled := s.driver.config.FeatureOptions.FeatureGCSFuseProfiles.Enabled
+	var profile *profiles.ProfileConfig
+	if profilesEnabled {
+		klog.V(4).Infof("NodeStageVolume gcsfuse profiles feature is enabled for mounter pod %s/%s", podNamespace, podName)
+		profile, err = profiles.BuildProfileConfig(&profiles.BuildProfileConfigParams{
+			VolumeName:          vc[util.VolumeContextKeyPVName],
+			Clientset:           s.k8sClients,
+			ContainerName:       util.MounterPodNamePrefix,
+			IsInitContainer:     false,
```
uriel-guzman: Isn't this `false` by default?

pkg/csi_driver/node.go
```diff
@@ -135,7 +141,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	// Validate arguments
-	args, err := parseRequestArguments(req, vc)
+	args, err := parseRequestArguments(req.GetVolumeId(), req.GetReadonly(), req.GetVolumeCapability(), vc)
```
uriel-guzman, line 187: The `vc` you are validating here doesn't yet have the fields introduced by `MergeVolumeAttributesOnRecommendedMissingKeys`, so this isn't right. See https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L138, `vc` is passed to `parseRequestArgument` **after**.

pkg/csi_driver/node.go
```diff
@@ -723,6 +729,47 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+	// Validate arguments
+	args, err := parseRequestArguments(volumeID, false, req.GetVolumeCapability(), vc)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse request arguments: %v", err)
+	}
+
+	if profile != nil {
+		args.fuseMountOptions, err = profile.MergeRecommendedMountOptionsOnMissingKeys(args.fuseMountOptions)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to merge recommended mount options: %v", err)
```
uriel-guzman: This is incorrectly turning all of the error codes contained inside `MergeRecommendedMountOptionsOnMissingKeys` into `Internal`: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/recommender.go#L443

uriel-guzman (changes requested):

pkg/profiles/recommender.go
yaozile123: Done, I also add a test case to make sure `BuildProfileConfig` should return error if pod is nil.

pkg/csi_driver/node.go
yaozile123: Done, I have removed the unused field PodName and PodNamespace here as well.

pkg/csi_driver/controller_test.go
yaozile123: Done, I have enabled profiles by default in `initTestController`

pkg/csi_driver/node.go
yaozile123: Thanks for catching that, I have removed it from the config params

pkg/csi_driver/node.go

pkg/csi_driver/node.go
yaozile123: I called `MergeVolumeAttributesOnRecommendedMissingKeys` in line 137, so the vc I passed to `parseRequestArgument` should be right

pkg/csi_driver/node.go
```diff
@@ -104,6 +106,8 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Get the Pod object.
 	vc := req.GetVolumeContext()
+	isSharedMount := sharedMount(vc)
```
uriel-guzman: Nit: This is assigned to a variable, but it doesn't seem to fulfill any readability or efficiency,purposes. I'd just use `sharedMount(vc)` directly, instead of assigning it to a longer variable name `isSharedMount`, and save 1 line.

pkg/csi_driver/node.go
uriel-guzman: Sorry, I think I made the comment in the wrong place. I was actually referring to L756 in NodeStageVolume.

uriel-guzman: ignore this comment.

uriel-guzman (changes requested):

pkg/csi_driver/node.go
yaozile123: I rebased on your [PR](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1495). The NPV will call  NodePublishVolumeForSharedMount so we didn't need this variable anymore.

uriel-guzman (approved):

PR #1495, opened 2026-06-30
Implement NodePublishVolume for shared mount

uriel-guzman:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Impelement NodePublishVolume for shared node mount. This PR verifies basic arguments and verifies the mounter pod exists. If it does, it checks for its status. If it's non-running, it returns an error. This PR also allows NodePublishVolume to be continue the original code execution plan when shared mount is not enabled.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -90,6 +90,48 @@ func (s *nodeServer) NodeGetCapabilities(_ context.Context, _ *csi.NodeGetCapabi
 	}, nil
 }
 
+func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
+	// Validate arguments
+	stagingPath := req.GetStagingTargetPath()
+	if len(stagingPath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Staging Target Path must be provided")
+	}
+	publishContext := req.GetPublishContext()
+	if publishContext == nil {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Publish Context must be provided")
+	}
+	mounterPodNamespace := publishContext[PublishContextKeyMounterPodNamespace]
+	if len(mounterPodNamespace) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Mounter Pod Namespace must be provided")
+	}
+	mounterPodName := publishContext[PublishContextKeyMounterPodName]
+	if len(mounterPodName) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Mounter Pod Name must be provided")
+	}
+
+	// Verify mounter pod exists.
+	mounterPod, err := s.k8sClients.GetPod(mounterPodNamespace, mounterPodName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.FailedPrecondition, "failed to get mounter pod: %v", err)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get mounter pod: %v", err)
+	}
+	if mounterPod == nil {
+		return nil, status.Errorf(codes.Internal, "mounter pod %s/%s cannot be nil", mounterPodNamespace, mounterPodName)
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/node.go
```diff
@@ -90,6 +90,52 @@ func (s *nodeServer) NodeGetCapabilities(_ context.Context, _ *csi.NodeGetCapabi
 	}, nil
 }
 
+func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
+	// Validate arguments
+	stagingPath := req.GetStagingTargetPath()
+	if len(stagingPath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Staging Target Path must be provided")
+	}
+	targetPath := req.GetTargetPath()
```
yaozile123: I think we already check targetPath at the [beginning of the NPV](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/cde26084f3a704b0358d6a9296ff0785b8fdac09/pkg/csi_driver/node.go#L100-L103), so this seems redundant

pkg/csi_driver/node_test.go
```diff
@@ -228,7 +230,7 @@ func TestNodePublishVolume(t *testing.T) {
 func TestExecuteNodeStageVolume(t *testing.T) {
 	t.Parallel()
 
-	stagingPath, cleanup := setupMountTarget(t)
+	stagingPath, cleanup := setupTestTargetPath(t)
```
yaozile123: Should we call setupTestStagingPath() at here?

pkg/csi_driver/node_test.go
```diff
@@ -799,7 +801,7 @@ func TestNodeUnpublishVolume(t *testing.T) {
 func TestNodeStageVolume(t *testing.T) {
 	t.Parallel()
 
-	stagingPath, cleanup := setupMountTarget(t)
+	stagingPath, cleanup := setupTestTargetPath(t)
```
yaozile123: Same here, I think we should call setupTestStagingPath() instead

pkg/csi_driver/node_test.go
```diff
@@ -1003,7 +1005,7 @@ func TestNodeStageVolume(t *testing.T) {
 
 func TestNodeUnstageVolume(t *testing.T) {
 	t.Parallel()
-	stagingPath, cleanup := setupMountTarget(t)
+	stagingPath, cleanup := setupTestTargetPath(t)
```
yaozile123: Please use setupTestStagingPath() instead

pkg/csi_driver/node_test.go
```diff
@@ -107,27 +107,29 @@ func initTestNodeServerWithCustomClientset(t *testing.T, clientSet *clientset.Fa
...
-func setupMountTarget(t *testing.T) (string, func()) {
-	t.Helper()
-	defaultPerm := os.FileMode(0o750) + os.ModeDir
-	// Setup mount target path
-	tmpDir := "/tmp/var/lib/kubelet/pods/test-pod-id/volumes/kubernetes.io~csi/"
-	if err := os.MkdirAll(tmpDir, defaultPerm); err != nil {
-		t.Fatalf("failed to setup tmp dir path: %v", err)
-	}
-	base, err := os.MkdirTemp(tmpDir, "node-publish-")
+func setupTestDir(t *testing.T, path string) (string, func()) {
```
yaozile123, line 110: Nit: Please add t.Helper() back so failure line numbers attribute to the caller test functions instead of setupTestDir() itself.

pkg/csi_driver/node.go
uriel-guzman: done

pkg/csi_driver/node_test.go
uriel-guzman: done

pkg/csi_driver/node_test.go
uriel-guzman: done

pkg/csi_driver/node_test.go
uriel-guzman: done

pkg/csi_driver/node_test.go
uriel-guzman: done

yaozile123 (approved):

PR #1501, opened 2026-07-01
Kernel Parameters Integration for shared mount

yaozile123:
**What type of PR is this?**
/kind feature

Integrates GCSFuse Kernel Parameters support for Shared Node Mount mode (`NodeStageVolume` and `NodeUnstageVolume`):
- Passes `--enable-gcsfuse-kernel-params=true` FUSE mount option in `NodeStageVolume` when the feature flag is enabled and the Mounter Pod image version supports it.
- Refactors `MonitorKernelParamsFile` to support shared node mount.
- Starts asynchronous kernel parameter monitoring in `NodeStageVolume` and cleans up the monitoring goroutine in `NodeUnstageVolume`.
- Adds unit test coverage for helper functions and `NodeStageVolume` for kernel parameter.

Fixes #

NONE

```release-note
NONE

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -271,13 +271,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Use target path as an volume identifier because it corresponds to Pods and volumes.
 	// Pods may belong to different namespaces and would need their own bucket access check.
-	// We initialize volumeState iff bucket name is NOT "_", I.e. It's not a dynamic mount.
+	// We initialize volumeState if bucket name is NOT "_", I.e. It's not a dynamic mount.
 	var vs *util.VolumeState
 	if args.bucketName != "_" {
 		var ok bool
 		if vs, ok = s.volumeStateStore.Load(targetPath); !ok {
-			s.volumeStateStore.Store(targetPath, &util.VolumeState{})
-			vs, _ = s.volumeStateStore.Load(targetPath)
+			vs = &util.VolumeState{}
+			s.volumeStateStore.Store(targetPath, vs)
 		}
 	}
 	// volumeState is safe to access if its not nil for remaining of function since volumeLock prevents
@@ -442,6 +442,16 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 	}
 
+	var podUID, volumeName, emptyDirBasePath string
+	if s.isGcsFuseKernelParamsFeatureSupported(gcsFuseSidecarImage, vs) {
+		var err error
+		if podUID, volumeName, err = util.ParsePodIDVolumeFromTargetpath(targetPath); err != nil {
+			klog.Warningf("Failed to parse pod ID and volume name from target path %q for kernel params monitor: %v. Monitor will not be started.", targetPath, err)
+		} else if emptyDirBasePath, err = util.PrepareEmptyDir(targetPath, false); err != nil {
+			klog.Warningf("Failed to prepare empty dir from target path %q for kernel params monitor: %v. Monitor will not be started.", targetPath, err)
+		}
+	}
+
 	// Check if the target path is already mounted
 	mounted, err := s.isDirMounted(targetPath)
 	if err != nil {
@@ -474,7 +484,9 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		// TODO: Check if the socket listener timed out
 
 		// Restart monitoring goroutine if the driver restarts for existing mounts.
-		s.startGcsFuseKernelParamsMonitoring(targetPath, gcsFuseSidecarImage, vs)
+		if emptyDirBasePath != "" {
+			s.startGcsFuseKernelParamsMonitoring(targetPath, emptyDirBasePath, podUID, volumeName, gcsFuseSidecarImage, vs)
+		}
 
 		klog.V(4).Infof("NodePublishVolume succeeded on volume %q to target path %q, mount already exists.", args.bucketName, targetPath)
 
@@ -521,34 +533,38 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	// Start monitoring goroutine for new mounts.
-	s.startGcsFuseKernelParamsMonitoring(targetPath, gcsFuseSidecarImage, vs)
+	if emptyDirBasePath != "" {
+		s.startGcsFuseKernelParamsMonitoring(targetPath, emptyDirBasePath, podUID, volumeName, gcsFuseSidecarImage, vs)
+	}
 	klog.V(4).Infof("NodePublishVolume succeeded on volume %q to target path %q", args.bucketName, targetPath)
 
 	return &csi.NodePublishVolumeResponse{}, nil
 }
 
 // startGcsFuseKernelParamsMonitoring starts a single long lived goroutine to monitor the
-// GCSFuse kernel parameters file per target path if the feature is enabled and supported.
-// It uses sync.Once to ensure the monitor is started only once. This monitoring goroutine is
-// stopped in the NodeUnpublishVolume operation.
-func (s *nodeServer) startGcsFuseKernelParamsMonitoring(targetPath, gcsFuseSidecarImage string, vs *util.VolumeState) {
-	if s.isGcsFuseKernelParamsFeatureSupported(gcsFuseSidecarImage, vs) {
+// GCSFuse kernel parameters file per mount point if the feature is enabled and supported.
+// It uses sync.Once to ensure the monitor is started only once.
+func (s *nodeServer) startGcsFuseKernelParamsMonitoring(mountPoint, emptyDirBasePath, podUID, volumeName, gcsFuseContainerImage string, vs *util.VolumeState) {
+	if s.isGcsFuseKernelParamsFeatureSupported(gcsFuseContainerImage, vs) {
 		vs.GCSFuseKernelMonitorState.StartKernelParamsFileMonitorOnce.Do(func() {
 			ctx, cancel := context.WithCancel(context.Background())
 			vs.GCSFuseKernelMonitorState.CancelFunc = cancel
+			logPrefix := fmt.Sprintf("Kernel Params Monitor[Pod %v, Volume %v]", podUID, volumeName)
+
 			// Fire kernel params monitoring goroutine.
-			go util.MonitorKernelParamsFile(ctx, targetPath, GCSFuseKernelParamsFilePollInterval)
+			go util.MonitorKernelParamsFile(ctx, mountPoint, emptyDirBasePath, logPrefix, GCSFuseKernelParamsFilePollInterval)
 		})
 	}
 }
 
-// isGcsFuseKernelParamsFeatureSupported returns true if the GCSFuse kernel parameters feature is enabled and supported by sidecar version.
+// isGcsFuseKernelParamsFeatureSupported returns true if the GCSFuse kernel parameters feature is enabled and supported by sidecar/mounter version.
 // GCSFuse Kernel Params feature is not applicable for dynamic mounts as a single kernel parameter setting doesn’t apply for different type
 // of buckets that a dynamic mount serves. This feature is only applicable for non-dynamic mounts i.e when volumeState(vs) is not nil.
-func (s *nodeServer) isGcsFuseKernelParamsFeatureSupported(gcsFuseSidecarImage string, vs *util.VolumeState) bool {
+func (s *nodeServer) isGcsFuseKernelParamsFeatureSupported(gcsFuseContainerImage string, vs *util.VolumeState) bool {
 	return vs != nil &&
+		s.driver.config.FeatureOptions != nil &&
 		s.driver.config.FeatureOptions.EnableGCSFuseKernelParams &&
-		s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, GCSFuseKernelParamsMinVersion)
+		s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseContainerImage, GCSFuseKernelParamsMinVersion)
 }
 
 func (s *nodeServer) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
@@ -571,7 +587,7 @@ func (s *nodeServer) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpubli
 	}
 
 	// Stop GCSFuse Kernel Params monitoring.
-	if vs, ok := s.volumeStateStore.Load(targetPath); ok {
+	if vs, ok := s.volumeStateStore.Load(targetPath); ok && vs != nil {
 		if vs.GCSFuseKernelMonitorState.CancelFunc != nil {
 			vs.GCSFuseKernelMonitorState.CancelFunc()
 		}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: pkg/csi_driver/node_test.go [hunk 1]
```diff
@@ -2235,3 +2237,113 @@ func TestNodePublishVolumeForSharedMount(t *testing.T) {
...
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			testStagingPath, cleanupStaging := setupTestStagingPath(t)
+			defer cleanupStaging()
+
+			tmpDir, err := os.MkdirTemp("/tmp", "s")
+			if err != nil {
+				t.Fatalf("failed to create temp dir: %v", err)
+			}
+			defer os.RemoveAll(tmpDir)
```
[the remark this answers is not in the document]
yaozile123: Use TempDir() might hitting the 108-character limit for Unix domain sockets.

pkg/csi_driver/node.go
```diff
@@ -810,24 +830,60 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
 
+	// We initialize volumeState iff bucket name is NOT "_", I.e. It's not a dynamic mount.
+	var vs *util.VolumeState
+	if args.bucketName != "_" {
+		var ok bool
+		if vs, ok = s.volumeStateStore.Load(stagingPath); !ok {
+			s.volumeStateStore.Store(stagingPath, &util.VolumeState{})
+			vs, _ = s.volumeStateStore.Load(stagingPath)
+		}
+	}
```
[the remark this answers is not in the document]
yaozile123: Done, add the check at the beginning of executeNodeStageVolume

pkg/csi_driver/node.go
```diff
@@ -917,6 +973,14 @@ func (s *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstage
 }
 
 func (s *nodeServer) cleanupStagingPath(stagingPath string) error {
+	// Stop GCSFuse Kernel Params monitoring.
+	if vs, ok := s.volumeStateStore.Load(stagingPath); ok {
+		if vs.GCSFuseKernelMonitorState.CancelFunc != nil {
+			vs.GCSFuseKernelMonitorState.CancelFunc()
+		}
+		s.volumeStateStore.Delete(stagingPath)
+	}
```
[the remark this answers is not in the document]

pkg/util/util.go
```diff
@@ -201,6 +201,17 @@ func PrepareEmptyDir(targetPath string, createEmptyDir bool) (string, error) {
...
+func PrepareMounterPodEmptyDir(mounterPodEmptyDirBasePath, volumeID string) (string, error) {
+	if mounterPodEmptyDirBasePath == "" {
+		return "", fmt.Errorf("mounterPodEmptyDirBasePath cannot be empty")
+	}
+	if volumeID == "" {
+		return "", fmt.Errorf("volumeID cannot be empty")
+	}
+
+	return filepath.Join(mounterPodEmptyDirBasePath, SidecarContainerTmpVolumeName, ".volumes", volumeID), nil
+}
```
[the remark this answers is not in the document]
yaozile123: According to the [Kubernetes Official Documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names), the API Server validates some of the resource names against RFC 1123 rules upfront. But it seems like the volumeHandle will not be validated by the API server after I verified locally. I have updated the method to prevent traversal attacks

pkg/csi_driver/node.go
```diff
@@ -810,24 +834,60 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
 	}
 
+	// We initialize volumeState iff bucket name is NOT "_", I.e. It's not a dynamic mount.
+	var vs *util.VolumeState
+	if args.bucketName != "_" {
+		var ok bool
+		if vs, ok = s.volumeStateStore.Load(stagingPath); !ok {
+			s.volumeStateStore.Store(stagingPath, &util.VolumeState{})
+			vs, _ = s.volumeStateStore.Load(stagingPath)
+		}
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/csi_driver/node.go
```diff
@@ -740,6 +770,10 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 }
 
 func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
+	if s.driver.config.FeatureOptions == nil || s.driver.config.FeatureOptions.SharedMountOptions == nil {
+		return nil, status.Errorf(codes.Internal, "shared mount options are not fully configured")
+	}
```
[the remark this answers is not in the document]

pkg/csi_driver/node.go
```diff
@@ -487,13 +487,43 @@ func (s *nodeServer) startGcsFuseKernelParamsMonitoring(targetPath, gcsFuseSidec
 	}
 }
 
-// isGcsFuseKernelParamsFeatureSupported returns true if the GCSFuse kernel parameters feature is enabled and supported by sidecar version.
+func getMounterPodImage(pod *corev1.Pod) (string, error) {
```
uriel-guzman: Function and method names should not use a Get or get prefix, unless the underlying concept uses the word "get" (e.g. an HTTP GET). Prefer starting the name with the noun directly, for example use Counts over GetCounts.

go/go-style/decisions#getters

uriel-guzman: Probably better to move this to `shared_mount.go`. And since it's a new helper function, write a small unit test for it.

pkg/util/util.go
```diff
@@ -201,6 +201,28 @@ func PrepareEmptyDir(targetPath string, createEmptyDir bool) (string, error) {
...
+// ensuring that the resolved path does not escape the expected base directory.
+func PrepareMounterPodEmptyDir(mounterPodEmptyDirBasePath, volumeID string) (string, error) {
+	if mounterPodEmptyDirBasePath == "" {
+		return "", fmt.Errorf("mounterPodEmptyDirBasePath cannot be empty")
+	}
+	if volumeID == "" {
+		return "", fmt.Errorf("volumeID cannot be empty")
+	}
+
+	expectedBase := filepath.Join(mounterPodEmptyDirBasePath, SidecarContainerTmpVolumeName, ".volumes")
```
uriel-guzman: In shared mount, we don't use `.volumes` subdir, because mounter pod only handles a single GCSFuse process / volume, so we don't need this. This means that this entire function isn't necessary, because the kernel params file is going to live directly on top of the emptyDir path, which is always guaranteed to exist, and doesn't require explicit creation.

uriel-guzman (changes requested):

pkg/csi_driver/node.go

yaozile123: Done, renamed it to mounterPodImage and moved it to shared_mount.go

pkg/util/util.go
yaozile123: Thanks for catching that, I have refactored the code to use mptyDirBasePath() instead

pkg/csi_driver/node.go
```diff
@@ -542,13 +542,34 @@ func (s *nodeServer) startGcsFuseKernelParamsMonitoring(targetPath, gcsFuseSidec
...
+		// Fire kernel params monitoring goroutine for shared mount.
+		go util.MonitorKernelParamsFileForSharedMount(ctx, stagingPath, emptyDirBasePath, volumeID, GCSFuseKernelParamsFilePollInterval)
+	})
+}
+
+// isGcsFuseKernelParamsFeatureSupported returns true if the GCSFuse kernel parameters feature is enabled and supported by sidecar/mounter version.
 // GCSFuse Kernel Params feature is not applicable for dynamic mounts as a single kernel parameter setting doesn’t apply for different type
 // of buckets that a dynamic mount serves. This feature is only applicable for non-dynamic mounts i.e when volumeState(vs) is not nil.
-func (s *nodeServer) isGcsFuseKernelParamsFeatureSupported(gcsFuseSidecarImage string, vs *util.VolumeState) bool {
+func (s *nodeServer) isGcsFuseKernelParamsFeatureSupported(gcsFuseImage string, vs *util.VolumeState) bool {
```
uriel-guzman: `gcsFuseImage` can be misleading, as it may be confused with the gcsfuse image binary. I think it's be better to just acll this `gcsFuseContainerImage`?

pkg/csi_driver/node.go
```diff
@@ -879,24 +896,60 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		}
 	}
 
+	// We initialize volumeState iff bucket name is NOT "_", I.e. It's not a dynamic mount.
+	var vs *util.VolumeState
+	if args.bucketName != "_" {
+		var ok bool
+		if vs, ok = s.volumeStateStore.Load(stagingPath); !ok {
+			vs = &util.VolumeState{}
+			s.volumeStateStore.Store(stagingPath, vs)
```
uriel-guzman, line 904: Why isn't this calling Load again? https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L280. This line is removed and it's not obvious to me why

pkg/csi_driver/node.go
```diff
@@ -879,24 +896,60 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
+		}
+	}
+
+	mounterPodImage, err := mounterPodImage(pod)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "%v in mounter pod %s/%s", err, podNamespace, podName)
```
uriel-guzman: This error message is kind of odd. It doesn't give any useful information about what / where it failed. Maybe `failed to check mount pod %s/%s image: %v`?

pkg/csi_driver/node.go
yaozile123: The `vs` object would always points to a empty `VolumeState` object, so the `Load()` would be redundant at here

pkg/csi_driver/node.go
```diff
@@ -276,8 +276,8 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	if args.bucketName != "_" {
 		var ok bool
 		if vs, ok = s.volumeStateStore.Load(targetPath); !ok {
-			s.volumeStateStore.Store(targetPath, &util.VolumeState{})
-			vs, _ = s.volumeStateStore.Load(targetPath)
+			vs = &util.VolumeState{}
+			s.volumeStateStore.Store(targetPath, vs)
```
uriel-guzman, line 280: Why was this changed?

yaozile123: Since the `vs` object would always points to a empty `VolumeState` object at here, the `Load()` would be redundant and we only need to call a `Store()`

pkg/csi_driver/node.go

pkg/csi_driver/node.go

pkg/util/kernel_params_monitoring.go
```diff
@@ -196,3 +180,32 @@ func MonitorKernelParamsFile(ctx context.Context, targetPath string, interval ti
...
+		klog.Warningf("%v Aborting kernel params monitor. Failed to get emptyDir path: %v", logPrefix, err)
+		return
+	}
+	kernelParamsFilePath := filepath.Join(emptyDirBasePath, GCSFuseKernelParamsFileName)
+
+	monitorKernelParamsCore(ctx, targetPath, kernelParamsFilePath, logPrefix, interval)
+}
+
+// MonitorKernelParamsFileForSharedMount monitors the kernel params file for shared node mounts.
+func MonitorKernelParamsFileForSharedMount(ctx context.Context, stagingPath, emptyDirBasePath, volumeID string, interval time.Duration) {
```
uriel-guzman: The only difference between `MonitorKernelParamsFileForSharedMount` and `MonitorKernelParamsFile` seems to be that `emptyDirBasePath` is differnt. Can't we just add a new parameter to the existing `MonitorKernelParamsFile` function and just pass `emptyDirBasePath` differently from NPV / NSV?

Another difference is that one uses `stagingPath` and the other uses `targetPath`, both of which are ultimately passed as `mountPoint` in `monitorKernelParamsCore`, so can't we just pass `mountPoint` as another parameter from `MonitorKernelParamsFile` too? it doesn't really seem to matter if it's a staging or a target path, it only matters whether it's a mount point, so I feel like this logic can be greately simplified by keeping one common function with different ways of passing parametrs, rather than creating new functions that are almost identical to eachother, which I think increases code complexity and reduces readability.

uriel-guzman (changes requested):

pkg/util/kernel_params_monitoring.go
yaozile123: Done, I have merged `MonitorKernelParamsFileForSharedMount` into a single, `MonitorKernelParamsFile` function for both sidecar and shared mounts

pkg/csi_driver/node.go
```diff
@@ -831,18 +853,9 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
-		return nil, status.Errorf(codes.Internal, "failed to check if path %q is already mounted: %v", stagingPath, err)
-	}
-	if mounted {
-		klog.Infof("NodeStageVolume succeeded on staging path %q for volume %q, mount already exists.", stagingPath, req.GetVolumeId())
-		return &csi.NodeStageVolumeResponse{}, nil
-	}
-
 	volumeID := req.GetVolumeId()
 	vc := req.GetVolumeContext()
+	pvName := vc[util.VolumeContextKeyPVName]
```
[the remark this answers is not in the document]
yaozile123: The `util.VolumeContextKeyPVName` will be passed in volume attributes via webhook

pkg/csi_driver/node.go
```diff
@@ -521,34 +533,38 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
 // startGcsFuseKernelParamsMonitoring starts a single long lived goroutine to monitor the
-// GCSFuse kernel parameters file per target path if the feature is enabled and supported.
-// It uses sync.Once to ensure the monitor is started only once. This monitoring goroutine is
-// stopped in the NodeUnpublishVolume operation.
-func (s *nodeServer) startGcsFuseKernelParamsMonitoring(targetPath, gcsFuseSidecarImage string, vs *util.VolumeState) {
-	if s.isGcsFuseKernelParamsFeatureSupported(gcsFuseSidecarImage, vs) {
+// GCSFuse kernel parameters file per mount point if the feature is enabled and supported.
+// It uses sync.Once to ensure the monitor is started only once.
+func (s *nodeServer) startGcsFuseKernelParamsMonitoring(mountPoint, emptyDirBasePath, podUID, volumeName, gcsFuseContainerImage string, vs *util.VolumeState) {
+	if s.isGcsFuseKernelParamsFeatureSupported(gcsFuseContainerImage, vs) {
```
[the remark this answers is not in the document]
yaozile123: The `emptyDirBasePath` will guarantee to be existed

pkg/csi_driver/node.go
```diff
@@ -879,24 +892,61 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		}
 	}
 
+	// We initialize volumeState iff bucket name is NOT "_", I.e. It's not a dynamic mount.
```
chrisThePattyEater: iff -> if

pkg/csi_driver/node.go
uriel-guzman, line 858: Validate this. If this is empty, then that's a bug from our side.

uriel-guzman (approved):

pkg/csi_driver/node.go
yaozile123: Done, I have updated the unit test as well

PR #1503, opened 2026-07-01
Implement NodePublishVolumeWithSharedMount mount logic.

uriel-guzman:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR introduces logic to `NodePublishVolumeForSharedMount` to:
1. Verify if the staging path already is mounted. If it isn't return error.
2. Verify if the target path is already mounted. If it is, return success.
3. Create the target path directory and bind-mount the staging path to the target path, appending `ro` when the mount is read-only.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -98,6 +98,19 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 	if len(stagingPath) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Staging Target Path must be provided")
 	}
+	targetPath := req.GetTargetPath()
+	if len(targetPath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Target Path must be provided")
+	}
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Volume ID must be provided")
+	}
+	if req.GetVolumeCapability() == nil {
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Volume capability must be provided")
+	}
+	if err := s.driver.validateVolumeCapabilities([]*csi.VolumeCapability{req.GetVolumeCapability()}); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/node.go
```diff
@@ -144,7 +156,43 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
...
+	if targetPathMounted {
+		klog.Infof("NodePublishVolume succeeded on staging path %q to target path %q, mount already exists.", stagingPath, targetPath)
+		return &csi.NodePublishVolumeResponse{}, nil
+	}
+
+	// Create target path dir.
+	klog.Infof("NodePublishVolume attempting mkdir for target path %q", targetPath)
+	if err := os.MkdirAll(targetPath, 0o750); err != nil {
+		return nil, status.Errorf(codes.Internal, "mkdir failed for path %q: %v", targetPath, err)
+	}
```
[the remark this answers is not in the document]
uriel-guzman: out of scope, existing CSI driver doesn't validate this. If we ever fix this in the CSI Driver sidecar architecture, we'll also fix it here.

pkg/csi_driver/node.go
```diff
@@ -111,6 +115,12 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Mounter Pod Name must be provided")
 	}
```
yaozile123, line 126: Could we also validate `req.GetVolumeCapability()` via `s.driver.validateVolumeCapabilities()` here to match [NodePublishVolume](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/a8308b53dbb1364d93e5331c7b5ed23c751c8b6f/pkg/csi_driver/node.go#L207-L209)?

yaozile123 (approved):

pkg/csi_driver/node.go
uriel-guzman: done, I also added volume id validation to complete the match

PR #1512, opened 2026-07-08
filter Node driver PV cache to profile-managed volumes only to prevent OOMs

ILLUM1N0X:
**What type of PR is this?**
/kind bug

This PR resolves an issue where the Node driver DaemonSet periodically crashes with Out-Of-Memory (OOM) errors in clusters containing tens of thousands of PersistentVolumes (such as PD CSI PVs). Previously, the Node driver would aggressively cache the entire cluster's PVs into memory.

To fix this, the Node driver's PV Informer is updated to filter via a label selector (`gke-gcsfuse/profile-managed=true`), so it only caches PVs that specifically utilize the GCS Fuse Profiles feature.

Legacy PVs provisioned before this change will not have the required label. To upgrade them without breaking existing mounts, the Controller driver runs an unfiltered informer and automatically patches the label onto any older GCS Fuse profile PVs it discovers during its routine scanning loops.

N/A

N/A

```release-note
Fixed an issue where the GCS Fuse CSI Node driver would OOM in clusters with tens of thousands of unrelated PersistentVolumes.
```

### DIFF
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -181,6 +181,13 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(
 		c.k8sClients,
 		time.Duration(c.informerResyncDurationSec)*time.Second,
+		// To prevent OOMs, the Node driver uses a server-side filter to cache only profile-managed PVs since profiles is the only feature using GetPV.
+		// Leaving the Controller unfiltered to discover and patch legacy volumes.
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			if !c.runController {
+				options.LabelSelector = fmt.Sprintf("%s=%s", webhook.GcsfuseProfilesManagedLabel, util.TrueStr)
+			}
+		}),
 		informers.WithTransform(trim),
 	)
 	pvLister := informerFactory.Core().V1().PersistentVolumes().Lister()
@@ -463,6 +470,12 @@ func (c *Clientset) GetNode(name string) (*corev1.Node, error) {
 	return c.nodeLister.Get(name)
 }
 
+// GetPV retrieves a PersistentVolume from the informer cache by name.
+// IMPORTANT: In the Node driver, the PV informer cache is intentionally filtered down
+// to ONLY track PVs that utilize the GCS Fuse profiles feature (via the gke-gcsfuse/profile-managed label)
+// to minimize the memory footprint in large-scale clusters.
+// Do not attempt to use this function from the Node driver to look up generic PVs or non-profile GCS Fuse PVs,
+// as they are explicitly prevented from entering the Node's cache.
 func (c *Clientset) GetPV(name string) (*corev1.PersistentVolume, error) {
 	if c.pvLister == nil {
 		return nil, errors.New("pv informer is not ready")
--- a/pkg/profiles/scanner.go
+++ b/pkg/profiles/scanner.go
@@ -1544,6 +1544,39 @@ func defaultScanBucketWithDataflux(ctx context.Context, gcsClient *storage.Clien
 	}
 }
 
+// ensurePVTrackingLabel applies the PV tracking label if it's currently missing.
+func (s *Scanner) ensurePVTrackingLabel(ctx context.Context, pv *v1.PersistentVolume) error {
+	if err := ctx.Err(); err != nil {
+		return status.Errorf(codes.Canceled, "context cancelled: %v", err)
+	}
+	if pv == nil {
+		return status.Error(codes.Internal, "persistent volume is nil")
+	}
+	if _, hasTrackedLabel := pv.Labels[profileManagedLabelKey]; hasTrackedLabel {
+		return nil
+	}
+
+	patchData := map[string]any{
+		"metadata": map[string]any{"labels": map[string]string{profileManagedLabelKey: util.TrueStr}},
+	}
+	patchBytes, marshalErr := json.Marshal(patchData)
+	if marshalErr != nil {
+		return status.Errorf(codes.Internal, "failed to marshal label patch data for PV %q: %v", pv.Name, marshalErr)
+	}
+	klog.V(6).Infof("PV %q is missing the tracking label. Patching it with: %q", pv.Name, string(patchBytes))
+	_, patchErr := s.kubeClient.CoreV1().PersistentVolumes().Patch(ctx, pv.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	if patchErr != nil {
+		if apierrors.IsNotFound(patchErr) {
+			klog.Warningf("Failed to patch PV %q because it was not found", pv.Name)
+			return nil
+		}
+		return status.Errorf(codes.Internal, "failed to patch PV %q with label: %v", pv.Name, patchErr)
+	}
+
+	klog.V(6).Infof("Successfully added %s=%s label to PV %q", profileManagedLabelKey, util.TrueStr, pv.Name)
+	return nil
+}
+
 // patchPVAnnotations updates the annotations for a given PV.
 // annotationsToUpdate should contain the desired state of the annotations to be patched.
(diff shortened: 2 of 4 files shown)


pkg/cloud_provider/clientset/clientset.go
```diff
@@ -178,10 +178,21 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
...
+		opts = append(opts, informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.LabelSelector = pvLabelSelector
+		}))
+	}
+
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(
 		c.k8sClients,
 		time.Duration(c.informerResyncDurationSec)*time.Second,
-		informers.WithTransform(trim),
+		opts...,
```
uriel-guzman: A cleaner way of doing this is using `WithTweakListOptions`. See https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/cloud_provider/clientset/clientset.go#L391 as an example, then you can put the `c.runController` check inside here.

pkg/util/util.go
```diff
@@ -41,6 +41,9 @@ const (
 	TrueStr  = "true"
 	FalseStr = "false"
 
+	// Label for filtering Persistent Volumes
+	TrackedPVLabelKey = "gke-gcsfuse/tracked-pv"
```
uriel-guzman: There's already a label for profiles resources: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/scanner.go#L129. Let's use that instead.

pkg/profiles/scanner.go
```diff
@@ -1639,6 +1670,10 @@ func (s *Scanner) checkPVRelevance(ctx context.Context, pv *v1.PersistentVolume,
 		return nil, false, nil
 	}
 
+	if err := s.ensurePVTrackingLabel(ctx, pv); err != nil {
+		klog.Errorf("Failed to ensure tracking label for PV %q: %v", pv.Name, err)
+	}
```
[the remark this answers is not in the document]
amacaskill: This is intended per what we discussed in the Chat (https://chat.google.com/room/AAQA4YOBNw4/1HviAgVDJyc), but @ILLUM1N0X please add comments to this code explaining why we are doing this, and add comment to GetPV explaining this cannot be called by anything besides PV using profiles.

pkg/profiles/scanner.go
```diff
@@ -1544,6 +1544,37 @@ func defaultScanBucketWithDataflux(ctx context.Context, gcsClient *storage.Clien
...
+
+	patchData := map[string]any{
+		"metadata": map[string]any{"labels": map[string]string{util.TrackedPVLabelKey: util.TrueStr}},
+	}
+	patchBytes, marshalErr := json.Marshal(patchData)
+	if marshalErr != nil {
+		return status.Errorf(codes.Internal, "failed to marshal label patch data for PV %q: %v", pv.Name, marshalErr)
+	}
+	klog.Infof("PV %q is missing the tracking label. Patching it with: %q", pv.Name, string(patchBytes))
```
uriel-guzman: This is going to log for each PV. It will cause spam / noise in the controller logs. Downgrade this to V6 verbosity.

pkg/profiles/scanner.go
```diff
@@ -1544,6 +1544,37 @@ func defaultScanBucketWithDataflux(ctx context.Context, gcsClient *storage.Clien
 	}
 }
 
+// ensurePVTrackingLabel applies the PV tracking label if it's currently missing.
+func (s *Scanner) ensurePVTrackingLabel(ctx context.Context, pv *v1.PersistentVolume) error {
```
uriel-guzman, line 1548: If the caller of ensurePVTrackingLabel relies on status.FromError(err) to inspect gRPC codes, the context cancellation error will lose its specific gRPC identity (which should ideally be codes.Canceled).

Instead, do:
```
if err := ctx.Err(); err != nil {
    return status.Errorf(codes.Canceled, "context cancelled: %v", err)
}
```

pkg/profiles/scanner.go
```diff
@@ -1544,6 +1544,37 @@ func defaultScanBucketWithDataflux(ctx context.Context, gcsClient *storage.Clien
...
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("context cancelled: %w", err)
+	}
+
+	if _, hasTrackedLabel := pv.Labels[util.TrackedPVLabelKey]; hasTrackedLabel {
```
uriel-guzman: If pv is ever passed as nil, pv.Labels will cause a segmentation fault. Perform a defensive nil pointer check here and return Internal status code if it's nil.

pkg/profiles/scanner.go
```diff
@@ -1544,6 +1544,37 @@ func defaultScanBucketWithDataflux(ctx context.Context, gcsClient *storage.Clien
...
+	_, patchErr := s.kubeClient.CoreV1().PersistentVolumes().Patch(ctx, pv.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	if patchErr != nil {
+		if apierrors.IsNotFound(patchErr) {
+			klog.Warningf("Failed to patch PV %q because it was not found", pv.Name)
+			return nil
+		}
+		return status.Errorf(codes.Internal, "failed to patch PV %q with label: %v", pv.Name, patchErr)
+	}
+
+	klog.Infof("Successfully added %s=%s label to PV %q", util.TrackedPVLabelKey, util.TrueStr, pv.Name)
```
uriel-guzman: Downgrade to V6 to reduce control plane noise logs.

uriel-guzman (changes requested):

pkg/profiles/scanner.go
```diff
@@ -1639,6 +1670,10 @@ func (s *Scanner) checkPVRelevance(ctx context.Context, pv *v1.PersistentVolume,
 		return nil, false, nil
 	}
 
+	if err := s.ensurePVTrackingLabel(ctx, pv); err != nil {
+		klog.Errorf("Failed to ensure tracking label for PV %q: %v", pv.Name, err)
```
uriel-guzman: If `ensurePVTrackingLabel` fails to patch the label, the PV won't be visible to the CSI Node informer, and will cause profiles users to experience bugs / their profiles volume won't work, so I feel like this is a pretty impotant error to monitor. 

There's an SLO called `gcsfusecsi_controller_sync_pv_success rate` and fires when `syncPV` fails, so I think we should return the error here instead of just logging it, because if this fails, the rest of the workflow is broken anyway for the profiles user.

pkg/util/util.go
ILLUM1N0X: I initially used that label as well, but @amacaskill and I decided to change it since it's already tied to the pods. I don't have a strong opinion either way, so just let me know which direction you'd prefer to go.

uriel-guzman: I don't see why it can't be used for both pods and pvs?

uriel-guzman: I discussed with @amacaskill offline:

> Yea profile label to me makes sense now. He used the profile label when he modified the webhook to patch the label on any pv.create + filtered for it in the node driver, but now with the new approach , I agree it should be profile label

pkg/cloud_provider/clientset/clientset.go
ILLUM1N0X: Looks much better

ILLUM1N0X:
Addressed comments

pkg/cloud_provider/clientset/clientset.go
```diff
@@ -181,6 +181,11 @@ func (c *Clientset) ConfigurePVLister(ctx context.Context) {
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(
 		c.k8sClients,
 		time.Duration(c.informerResyncDurationSec)*time.Second,
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
```
amacaskill, line 186: nit: please just add small comment here: 
			// To prevent OOMs, the Node driver uses a server-side filter to cache only profile-managed PVs since profiles is the only feature using GetPV, leaving the Controller unfiltered to discover and patch legacy volumes.

amacaskill (approved):

uriel-guzman (approved):

PR #1514, opened 2026-07-09
feat: add support for tuning host FUSE max_pages_limit during file descriptor mounting

meet2mky:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds support for dynamically tuning the host's FUSE `max_pages_limit` during the volume mount operation, the CSI driver:
1. Calculates the required page limit (based on the host's page size).
2. Acquires a lock to prevent concurrent modifications.
3. Temporarily raises the host's `/proc/sys/fs/fuse/max_pages_limit` (if the current value is lower).
4. Executes the FUSE mount.
5. Restores the host's original `max_pages_limit`.
This enables support for larger FUSE request sizes (e.g. up to 16 MiB), which can significantly improve throughput/performance.

Fixes #
b/526496388

* The new CSI option `node_fuse_max_request_limit_kb` is a "hidden" option: it is parsed and stripped by the CSI driver and not passed down to the `gcsfuse` binary.
* Concurrent mount operations on the same node are serialized during the limit adjustment and mount execution using a global lock (`util.FuseMaxMaxPagesMu`).
* Unit tests have been added to verify reading, setting, and checking support for `max_pages_limit` under `pkg/util/kernel_params_monitoring_test.go` and the parsing logic in `pkg/csi_mounter/csi_mounter_test.go`.

```release-note
NONE
```

### DIFF
--- a/deploy/base/node/node.yaml
+++ b/deploy/base/node/node.yaml
@@ -112,6 +112,9 @@ spec:
               mountPath: /sys
             - name: host-iproute
               mountPath: /etc/iproute2
+            - name: host-proc-sys-fs-fuse
+              mountPath: /host-proc-sys-fs-fuse
+
             - name: tmp
               mountPath: /tmp
         - name: csi-driver-registrar
@@ -168,6 +171,10 @@ spec:
         - name: host-iproute
           hostPath:
             path: /etc/iproute2
+            type: Directory
+        - name: host-proc-sys-fs-fuse
+          hostPath:
+            path: /proc/sys/fs/fuse
             type: Directory
         - name: tmp
           emptyDir:
--- a/pkg/csi_mounter/csi_mounter.go
+++ b/pkg/csi_mounter/csi_mounter.go
@@ -43,12 +43,24 @@ import (
 )
 
 const (
+	// Note: All variables here are in KiB instead of KB but they are added like this to ensure consistency in codebase.
 	socketName                       = "socket"
 	readAheadKBMountFlagRegexPattern = "^read_ahead_kb=(.+)$"
 	readAheadKBMountFlag             = "read_ahead_kb"
+	// Hidden CSI flag never to be passed to gcsfuse
+	nodeFuseMaxRequestLimitKBMountFlag             = "node_fuse_max_request_limit_kb"
+	nodeFuseMaxRequestLimitKBMountFlagRegexPattern = "^" + nodeFuseMaxRequestLimitKBMountFlag + "=(.+)$"
+	defaultNodeFuseMaxRequestLimitKB               = 16 * util.MiB / util.KiB // 16 MiB Request Size
+	// minPageSizeKB defines the minimum page size (4 KiB) to use as a fallback.
+	minPageSizeKB = 4
+	// maxFuseMaxPagesLimit defines the maximum pages limit supported by the Linux FUSE kernel (2^16 - 1).
+	maxFuseMaxPagesLimit = 65535
 )
 
-var readAheadKBMountFlagRegex = regexp.MustCompile(readAheadKBMountFlagRegexPattern)
+var (
+	readAheadKBMountFlagRegex               = regexp.MustCompile(readAheadKBMountFlagRegexPattern)
+	nodeFuseMaxRequestLimitKBMountFlagRegex = regexp.MustCompile(nodeFuseMaxRequestLimitKBMountFlagRegexPattern)
+)
 
 // Mounter provides the Cloud Storage FUSE CSI implementation of mount.Interface
 // for the linux platform.
@@ -78,7 +90,7 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 	m.mux.Lock()
 	defer m.mux.Unlock()
 
-	csiMountOptions, sidecarMountOptions, sysfsBDI, err := prepareMountOptions(options)
+	csiMountOptions, sidecarMountOptions, sysfsBDI, fuseMaxPagesLimit, err := prepareMountOptions(options)
 	if err != nil {
 		return err
 	}
@@ -104,8 +116,18 @@ func (m *Mounter) Mount(source string, target string, fstype string, options []s
 	}
 	csiMountOptions = append(csiMountOptions, fmt.Sprintf("fd=%v", fd))
 
-	klog.V(4).Infof("%v mounting the fuse filesystem", logPrefix)
-	err = m.MountSensitiveWithoutSystemdWithMountFlags(source, target, fstype, csiMountOptions, nil, []string{"--internal-only"})
+	mountFn := func() error {
+		klog.V(4).Infof("%v mounting the fuse filesystem", logPrefix)
+		return m.MountSensitiveWithoutSystemdWithMountFlags(source, target, fstype, csiMountOptions, nil, []string{"--internal-only"})
+	}
+	// TODO(mohitkyadav): Remove this check when kernel reader is enabled by default for Regional Buckets.
+	isKernelReaderEnabled := checkForKernelReader(options)
+	shouldMountUsingElevatedFuseMaxPagesLimit := util.FuseMaxMaxPagesUpdateSupported() && fuseMaxPagesLimit > 0 && isKernelReaderEnabled
+	if shouldMountUsingElevatedFuseMaxPagesLimit {
+		err = util.MountUsingElevatedFuseMaxPagesLimit(fuseMaxPagesLimit, logPrefix, mountFn)
+	} else {
+		err = mountFn()
+	}
 	if err != nil {
 		klog.Errorf("MountSensitiveWithoutSystemdWithMountFlags failed with error %v", err)
 		return fmt.Errorf("failed to mount the fuse filesystem: %w", err)
@@ -357,7 +379,7 @@ func startAcceptConn(l net.Listener, logPrefix string, msg []byte, fd int, cance
 	klog.V(4).Infof("%v exiting the listener goroutine.", logPrefix)
 }
 
-func prepareMountOptions(options []string) ([]string, []string, map[string]int64, error) {
+func prepareMountOptions(options []string) ([]string, []string, map[string]int64, int64, error) {
 	allowedOptions := map[string]bool{
 		"exec":    true,
 		"noexec":  true,
@@ -390,6 +412,11 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
 	}
 
 	sysfsBDI := make(map[string]int64)
+	pageSizeKB := max(minPageSizeKB, int64(os.Getpagesize()/util.KiB))
+	// Calculate default FUSE max pages limit. We use integer ceiling division to ensure we round
+	// up to the next page if the default size 16MB is not a perfect multiple of the page size on the machine.
+	fuseMaxPagesLimit := util.CeilDiv64(defaultNodeFuseMaxRequestLimitKB, pageSizeKB)
+
 	for _, o := range optionSet.List() {
 		if strings.HasPrefix(o, "o=") {
 			v := o[2:]
@@ -406,15 +433,44 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
 			// If found, it will be at index 1
 			readAheadKBInt, err := strconv.ParseInt(readAheadKB[1], 10, 0)
 			if err != nil {
-				return nil, nil, nil, fmt.Errorf("invalid read_ahead_kb mount flag %q: %w", o, err)
+				return nil, nil, nil, 0, fmt.Errorf("invalid read_ahead_kb mount flag %q: %w", o, err)
 			}
 			if readAheadKBInt < 0 {
-				return nil, nil, nil, fmt.Errorf("invalid negative value for read_ahead_kb mount flag: %q", o)
+				return nil, nil, nil, 0, fmt.Errorf("invalid negative value for read_ahead_kb mount flag: %q", o)
 			}
 			sysfsBDI[readAheadKBMountFlag] = readAheadKBInt
 			optionSet.Delete(o)
 		}
+
+		if maxReqSize := nodeFuseMaxRequestLimitKBMountFlagRegex.FindStringSubmatch(o); len(maxReqSize) == 2 {
+			nodeFuseMaxRequestLimitKB, err := strconv.ParseInt(maxReqSize[1], 10, 0)
+			if err != nil {
+				return nil, nil, nil, 0, fmt.Errorf("invalid %s mount flag %q: %w", nodeFuseMaxRequestLimitKBMountFlag, o, err)
+			}
+			if nodeFuseMaxRequestLimitKB < 0 {
+				return nil, nil, nil, 0, fmt.Errorf("invalid negative value for %s mount flag: %q", nodeFuseMaxRequestLimitKBMountFlag, o)
+			}
+			maxAllowedSizeKB := maxFuseMaxPagesLimit * pageSizeKB
+			if nodeFuseMaxRequestLimitKB > maxAllowedSizeKB {
+				return nil, nil, nil, 0, fmt.Errorf("invalid value for %s mount flag %q: exceeds maximum allowed limit of %d KiB on this host", nodeFuseMaxRequestLimitKBMountFlag, o, maxAllowedSizeKB)
+			}
+			// Convert the requested size in KB to pages. We use integer ceiling division
+			// to round up to the next page if the requested size is not a multiple of the
+			// page size. This ensures we can accommodate the full requested buffer size in
+			// a single fuse request.
+			fuseMaxPagesLimit = util.CeilDiv64(nodeFuseMaxRequestLimitKB, pageSizeKB)
+			optionSet.Delete(o)
+		}
 	}
 
-	return csiMountOptions, optionSet.List(), sysfsBDI, nil
+	return csiMountOptions, optionSet.List(), sysfsBDI, fuseMaxPagesLimit, nil
+}
+
+func checkForKernelReader(options []string) bool {
+	for _, o := range options {
+		if o == "enable-kernel-reader" || o == "enable-kernel-reader=true" || o == "file-system:enable-kernel-reader:true" {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/csi_mounter/csi_mounter_test.go
+++ b/pkg/csi_mounter/csi_mounter_test.go
@@ -22,6 +22,8 @@ import (
 	"os"
 	"reflect"
 	"testing"
+
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 )
 
 var defaultCsiMountOptions = []string{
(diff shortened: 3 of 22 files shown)


pkg/csi_mounter/csi_mounter.go
```diff
@@ -406,15 +435,45 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
...
+			fuseMaxPagesLimit = (nodeFuseMaxRequestLimitKB + pageSizeKB - 1) / pageSizeKB
+			optionSet.Delete(o)
+		}
 	}
 
-	return csiMountOptions, optionSet.List(), sysfsBDI, nil
+	return csiMountOptions, optionSet.List(), sysfsBDI, fuseMaxPagesLimit, nil
+}
+
+func checkForKernelReader(options []string) bool {
```
abhishek10004, line 469: Do we need to check for just enable-kernel-reader as well (i.e just specifying the flag without explicitly setting it to true)? or is that not allowed on GKE?

pkg/csi_mounter/csi_mounter.go
```diff
@@ -390,6 +413,12 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
 	}
 
 	sysfsBDI := make(map[string]int64)
+	pageSizeKB := max(minPageSizeKB, int64(os.Getpagesize()/KiB))
```
abhishek10004: I get that this is safer but why do we need to have a minPageSizeKB as a fallback? I can't think of any scenario where pageSize is less than 4K.

pkg/csi_mounter/csi_mounter_test.go
```diff
@@ -37,12 +37,15 @@ var defaultCsiMountOptions = []string{
 func TestPrepareMountArgs(t *testing.T) {
 	t.Parallel()
 
+	pageSizeKB := max(minPageSizeKB, int64(os.Getpagesize()/1024))
```
abhishek10004: nit: can use KiB here instead of hardcoded value

pkg/csi_mounter/csi_mounter_test.go
```diff
@@ -90,14 +93,53 @@ func TestPrepareMountArgs(t *testing.T) {
 			inputMountOptions: append(defaultCsiMountOptions, "read_ahead_kb=-1"),
 			expectErr:         true,
 		},
+		{
+			name:                       "should return valid options correctly with node_fuse_max_request_limit_kb",
+			inputMountOptions:          []string{"ro", "implicit-dirs", "node_fuse_max_request_limit_kb=8192"},
+			expecteCsiMountOptions:     append(defaultCsiMountOptions, "ro"),
```
abhishek10004: nit: expectedCsiMountOptions

pkg/csi_mounter/csi_mounter_test.go
```diff
@@ -90,14 +93,53 @@ func TestPrepareMountArgs(t *testing.T) {
...
+			expecteSidecarMountOptions: []string{"implicit-dirs"},
```
abhishek10004: nit: expectedSidecarMountOptions

pkg/util/kernel_params_monitoring.go
```diff
@@ -183,3 +211,29 @@ func MonitorKernelParamsFile(ctx context.Context, mountPoint, emptyDirBasePath,
 		}
 	}
 }
+
+// TuneFuseMaxPagesLimit temporarily increases the host's FUSE max_pages_limit if the current limit is lower,
+// executes the provided function, and then restores the original limit.
+func TuneFuseMaxPagesLimit(targetLimit int64, logPrefix string, fn func() error) error {
```
abhishek10004: nit: Would RunWithTunedFuseMaxPagesLimit or RunWithElevatedFuseMaxPagesLimit be a better name for this function?

pkg/util/kernel_params_monitoring_test.go
```diff
@@ -381,3 +383,288 @@ func TestValidateParamValue(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("failed to read final limit: %v", err)
+			}
+			if finalLimit != tc.expectedFinalLimit {
+				t.Errorf("Expected final limit to be %d, got %d", tc.expectedFinalLimit, finalLimit)
+			}
+		})
+	}
+
+	t.Run("should restore limit even if fn panics", func(t *testing.T) {
```
abhishek10004: nit: should we move this panic test scenario to a separate test itself?

test/vendor/github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util/kernel_params_monitoring.go
abhishek10004, line 1: What is this file?

abhishek10004 (commented):
Looks good. Left a few minor comments.

pkg/csi_mounter/csi_mounter.go
meet2mky: Thanks, It does. Added it.

pkg/csi_mounter/csi_mounter.go
meet2mky: It's just defensive programming.

pkg/csi_mounter/csi_mounter_test.go
meet2mky: Done

pkg/csi_mounter/csi_mounter_test.go
meet2mky: Done.

pkg/csi_mounter/csi_mounter_test.go
meet2mky: Done

pkg/util/kernel_params_monitoring.go
meet2mky: Made it more explicit. Thanks!

pkg/util/kernel_params_monitoring_test.go
meet2mky: Done.

test/vendor/github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util/kernel_params_monitoring.go
meet2mky: vendors are committed here.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -471,7 +471,7 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
 
 func checkForKernelReader(options []string) bool {
 	for _, o := range options {
-		if o == "enable-kernel-reader=true" || o == "file-system:enable-kernel-reader:true" {
+		if o == "enable-kernel-reader" || o == "enable-kernel-reader=true" || o == "file-system:enable-kernel-reader:true" {
```
abhishek10004, line 471: should we trim spaces etc. before comparison here or is that not allowed when specifying with = or :?

meet2mky: All mount options must be passed without spaces. That's how the current implementation is in the Driver. If there are any unintended spaces then it should result in error.

meet2mky:
# E2E Test Report: Kernel Params Tuning

* **Status**: **SUCCESS (PASSED)**
* **Specs Run**: `5`
* **Specs Passed**: `5`
* **Ginkgo Run Time**: `38.08s`
* **Target Cluster**: `mohitkyadav-cluster`
* **GCSFuse Version**: `v3.11.1`
* **Test Focus Pattern**: `"enabling kernel reader results in 16MiB|enabling kernel reader with non-default setting|limit_kb=0 disables higher|enabling kernel reader on multiple volumes|reverted back to its original default value"`
* **Log Location**: [test_run.log](file:///usr/local/google/home/mohitkyadav/kernel-reader-regional-bkt/gcs-fuse-csi-driver/test_run.log)

---

### Executed Tests

| Test Spec Name | Status |
| :--- | :--- |
| `should verify enabling kernel reader results in 16MiB readFile I/O` | **PASSED** |
| `should verify enabling kernel reader with non-default setting results in 32MiB read and write file I/O` | **PASSED** |
| `should verify node_fuse_max_request_limit_kb=0 disables higher file I/O` | **PASSED** |
| `should verify enabling kernel reader on multiple volumes inside a single pod yields independent 16MiB and 32MiB block file I/O` | **PASSED** |
| `should verify that once the GCSFuse mount is done the FUSE max_pages_limit on the host is reverted back to its original default value` | **PASSED** |

test/vendor/github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util/kernel_params_monitoring.go
```diff
@@ -24,13 +24,41 @@ import (
...
 	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
 )
 
+var (
+	// FuseMaxMaxPagesMu serializes concurrent updates to the host's FUSE max_pages_limit.
+	FuseMaxMaxPagesMu sync.Mutex
+	// procSysFsFuseMaxPagesLimitPath is the host FUSE max_pages_limit path (overridable for unit testing).
+	procSysFsFuseMaxPagesLimitPath = "/proc/sys/fs/fuse/max_pages_limit"
+)
```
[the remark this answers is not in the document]
meet2mky: Fixed, also synced test vendor files.

pkg/util/kernel_params_monitoring.go
```diff
@@ -183,3 +211,29 @@ func MonitorKernelParamsFile(ctx context.Context, mountPoint, emptyDirBasePath,
...
+				klog.Infof("%v Restoring host FUSE max_pages_limit back to %d", logPrefix, origLimit)
+				if err := SetFuseMaxPagesLimit(origLimit); err != nil {
+					klog.Errorf("%v Failed to restore host FUSE max_pages_limit to %d: %v", logPrefix, origLimit, err)
+				}
+			}()
+		}
+	}
+
+	return fn()
+}
```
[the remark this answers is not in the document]
meet2mky: Done

pkg/csi_mounter/csi_mounter.go
```diff
@@ -390,6 +413,12 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
 	}
 
 	sysfsBDI := make(map[string]int64)
+	pageSizeKB := max(minPageSizeKB, int64(os.Getpagesize()/KB))
+	// Calculate default FUSE max pages limit. We use integer ceiling division
+	// (A + B - 1) / B to ensure we round up to the next page if the default size 16MB
```
uriel-guzman: There already exists a helper function for int64 ceil division using (a + b - 1) / b here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/recommender.go#L1052. You can just put it in util.go and share it: 
`fuseMaxPagesLimit := util.CeilDiv64(defaultNodeFuseMaxRequestLimitKB, pageSizeKB)`.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -46,9 +46,22 @@ const (
 	socketName                       = "socket"
 	readAheadKBMountFlagRegexPattern = "^read_ahead_kb=(.+)$"
 	readAheadKBMountFlag             = "read_ahead_kb"
+	// Hidden CSI flag never to be passed to gcsfuse
+	nodeFuseMaxRequestLimitKBMountFlag             = "node_fuse_max_request_limit_kb"
+	nodeFuseMaxRequestLimitKBMountFlagRegexPattern = "^" + nodeFuseMaxRequestLimitKBMountFlag + "=(.+)$"
+	KB                                             = 1024
+	MB                                             = 1024 * KB
```
uriel-guzman: MB (Megabyte) is 1000 * 1000. MiB (Mebibyte) is 1024 * 1024: https://developers.google.com/style/units-of-measure#byte-units. Looks like you want binary units, which already exist here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/profiles/recommender.go#L72. Put them into util.go and share them.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -46,9 +46,22 @@ const (
...
+	defaultNodeFuseMaxRequestLimitKB               = 16 * MB / KB // 16 MB Request Size
```
uriel-guzman: The comment should be 16 MiB request size, since you're using binary units.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -46,9 +46,22 @@ const (
...
+	// minPageSizeKB defines the minimum page size (4 KB) to use as a fallback.
```
uriel-guzman: // // minPageSizeKB defines the minimum page size (4 KiB) to use as a fallback.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -46,9 +46,22 @@ const (
...
+	minPageSizeKB = 4
```
uriel-guzman, line 55: minPageSizeKiB

pkg/csi_mounter/csi_mounter.go
```diff
@@ -46,9 +46,22 @@ const (
 	socketName                       = "socket"
 	readAheadKBMountFlagRegexPattern = "^read_ahead_kb=(.+)$"
 	readAheadKBMountFlag             = "read_ahead_kb"
+	// Hidden CSI flag never to be passed to gcsfuse
+	nodeFuseMaxRequestLimitKBMountFlag             = "node_fuse_max_request_limit_kb"
```
uriel-guzman, line 51: Should it be node_fuse_max_request_limit_kb or node_fuse_max_request_limit_kib? I assume we're trying to keep it consistent with read_ahead_kb, even though internally we know it's kib?

pkg/csi_mounter/csi_mounter.go
```diff
@@ -406,15 +435,45 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
...
+			}
+			maxAllowedSizeKB := maxFuseMaxPagesLimit * pageSizeKB
+			if nodeFuseMaxRequestLimitKB > maxAllowedSizeKB {
+				return nil, nil, nil, 0, fmt.Errorf("invalid value for %s mount flag %q: exceeds maximum allowed limit of %d KB on this host", nodeFuseMaxRequestLimitKBMountFlag, o, maxAllowedSizeKB)
+			}
+			// Convert the requested size in KB to pages. We use integer ceiling division
+			// to round up to the next page if the requested size is not a multiple of the
+			// page size. This ensures we can accommodate the full requested buffer size in
+			// a single fuse request.
+			fuseMaxPagesLimit = (nodeFuseMaxRequestLimitKB + pageSizeKB - 1) / pageSizeKB
```
uriel-guzman: Same comment as https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1514/changes#r3599414521

uriel-guzman (changes requested):

pkg/csi_mounter/csi_mounter.go
```diff
@@ -406,15 +435,45 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
...
 			if readAheadKBInt < 0 {
-				return nil, nil, nil, fmt.Errorf("invalid negative value for read_ahead_kb mount flag: %q", o)
+				return nil, nil, nil, 0, fmt.Errorf("invalid negative value for read_ahead_kb mount flag: %q", o)
 			}
 			sysfsBDI[readAheadKBMountFlag] = readAheadKBInt
 			optionSet.Delete(o)
 		}
+
+		if maxReqSize := nodeFuseMaxRequestLimitKBMountFlagRegex.FindStringSubmatch(o); len(maxReqSize) == 2 {
+			var err error
```
uriel-guzman: Is this err variable being used? In the line below, you're using `:=`, so it's creating a new error variable.

test/e2e/testsuites/kernel_params.go
```diff
@@ -140,6 +152,19 @@ func skipIfKernelParamsNotSupported() {
...
 }
 
+func skipIfFuseMaxRequestSizeNotSupported() {
+	gcsfuseVersion, branch := specs.GCSFuseVersionAndBranch()
+
+	// Running since we are on master branch.
+	if branch == utils.MasterBranchName {
+		return
+	}
+	// fuse-max-request-size-kb was introduced in v3.11.0-gke.0.
```
uriel-guzman: This says it was introduced in v3.11.0-gke.0, the actual `MinGCSFuseFuseMaxRequestSizeVersion` says it was introduced in v3.11.1-gke.0, which one is it?

uriel-guzman (changes requested):

test/vendor/github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -235,6 +179,80 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		time.Sleep(time.Second * 5)
+
+		if cmd.ProcessState == nil || !cmd.ProcessState.Exited() {
+			klog.Warningf("after 5 seconds, process with id %v has not exited, force kill the process", cmd.Process.Pid)
+			if err := cmd.Process.Kill(); err != nil {
+				klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
+			}
+		}
+	}(cmd)
+}
```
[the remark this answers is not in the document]
uriel-guzman: This is unrelated

test/vendor/github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter/mounter_server.go
```diff
@@ -0,0 +1,89 @@
...
+func (ms *MounterServer) Mount(ctx context.Context, req *mounter.MountRequest) (*mounter.MountResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.Internal, "mount request cannot be nil")
+	}
+	if req.GetVolumeId() == "" {
+		return nil, status.Error(codes.Internal, "volume id cannot be empty")
+	}
+	if req.GetMountPoint() == "" {
+		return nil, status.Error(codes.Internal, "mount point cannot be empty")
+	}
```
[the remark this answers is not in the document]
uriel-guzman: This is unrelated.

deploy/base/node/node.yaml
uriel-guzman, line 60: Are we not adding a feature flag like `enable-gcsfuse-kernel-params` because the mount option acts as the feature flag? Typically, these feature flags are for us to set to false rapidly in case we need to shut down a feature when something goes wrong. Is there any code path that is not covered by `node_fuse_max_request_limit_kb`? If there is, then I think a feature flag would be required, if there isn't, then I think having the mount option act as the only feature flag is ok.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -46,9 +46,22 @@ const (
 	socketName                       = "socket"
 	readAheadKBMountFlagRegexPattern = "^read_ahead_kb=(.+)$"
 	readAheadKBMountFlag             = "read_ahead_kb"
+	// Hidden CSI flag never to be passed to gcsfuse
```
uriel-guzman, line 50: Why does the comment say "hidden"? Isn't it going to be public? Otherwise, how would customers know which mount option they are supposed to pass to the CSI Driver to enable the feature?

uriel-guzman (changes requested):

pkg/csi_mounter/csi_mounter.go
meet2mky: If I change this variable to minPageSizeKiB then the all other variable would become inconsistent. For e.g `readAheadKBMountFlag` is created has KB in it's name but actually it's taking value in KiB meaning if you set the CSI option - `read_ahead_kb=1024` it results in `1MiB => 1024 * KiB`. 

All our public flags are of the form .*_kb in both GKE and GCSFuse but in practice they always assume values in KiB units. Let me know if you have better suggestions.

pkg/csi_mounter/csi_mounter.go
meet2mky: yup consistency across GKE and also GCSFuse because all our external flags ends in .*_mb or .*_kb.

pkg/csi_mounter/csi_mounter.go
meet2mky: done

pkg/csi_mounter/csi_mounter.go
meet2mky: Thanks for catching it. Removed it.

test/e2e/testsuites/kernel_params.go
meet2mky: Thanks it's v3.11.1-gke.0 fixed the comment.

pkg/csi_mounter/csi_mounter.go
abhishek10004: The feature would be gated by --enable-kernel-reader gcsfuse flag which is public. This flag is an internal hidden flag (this allows us to change the max pages limit that would be set in case of elevation; can also be set to 0 to disable the elevation logic).

deploy/base/node/node.yaml
meet2mky: Consider following scenarios for CSI Driver:


* [Current Release] Fuse Max pages elevation is OPT IN and only elevated to 16MiB where customer explicitly opts for GCSFuse kernel reader feature for regional bucket by specifying the flag `--enable-kernel-reader=true`.

* [Next Release] Fuse Max pages is by default elevated for certain workloads/machine type in GKE CSI Driver. Now it becomes OPT OUT feature for these scenario where either customer can specify `--enable-kernel-reader=false` or `--node_fuse_max_request_limit_kb=0`.
   * If customer specifies `--enable-kernel-reader=false` then feature is disabled and fuse max pages elevation is also disabled in GKE CSI Driver.
   * If customer specifies `--node_fuse_max_request_limit_kb=0` then only the fuse max pages elevation logic is disabled but customer can still use the kernel-reader-feature on regional bucket (it would be suboptimal performance from GCSFuse in this case).

Note: In future, If customer other pods or custom Fuse mounts or anything faces any issue then `--node_fuse_max_request_limit_kb=0` will disable the fuse max pages update logic completely in CSI Driver.
   
   
Let me know if there is anything required.

CC @abhishek10004

test/e2e/specs/specs.go
```diff
@@ -1145,6 +1145,11 @@ func (t *TestPod) VerifyDefaultingFlagsArePassed(namespace string, expectedMachi
 		"Should find MachineType flag string %q in stdout", expectedMachineTypeFlagString)
 }
 
+func (t *TestPod) GetSidecarLogs() (string, error) {
```
abhishek10004, line 1148: Why is this needed?

test/e2e/testsuites/kernel_params.go
```diff
@@ -40,15 +43,24 @@ import (
...
-	MaxReadAheadKb               ParamName = "max-read-ahead-kb"
-	MaxBackgroundRequests        ParamName = "fuse-max-background-requests"
-	CongestionWindowThreshold    ParamName = "fuse-congestion-window-threshold"
-	CustomCSIReadAhead3MB                  = "3072"
-	CustomGCSFuseMaxReadAhead6MB           = "6144"
-	CustomGCSFuseMaxReadAhead8MB           = "8192"
-	CustomMaxBackground                    = "999"
-	CustomCongestionThreshold              = "666"
-	tmpMountPath                           = "/gcsfuse-tmp"
+	KiB = 1024
```
abhishek10004: Can we not use the constants from util package here?

test/e2e/specs/specs.go
meet2mky: I get the logs from sidecar to validate I/O size.

test/e2e/testsuites/kernel_params.go
meet2mky: Good Idea. Yes

abhishek10004 (approved):
LGTM

deploy/base/node/node.yaml
uriel-guzman: I initially thought `node_fuse_max_request_limit_kb` was the feature flag. Since this PR does nothing without `--enable-kernel-reader` being `true`, and there's a way to turn off fuse max page limit logic with `--node_fuse_max_request_limit_kb=0`, then I agree a CSI Driver feature flag is unnecessary.

pkg/csi_mounter/csi_mounter.go
uriel-guzman: ok, I initially thought this was the feature flag, so that's why I was asking. Kernel reader flag (`--enable-kernel-reader`) is the feature flag, and it will be public.

pkg/csi_mounter/csi_mounter.go
uriel-guzman: I see, I think as long as we keep consistency it's okay. The units (`util.MiB` and `util.KiB`) already make it clear that we're using binary units and not decimal units.

uriel-guzman (approved):

uriel-guzman (changes requested):
please fix failing build check

uriel-guzman (approved):

pkg/csi_mounter/csi_mounter.go
```diff
@@ -406,15 +432,44 @@ func prepareMountOptions(options []string) ([]string, []string, map[string]int64
...
+			nodeFuseMaxRequestLimitKB, err := strconv.ParseInt(maxReqSize[1], 10, 0)
+			if err != nil {
+				return nil, nil, nil, 0, fmt.Errorf("invalid %s mount flag %q: %w", nodeFuseMaxRequestLimitKBMountFlag, o, err)
+			}
+			if nodeFuseMaxRequestLimitKB < 0 {
+				return nil, nil, nil, 0, fmt.Errorf("invalid negative value for %s mount flag: %q", nodeFuseMaxRequestLimitKBMountFlag, o)
+			}
+			maxAllowedSizeKB := maxFuseMaxPagesLimit * pageSizeKB
+			if nodeFuseMaxRequestLimitKB > maxAllowedSizeKB {
+				return nil, nil, nil, 0, fmt.Errorf("invalid value for %s mount flag %q: exceeds maximum allowed limit of %d KB on this host", nodeFuseMaxRequestLimitKBMountFlag, o, maxAllowedSizeKB)
```
amacaskill: I'm confused. We are saying KB here and in the variables, but its actually KiB, correct?

The error message should definitely be updated to say "%d KiB" instead of "%d KB" for exactness. 

The variable names I think are also off, correct? Ideally those would be corrected to say pageSizeKB-> pageSizeKiB, but you can instead add a comment at the top just explaining its wrong//means something else if you don't want to change it all, however, anything user facing, should be correct (like the error messages) otherwise its really confusing

amacaskill: Same for feature flag, please make sure that's the correct unit

pkg/csi_mounter/csi_mounter.go
amacaskill: Ok, can we add a comment explaining this somewhere? I think its weird all the variables are just different unit than what they actually mean. I find it really confusing. Flags for consistently can be _kb, but variables really shouldn't. But as I mentioned in other comment, you can add a comment explaining this, but what you really do need to fix is user facing errors . Those need to say the correct unit

pkg/util/kernel_params_monitoring.go
```diff
@@ -183,3 +211,35 @@ func MonitorKernelParamsFile(ctx context.Context, mountPoint, emptyDirBasePath,
...
+		return fn()
+	}
+
+	klog.Infof("%v Temporarily increasing host FUSE max_pages_limit from %d to %d for mount", logPrefix, origLimit, targetLimit)
+	if err := SetFuseMaxPagesLimit(targetLimit); err != nil {
+		klog.Warningf("%v Failed to set host FUSE max_pages_limit to %d: %v. Proceeding with default kernel limit.", logPrefix, targetLimit, err)
+		return fn()
+	}
+
+	defer func() {
```
amacaskill, line 240: Please add the note that because we blindly restore origLimit, if a cluster administrator or configuration daemon manually writes a new rule to /proc/sys/fs/fuse/max_pages_limit exactly during the milliseconds that fn() executes, our defer will silently overwrite and erase their changes as it winds down.

We can prevent this by verifying that the limit hasn't changed from our targetLimit before we blindly overwrite it, and please add a TOOD that this is implemented before default on. Please write this down in your design doc also.

meet2mky: I have added a TODO. We can discuss this in detail in design for Default ON scenario.

pkg/csi_mounter/csi_mounter.go
meet2mky: Thanks for the feedback, I have updated the error string.

Note: I have added a comment on top of variables that they are all in KiB.
I understand that all of this is wrong when we say the values are indeed in KiB but the units or flag names say simply "_kb" or variable names say "KB". However, this is already an existing pattern in the codebase and the existing public flags in both GKE and GCSFuse. We can't really change any public existing flags. We can only change the codebase for better understanding in Devs. 

Now in the codebase changing at one place few variables in KiB and while at some places other variables stay in KB creates a very inconsistent pattern causing more confusion as already stated in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1514#discussion_r3601056423.

If variable renames is really required across codebase then it can be done in separate refactoring PR which ensures consistent view of codebase be it KiB or KB.

amacaskill: I agree its a little late now to change, as long as user facing errors / public documentation explains this, the variables are fine for now. thanks

PR #1530, opened 2026-07-14
[Shared Mount] Implement MountToNode

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Implements mountToNode for shared node mount and prevents a sidecar specific logic from running in shared node mount.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -1025,6 +1025,10 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 
 	// Send GRPC to mounter pod to start GCSFuse.
 	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, args.fuseMountOptions); err != nil {
+		klog.Errorf("Failed to mount volume %q to staging path %q: %v", volumeID, stagingPath, err)
+		if code, err := checkMounterPodErrorFile(emptyDirBasePath); err != nil {
+			return nil, status.Error(code, err.Error())
+		}
 		return nil, err
 	}
 
--- a/pkg/sidecar_mounter/mounter_server.go
+++ b/pkg/sidecar_mounter/mounter_server.go
@@ -32,12 +32,18 @@ type MounterServer struct {
 	mounter.UnimplementedMounterServer
 	mounter   *Mounter
 	serverCtx context.Context
+	tmpDir    string
+	bufferDir string
+	cacheDir  string
 }
 
 func NewMounterServer(ctx context.Context, mounter *Mounter) *MounterServer {
 	return &MounterServer{
 		mounter:   mounter,
 		serverCtx: ctx,
+		tmpDir:    webhook.SidecarContainerTmpVolumeMountPath,
+		bufferDir: webhook.SidecarContainerBufferVolumeMountPath,
+		cacheDir:  webhook.SidecarContainerCacheVolumeMountPath,
 	}
 }
 
@@ -57,11 +63,11 @@ func (ms *MounterServer) Mount(ctx context.Context, req *mounter.MountRequest) (
 		BucketName:          util.ParseVolumeID(req.GetVolumeId()),
 		Options:             req.GetMountOptions(),
 		SharedMountPoint:    req.GetMountPoint(),
-		TempDir:             webhook.SidecarContainerTmpVolumeMountPath,
-		ErrWriter:           NewErrorWriter(filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, util.ErrorFileName)),
-		BufferDir:           webhook.SidecarContainerBufferVolumeMountPath,
-		CacheDir:            webhook.SidecarContainerCacheVolumeMountPath,
-		ConfigFile:          filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, "config.yaml"),
+		TempDir:             ms.tmpDir,
+		ErrWriter:           NewErrorWriter(filepath.Join(ms.tmpDir, util.ErrorFileName)),
+		BufferDir:           ms.bufferDir,
+		CacheDir:            ms.cacheDir,
+		ConfigFile:          filepath.Join(ms.tmpDir, "config.yaml"),
 		AutoGoMemLimitRatio: util.GoMemLimitCgroupPercentage,
 	}
 
--- a/pkg/sidecar_mounter/sidecar_mounter.go
+++ b/pkg/sidecar_mounter/sidecar_mounter.go
@@ -56,6 +56,8 @@ import (
 
 const metricEndpointFmt = "http://localhost:%v/metrics"
 
+var gracefulKillGracePeriod = 5 * time.Second
+
 // Mounter will be used in the sidecar container to invoke gcsfuse.
 type Mounter struct {
 	mounterPath           string
@@ -113,28 +115,29 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 
 	klog.Infof("start to mount bucket %q for volume %q", mc.BucketName, mc.VolumeName)
 
-	if err := os.MkdirAll(mc.BufferDir+TempDir, os.ModePerm); err != nil {
-		return fmt.Errorf("failed to create temp dir %q: %w", mc.BufferDir+TempDir, err)
+	if err := createBufferTempDir(mc); err != nil {
+		return err
 	}
 
-	args := []string{}
-	for k, v := range mc.FlagMap {
-		arg := k
-		if v != "" {
-			arg = fmt.Sprintf("%s=%s", k, v)
-		}
-		args = append(args, fmt.Sprintf("--%s", arg))
-	}
-
-	args = append(args, mc.BucketName)
 	// gcsfuse supports the `/dev/fd/N` syntax
 	// the /dev/fuse is passed as ExtraFiles below, and will always be FD 3
-	args = append(args, "/dev/fd/3")
-
-	cmd := m.prepareMountCommand(ctx, mc, args)
+	cmd := m.prepareMountCommand(ctx, mc, "/dev/fd/3")
 	cmd.ExtraFiles = []*os.File{os.NewFile(uintptr(mc.FileDescriptor), "/dev/fuse")}
 
-	startCleanupMonitor(ctx, cmd)
+	// when the ctx.Done() is closed,
+	// the main workload containers have exited,
+	// so it is safe to force kill the gcsfuse process.
+	go func(cmd *exec.Cmd) {
+		<-ctx.Done()
+		klog.V(4).Infof("context marked as done, sleep for 5 seconds, to evaluate gcsfuse process %d exit state", cmd.Process.Pid)
+		time.Sleep(time.Second * 5)
+		if cmd.ProcessState == nil || !cmd.ProcessState.Exited() {
+			klog.Warningf("after 5 seconds, process with id %v has not exited, force kill the process", cmd.Process.Pid)
+			if err := cmd.Process.Kill(); err != nil {
+				klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
+			}
+		}
+	}(cmd)
 
 	m.WaitGroup.Add(1)
 	go func() {
@@ -147,65 +150,156 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 
 		klog.Infof("gcsfuse for bucket %q, volume %q started with process id %v", mc.BucketName, mc.VolumeName, cmd.Process.Pid)
 
-		loggingSeverity := mc.ConfigFileFlagMap["logging:severity"]
-		if loggingSeverity == "debug" || loggingSeverity == "trace" {
-			go logMemoryUsage(ctx, cmd.Process.Pid)
-			go logVolumeUsage(ctx, mc.BufferDir, mc.CacheDir)
-		}
-
-		promPort, ok := mc.FlagMap["prometheus-port"]
-		if ok && promPort != "0" {
-			klog.Infof("start to collect metrics from port %v for volume %q", promPort, mc.VolumeName)
-			go collectMetrics(ctx, promPort, mc.TempDir)
-		}
+		startVolumeMonitoring(ctx, mc, cmd.Process.Pid)
 
 		// Since the gcsfuse has taken over the file descriptor,
 		// closing the file descriptor to avoid other process forking it.
 		syscall.Close(mc.FileDescriptor)
-		if err := cmd.Wait(); err != nil {
-			errMsg := fmt.Sprintf("gcsfuse exited with error: %v\n", err)
-			if strings.Contains(errMsg, "signal: terminated") {
-				klog.Infof("[%v] gcsfuse was terminated.", mc.VolumeName)
-			} else if strings.Contains(errMsg, "signal: killed") {
-				klog.Infof("[%v] gcsfuse was killed.", mc.VolumeName)
-			} else {
-				mc.ErrWriter.WriteMsg(errMsg)
-			}
-		} else {
-			klog.Infof("[%v] gcsfuse exited normally.", mc.VolumeName)
-		}
+		waitForGcsfuseExit(cmd, mc)
 	}()
 
 	return nil
 }
 
 // MountToNode uses the mount config to initialize the GCSFuse process for a share node mount architecture.
-// gcsfuseContext is the long running context from the sidecar, this prevents the gcsfuse process from being killed when the NodeStageVolume context (the ctx variable) is canceled.
-func (m *Mounter) MountToNode(ctx context.Context, gcsfuseContext context.Context, mc *MountConfig) error {
-	return status.Errorf(codes.Unimplemented, "method MountToNode not implemented")
+// mountServerContext is the long running context from the sidecar, using this prevents the gcsfuse process from being killed when the NodeStageVolume context (the ctx variable) is canceled.
+func (m *Mounter) MountToNode(ctx context.Context, mountServerContext context.Context, mc *MountConfig) error {
+	if mc == nil {
+		return status.Error(codes.Internal, "mount config cannot be nil")
+	}
+	if mc.SharedMountPoint == "" {
+		return status.Error(codes.Internal, "shared mount point cannot be empty")
+	}
+	mc.EnsureErrWriter()
+
+	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
+
+	klog.Infof("Start mounting GCS bucket %q", mc.BucketName)
+
+	if err := createBufferTempDir(mc); err != nil {
+		return status.Errorf(codes.Internal, "failed to create buffer temp dir: %v", err)
+	}
+
+	// processCtx is derived from the long-running sidecar container context
+	// (mountServerContext) to tie the gcsfuse daemon lifecycle to the pod container
+	// rather than the short-lived Mount gRPC call. If MountToNode fails (!success),
+	// processCancel is called to shut down the gcsfuse process.
+	processCtx, processCancel := context.WithCancel(mountServerContext)
+	var success bool
+	defer func() {
+		if !success {
+			processCancel()
+		}
+	}()
+
+	cmd := m.prepareMountCommand(processCtx, mc, mc.SharedMountPoint)
+
+	klog.Infof("GCSFuse CSI Driver starting gcsfuse in node mounter...")
+	if err := cmd.Start(); err != nil {
+		return status.Errorf(codes.Internal, "failed to start gcsfuse with error: %v", err)
+	}
+
+	klog.Infof("gcsfuse for bucket %q, volume %q started with process id %v", mc.BucketName, mc.VolumeName, cmd.Process.Pid)
+
+	// cmdDone is closed when waitForGcsfuseExit returns, signaling to forceKillCmd
+	// that the gcsfuse process has exited so it can return immediately without
+	// waiting out the grace period or risking PID reuse.
+	cmdDone := make(chan struct{})
+	// pollCtx is derived from the short-lived Mount gRPC request context (ctx)
+	// and bounds the polling loop in util.WaitForPathMounted so polling stops
+	// if the gRPC request times out, is canceled, or if gcsfuse exits early.
+	pollCtx, pollCancel := context.WithCancel(ctx)
+	defer pollCancel()
+
+	m.WaitGroup.Add(1)
+	go func() {
+		defer m.WaitGroup.Done()
+		waitForGcsfuseExit(cmd, mc)
+		// Broadcast that gcsfuse has exited so forceKillCmd does not call cmd.Process.Kill()
+		// on an already reaped process or a recycled PID.
+		close(cmdDone)
+		pollCancel()
+		processCancel()
+	}()
+
+	// Kill the gcsfuse process if the path has not been mounted within a timeout.
+	// This allows the next Mount attempt to start a fresh gcsfuse process.
+	if err := util.WaitForPathMounted(pollCtx, mc.SharedMountPoint); err != nil {
+		klog.Errorf("gcsfuse process did not mount path %q within the context timeout: %v, killing gcsfuse process with id %d", mc.SharedMountPoint, err, cmd.Process.Pid)
+		// Terminate the gcsfuse process asynchronously to avoid blocking the gRPC response.
+		go func() {
+			forceKillCmd(cmd, cmdDone)
+		}()
+		return err
+	}
+
+	startVolumeMonitoring(mountServerContext, mc, cmd.Process.Pid)
+
+	// Make the gcsfuse process depend on the life-cycle of the mounter pod's container context.
+	// This allows the gcsfuse process to be kept alive until the container is killed, as
+	// opposed to depending on the life-cycle of the Mount gRPC call.
+	go func() {
+		<-processCtx.Done()
+		if mountServerContext.Err() != nil {
+			forceKillCmd(cmd, cmdDone)
+		}
+	}()
+	// Successfully mounted the path, so the gcsfuse process is now tied to the long-running sidecar container context.
+	success = true
+	return nil
 }
 
-func startCleanupMonitor(ctx context.Context, cmd *exec.Cmd) {
-	// when the ctx.Done() is closed,
-	// the main workload containers have exited,
-	// so it is safe to force kill the gcsfuse process.
-	go func(cmd *exec.Cmd) {
-		<-ctx.Done()
+// forceKillCmd will force kill the process for a cmd
+// once cmdDone is closed.
+func forceKillCmd(cmd *exec.Cmd, cmdDone <-chan struct{}) {
+	if cmd == nil || cmd.Process == nil {
+		return
+	}
 
-		klog.V(4).Infof("context marked as done, sleep for 5 seconds, to evaluate gcsfuse process %d exit state", cmd.Process.Pid)
-		time.Sleep(time.Second * 5)
+	select {
+	case <-cmdDone:
+		// gcsfuse exited cleanly or responded to SIGTERM within the grace period.
+		// Return immediately without calling cmd.Process.Kill() to prevent PID reuse issues.
+		return
+	case <-time.After(gracefulKillGracePeriod):
+		// Grace period expired and gcsfuse has not exited; force kill the process.
+		klog.Warningf("after 5 seconds, process with id %v has not exited, force kill the process", cmd.Process.Pid)
 
-		if cmd.ProcessState == nil || !cmd.ProcessState.Exited() {
-			klog.Warningf("after 5 seconds, process with id %v has not exited, force kill the process", cmd.Process.Pid)
-			if err := cmd.Process.Kill(); err != nil {
-				klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
-			}
+		if err := cmd.Process.Kill(); err != nil {
+			klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
 		}
-	}(cmd)
+	}
+}
+
+// waitForGcsfuseExit waits for the gcsfuse process to exit and logs the exit state.
+func waitForGcsfuseExit(cmd *exec.Cmd, mc *MountConfig) {
+	if err := cmd.Wait(); err != nil {
+		errMsg := fmt.Sprintf("gcsfuse exited with error: %v\n", err)
+		if strings.Contains(errMsg, "signal: terminated") {
+			klog.Infof("[%v] gcsfuse was terminated.", mc.VolumeName)
+		} else if strings.Contains(errMsg, "signal: killed") {
+			klog.Infof("[%v] gcsfuse was killed.", mc.VolumeName)
+		} else {
+			mc.ErrWriter.WriteMsg(errMsg)
+		}
+	} else {
+		klog.Infof("[%v] gcsfuse exited normally.", mc.VolumeName)
+	}
 }
 
 // prepareMountCommand prepares the exec.Cmd with optional NUMA node binding using numactl.
-func (m *Mounter) prepareMountCommand(ctx context.Context, mc *MountConfig, args []string) *exec.Cmd {
+func (m *Mounter) prepareMountCommand(ctx context.Context, mc *MountConfig, mountPoint string) *exec.Cmd {
+	args := []string{}
+	for k, v := range mc.FlagMap {
+		arg := k
+		if v != "" {
+			arg = fmt.Sprintf("%s=%s", k, v)
+		}
+		args = append(args, fmt.Sprintf("--%s", arg))
+	}
+
+	args = append(args, mc.BucketName)
+	args = append(args, mountPoint)
 	cmdName := m.mounterPath
 	if mc.GcsFuseNumaNode >= 0 && numactlAllowed(ctx, mc.GcsFuseNumaNode) {
 		// A nonnegative numa node has been calculated by the csi driver by looking at network
@@ -270,6 +364,32 @@ func numactlAllowed(ctx context.Context, node int) bool {
 	}
 	klog.Infof("numactl checks out")
 	return true
+}
+
+// createBufferTempDir creates the buffer temporary directory for the volume mount.
+func createBufferTempDir(mc *MountConfig) error {
+	bufferTmpDir := filepath.Join(mc.BufferDir, TempDir)
+	if err := os.MkdirAll(bufferTmpDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create temp dir %q: %w", bufferTmpDir, err)
+	}
+
+	return nil
+}
+
+// startVolumeMonitoring starts background goroutines for memory usage logging, volume usage logging,
+// and Prometheus metrics collection if configured.
+func startVolumeMonitoring(ctx context.Context, mc *MountConfig, pid int) {
+	loggingSeverity := mc.ConfigFileFlagMap["logging:severity"]
+	if loggingSeverity == "debug" || loggingSeverity == "trace" {
+		go logMemoryUsage(ctx, pid)
+		go logVolumeUsage(ctx, mc.BufferDir, mc.CacheDir)
+	}
+
+	promPort, ok := mc.FlagMap["prometheus-port"]
+	if ok && promPort != "0" {
+		klog.Infof("start to collect metrics from port %v for volume %q", promPort, mc.VolumeName)
+		go collectMetrics(ctx, promPort, mc.TempDir)
+	}
 }
 
 // logMemoryUsage logs gcsfuse process VmRSS (Resident Set Size) usage every 30 seconds.
--- a/pkg/sidecar_mounter/sidecar_mounter_config.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_config.go
@@ -378,8 +378,13 @@ func (mc *MountConfig) prepareMountArgs() {
 		default:
 			klog.Warningf("got invalid value for %q: %q. Will discard and continue to mount.", util.FileCacheMediumConst, mc.FileCacheMedium)
 		}
-		// Ensure a unique directory is passed for each volume.
-		cacheDir = filepath.Join(cacheDir, ".volumes", mc.VolumeName)
+
+		isSidecarMounted := mc.SharedMountPoint == ""
+		// Ensure a unique directory is passed for each volume for sidecar architecture.
+		if isSidecarMounted {
+			cacheDir = filepath.Join(cacheDir, ".volumes", mc.VolumeName)
+		}
+
 		configFileFlagMap["cache-dir"] = cacheDir
 		mc.CacheDir = cacheDir
 		klog.Infof("Overriding cache-dir with %q for medium %q", cacheDir, mc.FileCacheMedium)
--- a/pkg/sidecar_mounter/sidecar_mounter_test.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_test.go
@@ -1,0 +1,434 @@
+/*
+Copyright 2026 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sidecarmounter
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestMain(m *testing.M) {
+	gracefulKillGracePeriod = 10 * time.Millisecond
+	os.Exit(m.Run())
+}
+
+func TestCreateBufferTempDir(t *testing.T) {
+	t.Parallel()
+
+	tempDir := t.TempDir()
+	blockingFile := filepath.Join(tempDir, "blocking-file")
+	if err := os.WriteFile(blockingFile, []byte("block"), 0644); err != nil {
+		t.Fatalf("failed to write blocking file: %v", err)
+	}
+
+	testCases := []struct {
+		name        string
+		mc          *MountConfig
+		expectedErr bool
+	}{
+		{
+			name: "should create valid buffer temp dir",
+			mc: &MountConfig{
+				BufferDir: tempDir,
+			},
+			expectedErr: false,
+		},
+		{
+			name: "should fail when buffer dir path is blocked by a file",
+			mc: &MountConfig{
+				BufferDir: blockingFile,
+			},
+			expectedErr: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := createBufferTempDir(tc.mc)
+			if (err != nil) != tc.expectedErr {
+				t.Errorf("Got error %v, but expected error %v", err, tc.expectedErr)
+			}
+			if !tc.expectedErr {
+				expectedDir := filepath.Join(tc.mc.BufferDir, TempDir)
+				info, err := os.Stat(expectedDir)
+				if err != nil || !info.IsDir() {
+					t.Errorf("expected directory %q to exist, got err: %v", expectedDir, err)
+				}
+			}
+		})
+	}
+}
+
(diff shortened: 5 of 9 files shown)


pkg/csi_driver/node.go
```diff
@@ -613,17 +613,7 @@ func (s *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpub
 
 // isDirMounted checks if the path is already a mount point.
```
uriel-guzman, line 580: Instead of doing all this refactoring and needing to create a new mounter object, you can just copy Parallelstore's `isDirMounted` into util.go?

```

func isDirMounted(targetPath string) (bool, error) {
--
mps, err := mount.ListProcMounts(procMountsPath)
if err != nil {
return false, err
}
for _, m := range mps {
if m.Path == targetPath {
return true, nil
}
}
 
return false, nil
}
```
Not sure why we're not copying stuff from Parallelstore if it already works?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,16 +184,115 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
 // MountToNode uses the mount config to initialize the GCSFuse process for a share node mount architecture.
 // gcsfuseContext is the long running context from the sidecar, this prevents the gcsfuse process from being killed when the NodeStageVolume context (the ctx variable) is canceled.
-func (m *Mounter) MountToNode(ctx context.Context, gcsfuseContext context.Context, mc *MountConfig) error {
-	return status.Errorf(codes.Unimplemented, "method MountToNode not implemented")
+func (m *Mounter) MountToNode(ctx context.Context, mountServerContext context.Context, mc *MountConfig) error {
+	mc.EnsureErrWriter()
+
+	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
+
+	// If enabled perform a bucket access check for shared node mount.
```
uriel-guzman: Was this supposed to be a TODO? If so, make it a TODO to implement bucket access check. Otherwise, it's not clear what this comment means?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,16 +184,115 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	// Add bucket access check.
+
+	// Clear the GCS Fuse error file after the  bucket access check completes and just before the GCS Fuse mount starts.
+	// This ensures the NodePublishVolume/NodeStageVolume will always fail if the error persists until the GCS Fuse mount actually starts.
+	if err := mc.ErrWriter.Clean(); err != nil {
```
uriel-guzman: This seems to have been copy and pasted from https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/sidecar_mounter/sidecar_mounter.go#L110, but it's actually tied to the sidecar bucket access check logic, which isn't even implemented yet? So, this line of code doesn't make sense in the context of this PR. Please remove, and implement it together with sidecar bucket access check PR.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,16 +184,115 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	}
+
+	klog.Infof("start to mount bucket %q", mc.BucketName)
+
+	bufferTmpDir := filepath.Join(mc.BufferDir, TempDir)
+	if err := os.MkdirAll(bufferTmpDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create temp dir %q: %w", bufferTmpDir, err)
+	}
+
+	args := []string{}
```
uriel-guzman: The only purpose of `args` in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/sidecar_mounter/sidecar_mounter.go#L120 is to be passed into `prepareMountCommand`. After that, `args` is never used again. The only difference with shared mount would be that instead of `"/dev/fd/3"`, it seems like you're just passing `SharedMountPoint`, other than that, everything is exactly the same.

This means that you can move all of the `args` related logic inside of `prepereMountCommand` and just create a new parameter called `extraArgs`. In `Mount`, you pass `/dev/fd/3` as `extraArgs`, and in `MountToNode`, you pass `mc.SharedMountPoint`. You already refactored the code, so you might as well finish it by doing it cleanly this way.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,16 +184,115 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	for k, v := range mc.FlagMap {
+		arg := k
+		if v != "" {
+			arg = fmt.Sprintf("%s=%s", k, v)
+		}
+		args = append(args, fmt.Sprintf("--%s", arg))
+	}
+
+	args = append(args, mc.BucketName)
+	args = append(args, mc.SharedMountPoint)
```
uriel-guzman: I am unclear how this works. In Parallelstore, we had to pass `/usr/bin/dfuse` as an argument, I would expect you also have to pass `/usr/bin/gcsfuse` or something like that here? Otherwise, how does the mounter know it's mounting a GCSFuse mount? See Parallelstore's code in sidecar_mounter.go:

```
cmd := exec.CommandContext(ms.ctx, "/usr/bin/dfuse",
		"-m", mc.MountPoint,
		"--pool", mc.PoolName,
		"--container", mc.ContainerName,
		"--foreground",
		"--multi-user",
	)
```

How did you verify this works?

chrisThePattyEater: gcsfuse path get passed to mounterPath https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/2570ab8458088cc91a1f3a2c75fa144392431174/cmd/sidecar_mounter/main.go#L100

mounterPath gets used by prepareMountCommand to create the gcsfuse command https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/2570ab8458088cc91a1f3a2c75fa144392431174/pkg/sidecar_mounter/sidecar_mounter.go#L322-L323 (cmdName := m.mounterPath)

I tested this by using this config file and verifying the gcsfuse process starts and node stage volume completes (which means the staging path was mounted )
```
apiVersion: v1
kind: PodTemplate
metadata:
  name: mounter-template-test
  namespace: default # <-- CHANGE THIS if using a custom namespace
template:
  spec:
    serviceAccountName: default # <-- ADD THIS (KSA with storage access)
    containers:
    - name: gcsfusecsi-mount
      image: gcr.io/fuechr-gke-dev/gcs-fuse-csi-driver-sidecar-mounter:testMounterPod21
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: gcs-shared-pv
spec:
  accessModes:
  - ReadWriteMany
  capacity:
    storage: 5Gi
  storageClassName: ""
  # Binds the PV specifically to your PVC so nothing else can steal it
  claimRef: # <-- ADD THIS (Highly Recommended)
    name: gcs-shared-pvc
    namespace: default # <-- CHANGE THIS if using a custom namespace
  csi:
    volumeHandle: mr_chris_test # <-- Your bucket name
    volumeAttributes:
      sharedMount: "true"
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: gcs-shared-pvc
  namespace: default # <-- CHANGE THIS if using a custom namespace
  annotations:
    gke-gcsfuse/mounter-pod-template: "mounter-template-test"
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 5Gi
  volumeName: gcs-shared-pv
  storageClassName: ""
---
apiVersion: v1
kind: Pod
metadata:
  name: gcs-shared-workload
  namespace: default # <-- CHANGE THIS if using a custom namespace
  annotations:
    gke-gcsfuse/volumes: "true"
spec:
  serviceAccountName: default # <-- ADD THIS (KSA with storage access)
  containers:
  - name: my-app
    image: nginx:latest
    volumeMounts:
    - name: gcs-vol
      mountPath: /data
  volumes:
  - name: gcs-vol
    persistentVolumeClaim:
      claimName: gcs-shared-pvc
```

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,16 +184,115 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	}
+
+	return nil
 }
 
-func startCleanupMonitor(ctx context.Context, cmd *exec.Cmd) {
-	// when the ctx.Done() is closed,
-	// the main workload containers have exited,
+func startCleanupMonitor(ctx context.Context, cmd *exec.Cmd) func() {
+	cleanupCtx, cancel := context.WithCancel(ctx)
```
uriel-guzman: I don't understand this refactor. You're creating a new `cancel` inside here, just so it can be returned, but aren't actually using it for anything inside the helper function, only outside? The cancel for the context that is passed to this function is already going to cancel this context, so why not just create the `cancel` outside? The current refactor isn't ideal because it's shared with the gke-gcsfuse-sidecar.

uriel-guzman (changes requested):

pkg/sidecar_mounter/sidecar_mounter.go
uriel-guzman: Ah, I see. I didn't see that `/gcsfuse` is a flag in sidecar_mounter.go. Then that makes sense.

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: yeam i was tryign to resolve a bug I had due to trying to reuse the logic from cleanup monitor. i think i makes more sense to just directly call the kill command so I moved this to ahelper

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: did this

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: removed

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: removed

pkg/csi_driver/node.go
chrisThePattyEater: added this but also added clean as per some gemeini recs

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,16 +184,115 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+
+	go func() {
+		defer m.WaitGroup.Done()
+		if err := cmd.Start(); err != nil {
+			mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to start gcsfuse with error: %v\n", err))
+
+			return
+		}
+
+		klog.Infof("gcsfuse for bucket %q, volume %q started with process id %v", mc.BucketName, mc.VolumeName, cmd.Process.Pid)
```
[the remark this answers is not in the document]
chrisThePattyEater: we said we wouldn't do this since this is how the gcsfuse sidecar code is implemented. WE have a action item to review this logic since as per this comment it may have a bug

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,8 +169,97 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	klog.Infof("start to mount bucket %q", mc.BucketName)
+
+	bufferTmpDir := filepath.Join(mc.BufferDir, TempDir)
+	if err := os.MkdirAll(bufferTmpDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create temp dir %q: %w", bufferTmpDir, err)
+	}
+
+	cmd := m.prepareMountCommand(mountServerContext, mc, mc.SharedMountPoint)
+
+	startCleanupMonitor(mountServerContext, cmd)
```
uriel-guzman: I don't understand this logic. You are setting up `startCleanupMonitor` here, which calls `forceKillCmd`, which results in force killing the gcsfuse process after 5 seconds if graceful termination did not work, which is aligned with the sidecar behavior, but then, in L200, you're creating another context, and then you're calling `forceKillCmd` again, which again, waits for the context to be cancelled, and wait 5 seconds to force kill the gcsfuse process? Why is it doing the exact same logic two times?

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,8 +169,88 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+
+	klog.Infof("start to mount bucket %q", mc.BucketName)
+
+	bufferTmpDir := filepath.Join(mc.BufferDir, TempDir)
+	if err := os.MkdirAll(bufferTmpDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create temp dir %q: %w", bufferTmpDir, err)
+	}
+
+	// Use a new context for gcsfuse monitoring, so that the monitoring goroutines can be canceled when the mount server context is canceled.
+	gcsfuseCtx, gcsfuseCancel := context.WithCancel(mountServerContext)
```
uriel-guzman: Why are you creating a new `gcsfuseCancel`? The intention here seems to be to make it get cancelled when the go routine blocking on `cmd.Wait` on L217 calls `defer gcsfuseCancel()`, but this seems redundant, because the `ctx` is actually `serverCtx`, which is actually the `ctx` from `runNodeMounter`: https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/main/cmd/sidecar_mounter/main.go#L58, which already has a `cancel`, but you're just deciding to not use it, and then you're just creating a new one?

chrisThePattyEater: connected offline, will pass mounter server context everywhere in this pr and use node stage volume context for non gcsfuse process tasks such as bucket access check

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,8 +169,87 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		}
+
+		klog.Infof("gcsfuse for bucket %q, volume %q started with process id %v", mc.BucketName, mc.VolumeName, cmd.Process.Pid)
+
+		loggingSeverity := mc.ConfigFileFlagMap["logging:severity"]
+		if loggingSeverity == "debug" || loggingSeverity == "trace" {
+			go logMemoryUsage(gcsfuseCtx, cmd.Process.Pid)
+			go logVolumeUsage(gcsfuseCtx, mc.BufferDir, mc.CacheDir)
+		}
+
```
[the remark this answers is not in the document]
chrisThePattyEater: cache dir is always already populated by this point for shared mount

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,8 +169,84 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	cmd := m.prepareMountCommand(mountServerContext, mc, mc.SharedMountPoint)
+
+	startCleanupMonitor(mountServerContext, cmd)
+
+	m.WaitGroup.Add(1)
+	go func() {
+		defer m.WaitGroup.Done()
+		if err := cmd.Start(); err != nil {
+			mc.ErrWriter.WriteMsg(fmt.Sprintf("failed to start gcsfuse with error: %v\n", err))
```
uriel-guzman: If `cmd.Start()` failed, the GCSFuse process failed to even start. In GCSFuse sidecar, the error is written to the error file because it's only way the sidecar can communicate back to the CSI Driver that there was an issue starting GCSFuse. However, in `MountToNode`, you already have the capability of just returning the error back via the gRPC server, so there is absolutely no need to write to the error file. 

If you actually look at the parallelstore code what it's doing, it's doing exactly this:

```
if err := cmd.Start(); err != nil {
--
return status.Errorf(codes.Internal, "failed to start dfuse with error: %v", err)
}
```

This ensures that NodeStageVolume **immediately** fails if GCSFuse fails to start.

Also, move `cmd.Start()` outside of the go routine.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,8 +169,84 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+				klog.Infof("[%v] gcsfuse was killed.", mc.VolumeName)
+			} else {
+				mc.ErrWriter.WriteMsg(errMsg)
+			}
+		} else {
+			klog.Infof("[%v] gcsfuse exited normally.", mc.VolumeName)
+		}
+	}()
+
+	if err := util.WaitForPathMounted(ctx, mc.SharedMountPoint); err != nil {
```
uriel-guzman: Once you have moved `cmd.Start()` outside of the go routine, as I suggested in the comment: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1530/changes#r3605821936

The next step should be to actually check that the path is mounted, so you can call `WaitForPathMounted` immediately after you verified there's no error from `cmd.Start()`, since we don't care about starting any monitoring or cleanup go rotuines if the GCSFuse mount didn't even start and mount successfully in the first place. 

Take a look at the Parallelstore codebase, which already handles this correctly.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -181,8 +169,84 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		}
+		klog.Errorf("gcsfuse process did not mount path %q within the context timeout: %v, killing gcsfuse process with id %d", mc.SharedMountPoint, err, pid)
+		if cmd.Process != nil {
+			// First attempt to gracefully terminate gcsfuse by sending a SIGTERM signal.
+			klog.Infof("sending SIGTERM to gcsfuse: %v", cmd)
+			if sigErr := cmd.Process.Signal(syscall.SIGTERM); sigErr != nil {
+				klog.Errorf("error sending SIGTERM signal to gcsfuse: %v", sigErr)
+			}
+			// Trigger cleanup monitor to force kill gcsfuse after 5 seconds if graceful termination did not work.
+			go func(cmd *exec.Cmd) {
```
uriel-guzman: Why are we creating a new go routine here to force kill the GCSFuse process? If the path didn't mount, you should just try to force kill it synchronosuly? Not sure what a new go routine does for us here. See Parallelstore code:

```
// Kill the dfuse process if the path has not been mounted within a timeout.
--
// This allows the next Mount attempt to start a fresh dfuse process.
if err := util.WaitForPathMounted(ctx, mc.MountPoint); err != nil {
klog.Errorf("dfuse process did not mount path %q within the context timeout: %v, killing dfuse process with id %d", mc.MountPoint, err, cmd.Process.Pid)
 
// First attempt to gracefully terminate dfuse by sending a SIGTERM signal.
klog.Infof("sending SIGTERM to dfuse: %v", cmd)
if err := cmd.Process.Signal(syscall.SIGTERM); err != nil {
return status.Errorf(codes.Internal, "error sending SIGTERM signal to dfuse: %v", err)
}
 
// Force kill the dfuse process if the graceful termination did not work.
if err := forceKillCmd(cmd, ctx); err != nil {
return status.Errorf(codes.Internal, "failed to force kill dfuse process with id %d: %v", cmd.Process.Pid, err)
}
 
klog.Infof("dfuse with process id %d has been shut down", cmd.Process.Pid)
return err
}

```
Just follow that.

uriel-guzman (changes requested):

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -162,50 +156,140 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+			klog.Infof("gcsfuse with process id %d has been shut down", cmd.Process.Pid)
+		}()
+
+		return err
+	}
+
+	loggingSeverity := mc.ConfigFileFlagMap["logging:severity"]
+	if loggingSeverity == "debug" || loggingSeverity == "trace" {
+		go logMemoryUsage(mountServerContext, cmd.Process.Pid)
+		go logVolumeUsage(mountServerContext, mc.BufferDir, mc.CacheDir)
```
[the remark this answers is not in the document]
chrisThePattyEater: This seems like it would also be an issue in the sidecar architecture adding it to the bug but leaving it alone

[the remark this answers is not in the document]
chrisThePattyEater: also this is harcoded in shared node mount

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -162,50 +156,140 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	if cmd.ProcessState == nil || !cmd.ProcessState.Exited() {
+		klog.Warningf("after 5 seconds, process with id %v has not exited, force kill the process", cmd.Process.Pid)
+		if err := cmd.Process.Kill(); err != nil {
+			klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is an issue that also exists in sidecar, resolving and we can look at it later

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -162,50 +156,140 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-	go func(cmd *exec.Cmd) {
-		<-ctx.Done()
+	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
+
+	// TODO(FUECHR) Add bucket access check.
+
+	klog.Infof("Start mounting GCS bucket %q", mc.BucketName)
+
+	bufferTmpDir := filepath.Join(mc.BufferDir, TempDir)
+	if err := os.MkdirAll(bufferTmpDir, os.ModePerm); err != nil {
```
uriel-guzman: Refactor L176 - L179 into a helper function to share with the sidecar mounter and call it `createBufferDir`. If they ever change the permissions, it should probably be the same as both.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -162,50 +156,140 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		return err
+	}
+
+	loggingSeverity := mc.ConfigFileFlagMap["logging:severity"]
+	if loggingSeverity == "debug" || loggingSeverity == "trace" {
+		go logMemoryUsage(mountServerContext, cmd.Process.Pid)
+		go logVolumeUsage(mountServerContext, mc.BufferDir, mc.CacheDir)
+	}
+
+	promPort, ok := mc.FlagMap["prometheus-port"]
```
uriel-guzman: L220 to L230 are exactly the same as the ones in sidecar mounter. Refactor into a helper function called `startVolumeMonitoring` and call it from both.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -162,50 +156,140 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-		time.Sleep(time.Second * 5)
+	pollCtx, pollCancel := context.WithCancel(ctx)
+	defer pollCancel()
+
+	m.WaitGroup.Add(1)
+	go func() {
+		defer m.WaitGroup.Done()
+		waitForGcsfuseExit(cmd, mc)
+		pollCancel()
+	}()
```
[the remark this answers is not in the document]
uriel-guzman: This is a solid suggestion. Since `processCtx` is a child context of `mountServerContext`, the forceKillCmd would be called either when the sidecar receives a SIGTERM (current behavior), but also if the gcsfuse process exits (processCancel() is called), ensuring we don't cause memory leaks on failed mounts.

pkg/sidecar_mounter/sidecar_mounter.go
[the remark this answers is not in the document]
uriel-guzman: I don't know how this is possible. `cmd.Start()` is called before `forceKillCmd()`. If `cmd.Start()` errors out, it returns immediately and never reaches `forceKillCmd(). The only way it could reach it, is if `cmd.Start()` somehow fails after `cmd.Start()` returned nil error.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -162,50 +156,140 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	// opposed to depending on the life-cycle of the Mount gRPC call.
+	go func() {
+		<-mountServerContext.Done()
+		if err := forceKillCmd(cmd); err != nil {
+			klog.Errorf("Failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
+			return
 		}
-	}(cmd)
+		klog.Infof("gcsfuse with process id %d has been shut down", cmd.Process.Pid)
+	}()
```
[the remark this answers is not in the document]
uriel-guzman: Apply fix as part of the other comment: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1530/#discussion_r3616700035

pkg/sidecar_mounter/sidecar_mounter.go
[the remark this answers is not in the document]
uriel-guzman: I see. The bug is specifically for the sidecar mounter, not on the shared mount. Since you just refactored the code but didn't change sidecar mounter at all, then we can ignore this comment.

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -378,8 +378,13 @@ func (mc *MountConfig) prepareMountArgs() {
...
 			klog.Warningf("got invalid value for %q: %q. Will discard and continue to mount.", util.FileCacheMediumConst, mc.FileCacheMedium)
 		}
-		// Ensure a unique directory is passed for each volume.
-		cacheDir = filepath.Join(cacheDir, ".volumes", mc.VolumeName)
+
+		isSidecarMounted := mc.SharedMountPoint == ""
+		// Ensure a unique directory is passed for each volume for sidecar architecture.
+		if isSidecarMounted {
+			cacheDir = filepath.Join(cacheDir, ".volumes", mc.VolumeName)
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: This is not what is meant by shared node mount "GCSFuse instances on the node will share the exact same cache-dir"

pkg/sidecar_mounter/sidecar_mounter_test.go
```diff
@@ -0,0 +1,417 @@
...
+				SharedMountPoint: unmountedPath,
+				BufferDir:        t.TempDir(),
+				TempDir:          t.TempDir(),
+				FlagMap:          map[string]string{},
+			},
+			timeout:     200 * time.Millisecond,
+			acceptCodes: []codes.Code{codes.DeadlineExceeded, codes.Canceled},
+			expectError: true,
+		},
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: added

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -113,28 +113,24 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-	startCleanupMonitor(ctx, cmd)
+	// when the ctx.Done() is closed,
+	// the main workload containers have exited,
+	// so it is safe to force kill the gcsfuse process.
+	go func(cmd *exec.Cmd) {
+		<-ctx.Done()
+		if err := forceKillCmd(cmd); err != nil {
+			klog.Warningf("failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
+		}
+	}(cmd)
```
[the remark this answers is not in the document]
chrisThePattyEater: this is sidecar code leaving it alone

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +145,153 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	time.Sleep(forceKillGracePeriod)
+	// To avoid a data race on cmd.ProcessState, we can directly attempt to kill the process.
+	// If the process has already exited, Kill() will return an error (like os.ErrProcessDone) which we safely ignore.
+	klog.Warningf("after 5 seconds, force killing the process with id %v", cmd.Process.Pid)
+	if err := cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) && !strings.Contains(err.Error(), "bad file descriptor") && !strings.Contains(err.Error(), "process already finished") {
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: sigh, i think this is valid adding

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +145,153 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		<-processCtx.Done()
+		if mountServerContext.Err() != nil {
+			if err := forceKillCmd(cmd); err != nil {
+				klog.Warningf("Failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
+				return
 			}
+			klog.Infof("gcsfuse with process id %d has been shut down", cmd.Process.Pid)
 		}
-	}(cmd)
+	}()
```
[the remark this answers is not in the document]
chrisThePattyEater: this is the part two for the one above

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -113,28 +115,24 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	// so it is safe to force kill the gcsfuse process.
+	go func(cmd *exec.Cmd) {
+		<-ctx.Done()
+		if err := forceKillCmd(cmd); err != nil {
+			klog.Warningf("failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
+		}
+	}(cmd)
 
 	m.WaitGroup.Add(1)
 	go func() {
```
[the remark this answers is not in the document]
chrisThePattyEater: added

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +153,177 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		go func() {
+			klog.Infof("sending SIGTERM to gcsfuse: %v", cmd)
+			if err := cmd.Process.Signal(syscall.SIGTERM); err != nil {
+				klog.Warningf("failed to send SIGTERM to gcsfuse: %v", err)
+			}
+			if err := forceKillCmd(cmd, cmdDone); err != nil {
+				klog.Warningf("failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
+			}
+			klog.Infof("gcsfuse with process id %d has been shut down", cmd.Process.Pid)
+		}()
```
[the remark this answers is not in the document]
chrisThePattyEater: added

pkg/sidecar_mounter/mounter_server.go
```diff
@@ -57,11 +63,11 @@ func (ms *MounterServer) Mount(ctx context.Context, req *mounter.MountRequest) (
...
-		TempDir:          webhook.SidecarContainerTmpVolumeMountPath,
-		ErrWriter:        NewErrorWriter(filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, util.ErrorFileName)),
-		BufferDir:        webhook.SidecarContainerBufferVolumeMountPath,
-		CacheDir:         webhook.SidecarContainerCacheVolumeMountPath,
-		ConfigFile:       filepath.Join(webhook.SidecarContainerTmpVolumeMountPath, "config.yaml"),
+		TempDir:          ms.tmpDir,
+		ErrWriter:        NewErrorWriter(filepath.Join(ms.tmpDir, util.ErrorFileName)),
+		BufferDir:        ms.bufferDir,
+		CacheDir:         ms.cacheDir,
+		ConfigFile:       filepath.Join(ms.tmpDir, "config.yaml"),
```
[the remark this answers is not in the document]
chrisThePattyEater: shared mount has one volume per mounter pod so this is invalid

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +153,183 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-		<-ctx.Done()
+	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
+
+	// TODO(FUECHR) Add bucket access check.
+
+	klog.Infof("Start mounting GCS bucket %q", mc.BucketName)
+
+	if err := createBufferTempDir(mc); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: we already do this in node stage volume 	// Check if the staging path is already mounted.
	mounted, err := s.isDirMounted(stagingPath)

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +153,183 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+// forceKillCmd will force kill the process for a cmd
+// once ctx.Done() is closed.
+func forceKillCmd(cmd *exec.Cmd, cmdDone ...<-chan struct{}) error {
+	if cmd == nil || cmd.Process == nil {
+		return nil
+	}
+	var done <-chan struct{}
+	if len(cmdDone) > 0 {
+		done = cmdDone[0]
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: fixed and removed the redundant var done <-chan struct{}
	if len(cmdDone) > 0 {
		done = cmdDone[0]
	}

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +153,179 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-		<-ctx.Done()
+	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
+
+	// TODO(FUECHR) Add bucket access check.
+
+	klog.Infof("Start mounting GCS bucket %q", mc.BucketName)
+
+	if err := createBufferTempDir(mc); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: added since its true this should throw error for gRPC

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +153,179 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+			}
+			klog.Infof("sending SIGTERM to gcsfuse: %v", cmd)
+			if err := cmd.Process.Signal(syscall.SIGTERM); err != nil {
+				klog.Warningf("failed to send SIGTERM to gcsfuse: %v", err)
 			}
+			if err := forceKillCmd(cmd, cmdDone); err != nil {
+				klog.Warningf("failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
+			}
+			klog.Infof("gcsfuse with process id %d has been shut down", cmd.Process.Pid)
+		}()
```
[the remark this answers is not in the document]
chrisThePattyEater: yeah we can just rely on the context ending no need to manually kill

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +153,171 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+	// This allows the next Mount attempt to start a fresh gcsfuse process.
+	if err := util.WaitForPathMounted(pollCtx, mc.SharedMountPoint); err != nil {
+		klog.Errorf("gcsfuse process did not mount path %q within the context timeout: %v, killing gcsfuse process with id %d", mc.SharedMountPoint, err, cmd.Process.Pid)
+		// Terminate the gcsfuse process asynchronously to avoid blocking the gRPC response.
+		go func() {
+			if err := forceKillCmd(cmd, cmdDone); err != nil {
+				klog.Warningf("failed to force kill gcsfuse process with id %d: %v", cmd.Process.Pid, err)
 			}
+			klog.Infof("gcsfuse with process id %d has been shut down", cmd.Process.Pid)
+		}()
```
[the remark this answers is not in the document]
chrisThePattyEater: this syncs up to the other call for force kill added

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -113,32 +115,36 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-
-	cmd := m.prepareMountCommand(ctx, mc, args)
+	cmd := m.prepareMountCommand(ctx, mc, "/dev/fd/3")
 	cmd.ExtraFiles = []*os.File{os.NewFile(uintptr(mc.FileDescriptor), "/dev/fuse")}
 
-	startCleanupMonitor(ctx, cmd)
+	// cmdDone is closed when waitForGcsfuseExit returns, signaling to forceKillCmd
+	// that the gcsfuse process has exited so it can return immediately without
+	// waiting out the grace period or risking PID reuse.
+	cmdDone := make(chan struct{})
```
uriel-guzman: This looks reasonable but it's also modifying the gke-gcsfuse-sidecar's critical path. We should avoid making any non-trivial logical changes to the gke-gcsfuse-sidecar to minimize the risk of us breaking it. Instead, keep this logic (cmdDone stuff you added) only to the mounter pod code. For gke-gcsfuse-sidecar path, let's just use whatever was already existing here without any new changes (this means, you'd not call forceKillCmd here, only from the mounter pod).

chrisThePattyEater: changed it back to what is was before we pre reformatted

pkg/util/util.go
```diff
@@ -345,3 +348,51 @@ func CheckNotSymlink(path string) error {
...
+			}
+		case <-ctx.Done():
+			code := codes.DeadlineExceeded
+			if errors.Is(ctx.Err(), context.Canceled) {
+				code = codes.Canceled
+			}
+			return status.Errorf(code, "timeout waiting for path %q to be mounted: %v", path, ctx.Err())
+		}
+	}
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: added this

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: reverted it back to the original code

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -56,6 +56,8 @@ import (
 
 const metricEndpointFmt = "http://localhost:%v/metrics"
 
+var forceKillGracePeriod = 5 * time.Second
```
uriel-guzman: This should be called `gracefulKillGracePeriod`, because the period you are waiting for is for gracefull shutdown to finish successfully, not forcefull shutdown.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +150,167 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
-	// when the ctx.Done() is closed,
-	// the main workload containers have exited,
-	// so it is safe to force kill the gcsfuse process.
-	go func(cmd *exec.Cmd) {
-		<-ctx.Done()
+	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
 
-		klog.V(4).Infof("context marked as done, sleep for 5 seconds, to evaluate gcsfuse process %d exit state", cmd.Process.Pid)
-		time.Sleep(time.Second * 5)
+	// TODO(FUECHR) Add bucket access check.
```
uriel-guzman: Remove TODO, we're not implementing bucket access check anymore.

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +150,167 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+
+	select {
+	case <-cmdDone:
+		// gcsfuse exited cleanly or responded to SIGTERM within the grace period.
+		// Return immediately without calling cmd.Process.Kill() to prevent PID reuse issues.
+		return nil
+	case <-time.After(forceKillGracePeriod):
+		// Grace period expired and gcsfuse has not exited; force kill the process.
+		klog.Warningf("after %v, force killing the process with id %v", forceKillGracePeriod, cmd.Process.Pid)
+		if err := cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) && !strings.Contains(err.Error(), "bad file descriptor") && !strings.Contains(err.Error(), "process already finished") {
```
uriel-guzman: In sidecar architecture, we just log a warning if the force kill fails: https://github.com/chrisThePattyEater/gcs-fuse-csi-driver/blob/4370e55a0f3d09746dfc775340fa9fdb1343a404/pkg/sidecar_mounter/sidecar_mounter.go#L137. Why is it different in shared mount?

uriel-guzman: It seems like you're doing some string filtering (which I don't know where it comes from btw, is it a Gemini comment?) and then you just return the error to log a warning anyway. Can't this just be simplified to always log a warning with the error inside `forceKillCmd`, like the sidecar architecture version?

uriel-guzman (changes requested):

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: added

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: added a comment and moved the err to just be a warning in the same way the sidecar does

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -147,65 +150,168 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+		// gcsfuse exited cleanly or responded to SIGTERM within the grace period.
+		// Return immediately without calling cmd.Process.Kill() to prevent PID reuse issues.
+		return nil
+	case <-time.After(forceKillGracePeriod):
+		// Grace period expired and gcsfuse has not exited; force kill the process.
+		// Ignore benign race conditions where gcsfuse exited or was reaped by cmd.Wait() concurrently:
+		// - os.ErrProcessDone / "process already finished": process already exited.
+		// - "bad file descriptor" (EBADF): underlying pidfd was closed by cmd.Wait() reaping the child.
+		klog.Warningf("after %v, force killing the process with id %v", forceKillGracePeriod, cmd.Process.Pid)
+		if err := cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) && !strings.Contains(err.Error(), "bad file descriptor") && !strings.Contains(err.Error(), "process already finished") {
```
uriel-guzman: I am confused. You said: "as for the warning i think you are right we can just do what sidecar does since the callers both just log it as a warning and return nil so doesnt make sense to make it any different", which means that you were going to make cmd.Process.Kill only log a warning, but instead, you just added a comment?

uriel-guzman (changes requested):

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: apparently git had a save conflict? I resolved

chrisThePattyEater: If there is an error we just log it 

		if err := cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) && !strings.Contains(err.Error(), "bad file descriptor") && !strings.Contains(err.Error(), "process already finished") {
			klog.Warningf("failed to force kill process with id %v", cmd.Process.Pid)
		}
	}
}

pkg/sidecar_mounter/sidecar_mounter.go
chrisThePattyEater: fixed this

uriel-guzman (approved):

PR #1540, opened 2026-07-20
Make controller.go obtain mounter pod image from ConfigMap

uriel-guzman:
**What type of PR is this?**
This PR modifies controller.go so that it retreives the mounter pod image from the existing ConfigMap which is used by the webhook to obtain the sidecar mounter image. The PR introduces a new ConfigMap informer that only tracks for ConfigMaps in the driver's namespace and trims irrelevant data, making it resource efficient. The code also renames `leader-election-namespace` to a more generic `driver-namespace`. The rename is safe, since the controller.yaml downstream will be modified to `driver-namespace` at the same time the shared mount feature is released.

/kind feature

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/cloud_provider/clientset/clientset.go
+++ b/pkg/cloud_provider/clientset/clientset.go
@@ -55,6 +55,7 @@ type Interface interface {
 	ConfigurePVCLister(ctx context.Context)
 	ConfigureSCLister(ctx context.Context)
 	ConfigurePodTemplateLister(ctx context.Context)
+	ConfigureConfigMapLister(ctx context.Context, namespace string)
 	GetPod(namespace, name string) (*corev1.Pod, error)
 	GetMounterPod(namespace, name string) (*corev1.Pod, error)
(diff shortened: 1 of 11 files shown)


pkg/cloud_provider/clientset/clientset.go
```diff
@@ -358,6 +361,35 @@ func (c *Clientset) ConfigurePodTemplateLister(ctx context.Context) {
...
+		return &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cmObj.ObjectMeta.Name,
+				Namespace: cmObj.ObjectMeta.Namespace,
+			},
+			Data: cmObj.Data,
+		}, nil
+	}
+
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
```
yaozile123, line 379: Should we only watch one config map `gcsfusecsi-image-config`?

yaozile123 (approved):

PR #1550, opened 2026-07-21
Implement Machine Type Defaulting for Shared Node Mount

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR extends machine type defaulting to the shared node mount mode.

- `PutFlagsFromDriverToTargetPath` is renamed to `writeDriverFlagsFile` and made unexported. The old name was misleading (it operates on an emptyDir base path, not a target path), and the function is only ever called from the `populateDriverFlagsForDefaulting` helper, so making it unexported is safe. The `fileName` parameter is also removed from the method since the file name is always `FlagFileForDefaultingPath`.
- Add new method`ReadDriverFlagsForDefaulting` in `pkg/sidecar_mounter` so the same logic is shared between the sidecar main loop and the mounter server's `Mount` handler.
- Fix a bug in the defaulting flags E2E test suite (`test/e2e/testsuites/mount.go`) where the test case intended for `--disable-autoconfig=false` mistakenly passed `specs.DisableAutoconfig`, causing both E2E test cases to test `--disable-autoconfig=true`.

Fixes #

I have tested manually and verified the machine type defaulting works on shared mount and sidecar. I also verified the e2e tests passed after the change:
```
E2E_TEST_BUILD_DRIVER=true \
REGISTRY=${REGISTRY} \
STAGINGVERSION=${STAGINGVERSION} \
E2E_TEST_FOCUS="disable-autoconfig" \
make e2e-test

Ran 4 of 671 Specs in 13.539 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 667 Skipped


Ginkgo ran 1 suite in 29.087633152s
Test Suite Passed
```

```release-note

```

### DIFF
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -110,16 +110,11 @@ func main() {
 	if err != nil {
 		klog.Fatalf("failed to look up socket paths: %v", err)
 	}
-	flagsFromDriver := map[string]string{}
-	defaultingFlagFilePath := *volumeBasePath + "/" + driver.FlagFileForDefaultingPath
+	defaultingFlagFilePath := filepath.Join(*volumeBasePath, driver.FlagFileForDefaultingPath)
 	klog.Infof("Checking if defaulting-flag file %q exists", defaultingFlagFilePath)
-	if _, err := os.Stat(defaultingFlagFilePath); err == nil {
-		machineTypeBytes, err := os.ReadFile(defaultingFlagFilePath)
-		if err != nil {
-			klog.Fatalf("failed to read defaulting-flag file: %v", err)
-		}
-		fileContent := string(machineTypeBytes)
-		flagsFromDriver = driver.ParseFlagMapFromFlagFile(fileContent)
+	flagsFromDriver, err := sidecarmounter.ReadDriverFlagsForDefaulting(defaultingFlagFilePath)
+	if err != nil {
+		klog.Fatalf("failed to read defaulting-flag file: %v", err)
 	}
 	var startProfilerOnce sync.Once
 	for _, sp := range socketPaths {
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -215,6 +215,28 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 
 	return &csi.NodePublishVolumeResponse{}, nil
 }
+func (s *nodeServer) populateDriverFlagsForDefaulting(node *corev1.Node, mounterImage string, emptyDirBasePath string) error {
+	if node == nil {
+		return fmt.Errorf("node cannot be nil")
+	}
+	if !s.driver.isSidecarVersionSupportedForGivenFeature(mounterImage, MachineTypeAutoConfigSidecarMinVersion) {
+		return nil
+	}
+
+	shouldDisableAutoConfig := s.driver.config.DisableAutoconfig
+	machineType, ok := node.Labels[clientset.MachineTypeKey]
+	if !ok {
+		klog.Warningf("Unable to fetch target node %v's machine type", node.Name)
+		return nil
+	}
+
+	flagMap := map[string]string{
+		"machine-type":       machineType,
+		"disable-autoconfig": strconv.FormatBool(shouldDisableAutoConfig),
+	}
+
+	return writeDriverFlagsFile(flagMap, emptyDirBasePath)
+}
 
 func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
 	// Rate limit NodePublishVolume calls to avoid kube API throttling.
@@ -376,27 +398,16 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 	}
 
-	// Only pass mountOptions flags for defaulting if sidecar container is managed and satisifies min version requirement
-	if s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, MachineTypeAutoConfigSidecarMinVersion) {
-		shouldDisableAutoConfig := s.driver.config.DisableAutoconfig
-		machineType, ok := node.Labels[clientset.MachineTypeKey]
-		if ok {
-			flagMap := map[string]string{"machine-type": machineType, "disable-autoconfig": strconv.FormatBool(shouldDisableAutoConfig)}
-			if err := PutFlagsFromDriverToTargetPath(flagMap, targetPath, FlagFileForDefaultingPath); err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
-		} else {
-			klog.Warningf("Unable to fetch target node %v's machine type", node.Name)
-		}
+	emptyDirBasePath, err := util.PrepareEmptyDir(targetPath, true)
+	if err != nil {
+		klog.Warningf("Failed to prepare empty dir from target path %q: %v", targetPath, err)
 	}
 
-	var podUID, volumeName, emptyDirBasePath string
+	var podUID, volumeName string
 	if s.isGcsFuseKernelParamsFeatureSupported(gcsFuseSidecarImage, vs) {
 		var err error
 		if podUID, volumeName, err = util.ParsePodIDVolumeFromTargetpath(targetPath); err != nil {
 			klog.Warningf("Failed to parse pod ID and volume name from target path %q for kernel params monitor: %v. Monitor will not be started.", targetPath, err)
-		} else if emptyDirBasePath, err = util.PrepareEmptyDir(targetPath, false); err != nil {
-			klog.Warningf("Failed to prepare empty dir from target path %q for kernel params monitor: %v. Monitor will not be started.", targetPath, err)
 		}
 	}
 
@@ -439,6 +450,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		klog.V(4).Infof("NodePublishVolume succeeded on volume %q to target path %q, mount already exists.", args.bucketName, targetPath)
 
 		return &csi.NodePublishVolumeResponse{}, nil
+	}
+
+	// Only pass mountOptions flags for defaulting if sidecar container is managed and satisfies min version requirement
+	if emptyDirBasePath != "" {
+		if err := s.populateDriverFlagsForDefaulting(node, gcsFuseSidecarImage, filepath.Dir(emptyDirBasePath)); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 	}
 
 	// Unlike other features, we'll assume multi NIC can be used unless we know for certain we have a version mismatch.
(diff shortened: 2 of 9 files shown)


pkg/csi_driver/node.go
```diff
@@ -215,6 +215,25 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 
 	return &csi.NodePublishVolumeResponse{}, nil
 }
+func (s *nodeServer) populateDriverFlagsForDefaulting(node *corev1.Node, mounterImage string, emptyDirBasePath string) error {
+	if !s.driver.isSidecarVersionSupportedForGivenFeature(mounterImage, MachineTypeAutoConfigSidecarMinVersion) {
+		return nil
+	}
```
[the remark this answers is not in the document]
yaozile123: Done

cmd/sidecar_mounter/main.go
```diff
@@ -110,17 +110,9 @@ func main() {
 	if err != nil {
 		klog.Fatalf("failed to look up socket paths: %v", err)
 	}
-	flagsFromDriver := map[string]string{}
 	defaultingFlagFilePath := *volumeBasePath + "/" + driver.FlagFileForDefaultingPath
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -203,6 +204,18 @@ func mergeFlags(mountConfigFlagMap map[string]string, driverFlagMap map[string]s
...
+	if _, err := os.Stat(defaultFilePath); err != nil {
+		return map[string]string{}
+	}
+	machineTypeBytes, err := os.ReadFile(defaultFilePath)
+	if err != nil {
+		klog.Warningf("failed to read defaulting-flag file %q: %v", defaultFilePath, err)
+		return map[string]string{}
+	}
+	return driver.ParseFlagMapFromFlagFile(string(machineTypeBytes))
+}
```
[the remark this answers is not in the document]

pkg/csi_driver/node.go
```diff
@@ -376,18 +395,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
-			klog.Warningf("Unable to fetch target node %v's machine type", node.Name)
-		}
+	// Only pass mountOptions flags for defaulting if sidecar container is managed and satisfies min version requirement
+	sidecarEmptyDir, err := util.PrepareEmptyDir(targetPath, true)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get emptyDir path: %v", err)
+	}
+	if err := s.populateDriverFlagsForDefaulting(node, gcsFuseSidecarImage, filepath.Dir(sidecarEmptyDir)); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
 	}
```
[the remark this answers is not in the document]

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -203,6 +204,17 @@ func mergeFlags(mountConfigFlagMap map[string]string, driverFlagMap map[string]s
...
+func ReadDriverFlagsForDefaulting(defaultFilePath string) map[string]string {
+	machineTypeBytes, err := os.ReadFile(defaultFilePath)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			klog.Warningf("failed to read defaulting-flag file %q: %v", defaultFilePath, err)
+		}
+		return map[string]string{}
+	}
+	return driver.ParseFlagMapFromFlagFile(string(machineTypeBytes))
+}
```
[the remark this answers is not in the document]
yaozile123: I have replaced `strings.Split()` with `strings.Cut()` in `ParseFlagMapFromFlagFile()`

pkg/csi_driver/utils.go
```diff
@@ -680,27 +680,24 @@ func (d *GCSDriver) isSidecarVersionSupportedForGivenFeature(imageName string, s
...
+func writeDriverFlagsFile(flagMap map[string]string, emptyDirBasePath string) error {
+	if err := os.MkdirAll(emptyDirBasePath, 0750); err != nil {
+		return fmt.Errorf("failed to create directory %q: %w", emptyDirBasePath, err)
 	}
 
-	absolutePath := filepath.Dir(emptyDirBasePath) + "/" + fileName
+	absolutePath := filepath.Join(emptyDirBasePath, FlagFileForDefaultingPath)
 	klog.V(4).Infof("Writing flags needed for gcsfuse defaulting logic to file %q: %v", absolutePath, flagMap)
 
-	parentDir := filepath.Dir(emptyDirBasePath)
```
uriel-guzman, line 692: Why are we removing parentDir (`filepath.Dir(emptyDirBasePath)`) and instead directly using `emptyDirBasePath` in `unix.Open(emptyDirBasePath, unix.O_RDONLY|unix.O_DIRECTORY|unix.O_NOFOLLOW, 0)`? Doesn't `filepath.Dir` give you the parent of the dir? e.g. if `emptyDirBasePath` is `/var/lib/kubelet/pods/<pod-uuid>/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp/.volumes/<volume-name>`, `parentDir` would be `var/lib/kubelet/pods/<pod-uuid>/volumes/kubernetes.io~empty-dir/gke-gcsfuse-tmp/.volumes/`?

pkg/csi_driver/utils.go
```diff
@@ -680,27 +680,24 @@ func (d *GCSDriver) isSidecarVersionSupportedForGivenFeature(imageName string, s
...
 	// attacks during subsequent relative operations (using unix.Openat).
-	parentFd, err := unix.Open(parentDir, unix.O_RDONLY|unix.O_DIRECTORY|unix.O_NOFOLLOW, 0)
+	parentFd, err := unix.Open(emptyDirBasePath, unix.O_RDONLY|unix.O_DIRECTORY|unix.O_NOFOLLOW, 0)
 	if err != nil {
-		return fmt.Errorf("failed to open parent directory %q: %w", parentDir, err)
+		return fmt.Errorf("failed to open parent directory %q: %w", emptyDirBasePath, err)
 	}
 	defer unix.Close(parentFd)
 
-	fd, err := unix.Openat(parentFd, filepath.Base(fileName), unix.O_RDWR|unix.O_CREAT|unix.O_TRUNC|unix.O_NOFOLLOW, 0644)
```
uriel-guzman, line 703: Why are we changing `filepath.Base(fileName)` to just `FlagFileForDefaultingPath` (without the parent function?)

pkg/csi_driver/utils.go
```diff
@@ -730,12 +727,10 @@ func ParseFlagMapFromFlagFile(flagFileContent string) map[string]string {
 	configFlags := make(map[string]string)
 	lines := strings.Split(flagFileContent, "\n")
 	for _, line := range lines {
-		if line == "" { // Skip empty lines
```
uriel-guzman, line 733: Is there any particular reason we needed to change the logic in this PR? If not, let's just keep it as it was.

cmd/sidecar_mounter/main.go
```diff
@@ -110,17 +110,9 @@ func main() {
...
 		klog.Fatalf("failed to look up socket paths: %v", err)
 	}
-	flagsFromDriver := map[string]string{}
-	defaultingFlagFilePath := *volumeBasePath + "/" + driver.FlagFileForDefaultingPath
+	defaultingFlagFilePath := filepath.Join(*volumeBasePath, driver.FlagFileForDefaultingPath)
 	klog.Infof("Checking if defaulting-flag file %q exists", defaultingFlagFilePath)
-	if _, err := os.Stat(defaultingFlagFilePath); err == nil {
-		machineTypeBytes, err := os.ReadFile(defaultingFlagFilePath)
-		if err != nil {
-			klog.Fatalf("failed to read defaulting-flag file: %v", err)
```
uriel-guzman, line 119: This used to be fatal if we couldn't read the defaulting flags file. Now in `ReadDriverFlagsForDefaulting`, we're just logging a warning and returning an empty map. Since this code path is used by the gke-gcsfuse-sidecar, we should continue to preserve the original behavior of it being fatal.

We should avoid modifying existing behavior on non-shared mount paths.

uriel-guzman (changes requested):

pkg/csi_driver/utils.go
yaozile123: Becuase Gemini leave a [critical comment](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1550#discussion_r3626368341) and I think it's necessary to do the change, should we roll it back?

pkg/csi_driver/utils.go
yaozile123: Previously, fileName was always set to FlagFileForDefaultingPath and this method had only [one caller](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/0f5b7f24f9c32e23d53542d3e7857ad51774448a/pkg/csi_driver/node.go#L385). Since FlagFileForDefaultingPath is already a constant filename ("flags-for-defaulting"), calling filepath.Base is no longer needed.

pkg/csi_driver/utils.go
yaozile123: The caller in `node.go` already handles preparing the target parent directory before calling writeDriverFlagsFile. 

In sidecar mode, `filepath.Dir(sidecarEmptyDir)` is passed to strip the <volume-name> subfolder, while in shared mount mode, emptyDirBasePath (which does not contain .volumes) is passed directly. So emptyDirBasePath inside writeDriverFlagsFile should already be the parent directory.

cmd/sidecar_mounter/main.go
yaozile123: Done, I fixed it by adding `ReadDriverFlagsForDefaulting` to return an error.

pkg/csi_driver/node.go
```diff
@@ -441,6 +449,15 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	// Only pass mountOptions flags for defaulting if sidecar container is managed and satisfies min version requirement
+	sidecarEmptyDir, err := util.PrepareEmptyDir(targetPath, true)
```
uriel-guzman: `util.PrepareEmptyDir(targetPath, false)` is being called in NodePublish here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L398. Can we just keep one? Also, no need to call this "sidecarEmptyDir". We already know that this code path is for sidecar only, the specific path for shared mount is called NodePublishVolumeForSharedMount: https://github.com/yaozile123/gcs-fuse-csi-driver/blob/3f22c8fb324d4c2b93d72b7b27b4a23c57d3297b/pkg/csi_driver/node.go#L108, so we don't need to specify the differences between the sidecar and shared mount architecture in the sidecar-only path.

pkg/csi_driver/utils.go
uriel-guzman: Ah, I see. So the parent was already passed in:
```
if err := s.populateDriverFlagsForDefaulting(node, gcsFuseSidecarImage, filepath.Dir(sidecarEmptyDir)); err != nil {
		return nil, status.Error(codes.Internal, err.Error())
	}
```
I hadn't seen that, sgtm then.

uriel-guzman (changes requested):

uriel-guzman (approved):

PR #1558, opened 2026-07-22
Implement metrics collector register for shared mount.

uriel-guzman:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This PR handles registration of metrics collector during NodePublishVolumeWithSharedMount, and deregistration during NodeUnstageVolume. The PR also refactors `RegisterMetricsCollector` so it accepts a generic mount path, and allows passing `podUID`, `emptyDirBasePath`, and `volumeName` as parameters, breaking the dependency on the targetPath.

The registration is added during NodePublishVolume to ensure that if the driver restarts while the volume is already mounted, the next time that node re-publish is called, it gets registered again. 

The unregistration is performed during NodeStageVolume to ensure that all of the dependent workloads have already been unmounted and no longer require the metrics collector.

The same rationale is applicable to the GCSFuse Kernel Params Monitoring feature, so it is also moved with the same pattern.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -136,6 +136,10 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 	if len(mounterPodName) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Mounter Pod Name must be provided")
 	}
+	args, err := parseRequestArguments(req.GetVolumeId(), req.GetReadonly(), req.GetVolumeCapability(), req.GetVolumeContext())
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
 
 	// Acquire a lock on the target path instead of volumeID, since we do not want to serialize multiple node publish calls on the same volume.
 	if acquired := s.volumeLocks.TryAcquire(targetPath); !acquired {
@@ -169,6 +173,19 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Error(code, err.Error())
 	}
 
+	// We ignore the pod UID parsed from the target path, since we register the mounter pod's UID instead.
+	_, volumeName, err := util.ParsePodIDVolumeFromTargetpath(targetPath)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to parse volume name from target path %q: %v", targetPath, err)
+	}
+	mounterPodUID := string(mounterPod.UID)
+
+	// Start monitoring goroutine for new mounts.
+	mounterPodImage, err := mounterPodImage(mounterPod)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get image of mounter pod %s/%s: %v", mounterPodNamespace, mounterPodName, err)
+	}
+
 	// NodePublish is guaranteed to be called after NodeStage has waited for mounter pod to become running.
 	// Mounter pod not running at this stage means that the pod started and failed (e.g. killed by Kubelet
 	// during OOMKILL).
@@ -185,6 +202,30 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Errorf(codes.Internal, "staging path %s was not found mounted", stagingPath)
 	}
 
+	// Use staging path as the volume identifier because NodePublishVolumeForSharedMount is only
+	// called when the shared mount feature is being used. VolumeState stores the shared state of
+	// all bind-mounted target paths from the same staging path, and is used to idempotently start
+	// and re-start processes such as GCSFuse Kernel Params monitoring and metrics registration, which need to
+	// be re-invoked when the driver restarts, by leveraging the node re-publish capability.
+	startSharedMountProcesses := func() {
+		vs, _ := s.volumeStateStore.LoadOrStore(stagingPath, &util.VolumeState{
+			// Store the mounterPodUID and the volumeName, so that NodeUnstageVolume can call UnregisterMetricsCollector, since
+			// these fields aren't available during node unstaging. If the driver restarts, the VolumeState will no longer
+			// be in memory, so these fields will be lost, but the metrics collector will also not be running anymore, so we don't
+			// need to unregister it.
+			MounterPodUID: mounterPodUID,
+			VolumeName:    volumeName,
+		})
+
+		s.startGcsFuseKernelParamsMonitoring(stagingPath, emptyDirPath, mounterPodUID, volumeName, mounterPodImage, args.bucketName, vs)
+
+		if s.driver.config.MetricsManager != nil && !args.disableMetricsCollection {
+			klog.V(4).Infof("NodePublishVolume enabling metrics collector for staging path %q", stagingPath)
+			s.driver.config.MetricsManager.RegisterMetricsCollector(stagingPath, mounterPodNamespace, mounterPodName, args.bucketName, s.driver.config.NodeID, emptyDirPath, mounterPodUID, volumeName)
+		}
+		return
+	}
+
 	// Succeed if the target path was already mounted.
 	targetPathMounted, err := s.isDirMounted(targetPath)
 	if err != nil {
@@ -192,6 +233,7 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 	}
 	if targetPathMounted {
 		klog.Infof("NodePublishVolume succeeded on staging path %q to target path %q, mount already exists.", stagingPath, targetPath)
+		startSharedMountProcesses()
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
@@ -213,8 +255,10 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Errorf(codes.Internal, "failed to bind mount staging path %q to target path %q: %v", stagingPath, targetPath, err)
 	}
 
+	startSharedMountProcesses()
 	return &csi.NodePublishVolumeResponse{}, nil
 }
+
 func (s *nodeServer) populateDriverFlagsForDefaulting(node *corev1.Node, mounterImage string, emptyDirBasePath string) error {
 	if node == nil {
 		return fmt.Errorf("node cannot be nil")
@@ -376,6 +420,14 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Register metrics collector.
 	// It is idempotent to register the same collector in node republish calls.
+	podUID, volumeName, err := util.ParsePodIDVolumeFromTargetpath(targetPath)
+	if err != nil {
+		klog.Warningf("Failed to parse pod ID and volume name from target path %q: %v.", targetPath, err)
+	}
+	emptyDirBasePath, err := util.PrepareEmptyDir(targetPath, true)
+	if err != nil {
+		klog.Warningf("Failed to prepare empty dir from target path %q: %v.", targetPath, err)
+	}
 	if s.driver.config.MetricsManager != nil && !args.disableMetricsCollection {
 		gcsFuseVolumeCount, err := s.countGcsFuseVolumes(pod)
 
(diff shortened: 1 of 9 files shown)


pkg/csi_mounter/csi_mounter.go
```diff
@@ -278,7 +278,11 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
 	// Need to create symbolic link of emptyDirBasePath to socketBasePath,
 	// because the socket absolute path is longer than 104 characters,
 	// which will cause "bind: invalid argument" errors.
-	socketBasePath := util.GetSocketBasePath(target, m.fuseSocketDir)
+	podUID, volumeName, err := util.ParsePodIDVolumeFromTargetpath(target)
+	if err != nil {
+		klog.Warningf("Failed to parse pod ID and volume name from target path %q: %v.", target, err)
+	}
+	socketBasePath := util.GetSocketBasePath(podUID, volumeName, m.fuseSocketDir)
```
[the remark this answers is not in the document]
uriel-guzman: This is existing behavior, so we should just keep it as-is.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -322,7 +326,11 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
 }
 
 func (m *Mounter) cleanupSocket(target string) {
-	socketBasePath := util.GetSocketBasePath(target, m.fuseSocketDir)
+	podUID, volumeName, err := util.ParsePodIDVolumeFromTargetpath(target)
+	if err != nil {
+		klog.Warningf("Failed to parse pod ID and volume name from target path %q: %v.", target, err)
+	}
```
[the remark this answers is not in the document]
uriel-guzman: This will be fixed in a separate PR.

pkg/csi_mounter/csi_mounter.go
```diff
@@ -256,7 +256,11 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
 	// Need to create symbolic link of emptyDirBasePath to socketBasePath,
 	// because the socket absolute path is longer than 104 characters,
 	// which will cause "bind: invalid argument" errors.
-	socketBasePath := util.GetSocketBasePath(target, m.fuseSocketDir)
+	podUID, volumeName, err := util.ParsePodIDVolumeFromTargetpath(target)
+	if err != nil {
+		klog.Warningf("Failed to parse pod ID and volume name from target path %q: %v.", target, err)
+	}
```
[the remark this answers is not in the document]
uriel-guzman: This is existing behavior, so we should leave it as is.

pkg/csi_driver/node.go
```diff
@@ -1165,6 +1193,25 @@ func (s *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstage
 	}
 	defer s.volumeLocks.Release(stagingPath)
 
+	// If the driver restarted after NodeStageVolume and before NodeUnstageVolume, and before the next node re-publish,
+	// VolumeStateStore will be nil, and UnregisterMetricsCollector and vs.GCSFuseKernelMonitorState.CancelFunc
+	// will not be invoked. This is acceptable, since this implies that the metrics collector process and the
+	// GCSFuse Kernel Params monitoring go routine are no longer running anymore, meaning they don't need to be stopped.
+	if vs, ok := s.volumeStateStore.Load(stagingPath); ok && vs != nil {
+		// Unregister metrics collecter.
```
yaozile123: `collecter` -> `collector`

pkg/metrics/metrics.go
```diff
@@ -178,10 +169,10 @@ func (mm *manager) UnregisterMetricsCollector(targetPath, nodeName string) {
 	defer mm.mutex.Unlock()
 
 	if ok := mm.registry.Unregister(c); !ok {
-		klog.Infof("Unregister metrics collector for target path %q is not needed since the collector is not registered.", targetPath)
+		klog.Infof("Unregister metrics collector for mount path %q is not needed since the collector is not registered.", mountPath)
 	} else {
-		mm.volumePublishPathRegistered.Delete(targetPath)
-		klog.Infof("successfully unregistered a metrics collector for target path %q. There's %d collectors registered on node %q.", targetPath, mm.volumePublishPathRegistered.Len(), nodeName)
+		mm.volumeMountPathRegistered.Delete(mountPath)
+		klog.Infof("successfully unregistered a metrics collector for target path %q. There's %d collectors registered on node %q.", mountPath, mm.volumeMountPathRegistered.Len(), nodeName)
```
yaozile123: Should this log be "successfully unregistered a metrics collector for mount path" instead of target path?

pkg/csi_driver/node.go
```diff
@@ -136,13 +136,28 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
...
 		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, targetPath)
 	}
 	defer s.volumeLocks.Release(targetPath)
 
+	// Use staging path as the volume identifier because NodePublishVolumeForSharedMount is only
+	// called when the shared mount feature is being used. VolumeState stores the shared state of
+	// all bind-mounted target paths from the same staging path, and is used to idempotently start
+	// and re-start processes such as GCSFuse Kernel Params monitoring and metrics registration, which need to
+	// be re-invoked when the driver restarts, by leveraging the node re-publish capability.
+	vs, ok := s.volumeStateStore.Load(stagingPath)
```
yaozile123: In [NPV](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6fb9447c161be5ccab14d9fea1c91ec7fd9137f5/pkg/csi_driver/node.go#L302-L317) we acquired the lock and do `volumeStateStore.Store` both on `targetPath`. But in NPV for shared mount we acquired the lock on `targetPath` and do `volumeStateStore.Store` on `stagingPath`. How could we handle the race condition for multiple `volumeStateStore.Store` on the same`stagingPath`?

pkg/csi_driver/node.go
```diff
@@ -169,6 +184,38 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
...
+		return nil, status.Errorf(codes.Internal, "failed to parse volume name from target path %q: %v", targetPath, err)
+	}
+	mounterPodUID := string(mounterPod.UID)
+
+	// Start monitoring goroutine for new mounts.
+	mounterPodImage, err := mounterPodImage(mounterPod)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get image of mounter pod %s/%s: %v", mounterPodNamespace, mounterPodName, err)
+	}
+	s.startGcsFuseKernelParamsMonitoring(stagingPath, emptyDirPath, mounterPodUID, volumeName, mounterPodImage, vs)
```
yaozile123: In NPV, we call `s.startGcsFuseKernelParamsMonitoring` after [checked the targetPath is mounted](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6fb9447c161be5ccab14d9fea1c91ec7fd9137f5/pkg/csi_driver/node.go#L401-L435). Should we move it after confirming `mounted == true` and pod health checks pass to match the behavior in NPV?

pkg/csi_driver/node.go
uriel-guzman: Thanks, you're right. There's no locking mechanism for stagingPath under NPV. I added a function to lock and unlock the staging path while the VolumeState is using it. I opted to make a function instead of using defer because I don't want to lock the stagingPath for the entirety of the NPV duration, since that could make NPV slower at higher scales, since it would serialize instead of parallelize concurrent NPV calls from the same stagingPath.

pkg/csi_driver/node.go
uriel-guzman: It's actually called twice in existing NPV for sidecar architecture.

1. It's called here when NPV is about to succeed: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6fb9447c161be5ccab14d9fea1c91ec7fd9137f5/pkg/csi_driver/node.go#L490
2. And also called where you mentioned, which is when `mounted == true`. 

Which means that it's always called for each successful NodePublishVolume response. I think you're right though, it doesn't makes ense to start the monitoring when we know that it can fail. Same goes for the metrics collector registration. I have refactored this into a helper function so that it only starts both when we are sure they will succeed.

pkg/csi_driver/node.go
```diff
@@ -185,13 +202,69 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
...
+	//
+	// Locking on stagingPath here is critical: while s.volumeLocks.TryAcquire(targetPath)
+	// serializes publications for the same pod target, multiple pods can publish concurrent
+	// bind-mounts targeting the SAME shared stagingPath. This stagingPath lock creates a
+	// localized critical section to safely read, modify, and store state in s.volumeStateStore
+	// without serializing the heavy OS bind-mount or API calls.
+	startSharedMountProcesses := func() error {
+		// Acquire a lock on the staging path to prevent race conditions when initializing the shared
+		// volume state and starting the kernel param monitor across concurrent target path publications.
+		if acquired := s.volumeLocks.TryAcquire(stagingPath); !acquired {
```
yaozile123: If we returning `codes.Aborted` here after the bind-mount already succeeded, it might degrades pod startup time due to kubelet's exponential backoff retries.

I think we could implement an atomic `LoadOrStore` method on `VolumeStateStore` which wrapping `sync.Map.LoadOrStore`:

```go
func (vss *VolumeStateStore) LoadOrStore(volumeID string, state *VolumeState) (*VolumeState, bool) {
	actual, loaded := vss.store.LoadOrStore(volumeID, state)
	if !loaded {
		atomic.AddInt32(&vss.size, 1)
	}
	return actual.(*VolumeState), loaded
}
```

And this would allows concurrent NPV on the same stagingPath to atomically share the VolumeState without aborts or lock contention:
```go
vs, _ := s.volumeStateStore.LoadOrStore(stagingPath, &util.VolumeState{
			MounterPodUID: mounterPodUID,
			VolumeName:    volumeName,
		})
```

chrisThePattyEater: +1 this is what the one pager mentions as the solution for the race condition https://docs.google.com/document/d/1RW3tyswe0PWAKk6mN8K2zWU94Ln-xNB87RA5qTrduJs/edit?tab=t.0#bookmark=id.9rwuw0y0hrwn

pkg/csi_driver/node.go
```diff
@@ -185,13 +202,69 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Errorf(codes.Internal, "staging path %s was not found mounted", stagingPath)
 	}
 
+	// startSharedMountProcesses initializes the shared staging-path state, starts
+	// kernel parameter monitoring, and registers metrics collection under a short-lived
+	// staging-path lock.
+	//
+	// This helper MUST be called only after the target bind-mount has succeeded to
```
chrisThePattyEater: Nit: I don't think this is accurate. It shouldn't matter if bind mount succeeds or fails since NSV will clean it up anyways. Given that this is a shared node mount feature there will likely be other pods using the monitoring and metrics so if we have the lock on the staging path I think we can just start the processes and if needed NSV will clean up.

chrisThePattyEater: Its not a big deal I just think its misleading to say it MUST be called after bind-mount

uriel-guzman: Clean what up? I didn't really understand your comment.

pkg/csi_driver/node.go
uriel-guzman: Thanks, nice suggestion. I have added the improvement.

pkg/metrics/metrics.go
```diff
@@ -144,32 +137,30 @@ func (mm *manager) RegisterMetricsCollector(targetPath, podNamespace, podName, b
...
+		mm.volumeMountPathRegistered.Insert(mountPath)
+		klog.Infof("successfully registered a new metrics collector for mount path %q. There's %d collectors registered on node %q.", mountPath, mm.volumeMountPathRegistered.Len(), nodeName)
 	}
 }
 
 // UnregisterMetricsCollector unregisters the metrics collector. It is idempotent to unregister the same collector.
-func (mm *manager) UnregisterMetricsCollector(targetPath, nodeName string) {
-	podUID, volumeName, _ := util.ParsePodIDVolumeFromTargetpath(targetPath)
-
+func (mm *manager) UnregisterMetricsCollector(mountPath, nodeName, podUID, volumeName string) {
```
yaozile123, line 163: For shared mounts, the symlink is created using `mounterPodUID` and `volumeName`. However `UnregisterMetricsCollector` does not remove the symlink and later `Unmount(stagingPath)` cannot reconstruct shared mount socket path from stagingPath. For regular sidecar, the socket symlink is cleaned up by `Unmount(targetPath)` because the pod uid and volume name can be parsed from targetPath. Should we explicitly clean up the symlink for shared mount?

pkg/util/util.go
```diff
@@ -215,16 +214,14 @@ func sha1Hash(str string) string {
 
 // GetSocketBasePath constructs the base path for a Unix domain socket.
 // It takes the target path and the directory where Fuse sockets are stored as input.
```
yaozile123: nit: This comment still mentions target path but the function now takes podUID and volumeName. Can we update it accordingly?

uriel-guzman: done

pkg/metrics/metrics.go
uriel-guzman: This is a pre-existing bug. In a subsequent PR, we need to update NodeUnstageVolume to properly clean up the socket, and to remove the target path dependency in Unmount.

yaozile123 (approved):

PR #1559, opened 2026-07-22
[Shared Mount] Implement Cloud Profiler

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This pr enabled cloud profiler for mounter pod by adding the following
    - Passes pod uid and pod name to the gcsfuse mount args to allow gcsfuse to use cloud profiler.
    - Enabled cloud profiler on the sidecar during mount operation (part of NSV)

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/sidecar_mounter/main.go
+++ b/cmd/sidecar_mounter/main.go
@@ -26,14 +26,11 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
-	"sync"
 	"syscall"
 	"time"
 
-	"cloud.google.com/go/profiler"
 	driver "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/csi_driver"
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
-	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
 	"k8s.io/klog/v2"
 )
@@ -116,7 +113,6 @@ func main() {
 	if err != nil {
 		klog.Fatalf("failed to read defaulting-flag file: %v", err)
 	}
-	var startProfilerOnce sync.Once
 	for _, sp := range socketPaths {
 		klog.V(4).Infof("in sidecar mounter, found socket path %s", sp)
 		// sleep 1.5 seconds before launch the next gcsfuse to avoid
@@ -127,23 +123,7 @@ func main() {
 		if mc != nil {
 			mc.EnsureErrWriter()
 			if mc.EnableCloudProfilerForSidecar {
-				serviceVersion := util.GetCloudProfilerServiceVersion(mc.PodName, mc.PodUID)
-				if serviceVersion == "" {
-					klog.Warning("Cloud Profiler is disabled because both PodName and PodUID are empty.")
-				} else {
-					// Use sync.Once to ensure Cloud Profiler only start once per process.
-					startProfilerOnce.Do(func() {
-						cfg := profiler.Config{
-							Service:        "gke-gcsfuse-sidecar",
-							ServiceVersion: serviceVersion,
-						}
-						if err := profiler.Start(cfg); err != nil {
-							klog.Errorf("Errored while starting cloud profiler, got %v", err)
-						} else {
-							klog.Infof("Running cloud profiler on %s with version %s", cfg.Service, cfg.ServiceVersion)
-						}
-					})
-				}
+				sidecarmounter.StartCloudProfiler(webhook.GcsFuseSidecarName, mc.PodName, mc.PodUID)
 			}
 			if mc.EnableSidecarBucketAccessCheck {
 				mc.SidecarRetryConfig.Cap = *storageServiceAndBucketAccessCap
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -338,15 +338,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	s.populateTokenAndBucketAccessCheckOptions(pod, gcsFuseSidecarImage, vc, &args, vc[VolumeContextKeyPodNamespace], vc[VolumeContextKeyServiceAccountName])
-
-	if args.enableCloudProfilerForSidecar && s.driver.isSidecarVersionSupportedForGivenFeature(gcsFuseSidecarImage, SidecarCloudProfilerMinVersion) {
-		args.fuseMountOptions = joinMountOptions(args.fuseMountOptions, []string{
-			util.EnableCloudProfilerForSidecarConst + "=" + strconv.FormatBool(args.enableCloudProfilerForSidecar),
-			util.PodNameConst + "=" + vc[VolumeContextKeyPodName],
-			util.PodUIDConst + "=" + string(pod.UID),
-		})
-	}
-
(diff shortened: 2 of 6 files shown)


pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -472,3 +477,25 @@ func (mc *MountConfig) prepareConfigFile() error {
...
+
+// StartCloudProfiler starts the Google Cloud Profiler.
+// It uses sync.Once to ensure the profiler is started at most once per process.
+func StartCloudProfiler(serviceName string, podName string, podUID string) {
+	serviceVersion := util.GetCloudProfilerServiceVersion(podName, podUID)
+
+	if serviceVersion == "" {
+		klog.Warning("Cloud Profiler is disabled because both PodName and PodUID are empty.")
+		return
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: i don't think this adds anything, callers are only the sidecar and driver and we are properly setting them up. unless we want to block on podName == "" || podUID == "" i don't think this is worth adding

pkg/csi_driver/node.go
```diff
@@ -773,6 +765,27 @@ func (s *nodeServer) populateTokenAndBucketAccessCheckOptions(pod *corev1.Pod, i
...
 }
 
+// appendCloudProfilerOptions evaluates and appends enable-cloud-profiler-for-sidecar, pod-name, and pod-uid
+// mount options if enabled in driver options and supported by the container image.
+func (s *nodeServer) appendCloudProfilerOptions(mounterImage string, enableCloudProfiler bool, podName string, podUID string, mountOptions []string) []string {
+	if !enableCloudProfiler || !s.driver.isSidecarVersionSupportedForGivenFeature(mounterImage, SidecarCloudProfilerMinVersion) {
+		return mountOptions
+	}
+	if podName == "" {
+		klog.Warningf("Pod name is empty, skipping cloud profiler mount options")
```
yaozile123: Can you replace to `klog.Warning` since there is no format at here?

pkg/csi_driver/node.go
```diff
@@ -773,6 +765,27 @@ func (s *nodeServer) populateTokenAndBucketAccessCheckOptions(pod *corev1.Pod, i
...
+		return mountOptions
+	}
+	if podUID == "" {
+		klog.Warningf("Pod UID is empty, skipping cloud profiler mount options")
```
yaozile123: Same as above, can you replace to `klog.Warning`?

pkg/csi_driver/node.go
chrisThePattyEater: added

pkg/csi_driver/node.go
chrisThePattyEater: added

cmd/sidecar_mounter/main.go
```diff
@@ -50,11 +47,17 @@ var (
 	storageServiceAndBucketAccessJitter   = flag.Float64("storage-service-check-retry-jitter", 0.1, "storage service creation and bucket access check exponential retry jitter")
 	storageServiceAndBucketAccessDuration = flag.Duration("storage-service-check-retry-duration", 5*time.Second, "storage service creation and bucket access check exponential retry initial duration")
 	enableSharedMount                     = flag.Bool("enable-shared-mount", false, "whether to run gcsfuse in a mounter pod")
+	enableCloudProfiler                   = flag.Bool("enable-cloud-profiler", false, "whether to enable Cloud Profiler for the sidecar mounter container")
```
uriel-guzman: Why do we need a new feature flag for cloud profiler?

cmd/sidecar_mounter/main.go
```diff
@@ -50,11 +47,17 @@ var (
...
 
 	// This is set at compile time.
 	version = "unknown"
 )
 
+const (
+	cloudProfilerSidecarServiceName     = "gke-gcsfuse-sidecar"
```
uriel-guzman: This already exists in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/b8f457c3bbd5c301c0961444f7ceea0b8e4cd85b/pkg/webhook/sidecar_spec.go#L30 and https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/b8f457c3bbd5c301c0961444f7ceea0b8e4cd85b/pkg/util/util.go#L80

cmd/sidecar_mounter/main.go
```diff
@@ -127,23 +132,7 @@ func main() {
...
-							ServiceVersion: serviceVersion,
-						}
-						if err := profiler.Start(cfg); err != nil {
-							klog.Errorf("Errored while starting cloud profiler, got %v", err)
-						} else {
-							klog.Infof("Running cloud profiler on %s with version %s", cfg.Service, cfg.ServiceVersion)
-						}
-					})
-				}
+				sidecarmounter.StartCloudProfiler(cloudProfilerSidecarServiceName, mc.PodName, mc.PodUID)
```
uriel-guzman: The "enable-cloud-profiler" is just ignored for the gke-gcsfuse-sidecar, so what is the point of having a feature flag if it's just ignored anyway? I think we should remove it, since it doesn't exist anyway.

cmd/sidecar_mounter/main.go
```diff
@@ -101,6 +104,9 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	if *enableSharedMount {
+		if *enableCloudProfiler {
+			sidecarmounter.StartCloudProfiler(cloudProfilerSharedMountServiceName, os.Getenv("POD_NAME"), os.Getenv("POD_UID"))
```
uriel-guzman: In gke-gcsfuse-sidecar, cloud profiler doesn't actually start until NodePublishVolume succeeds, because it sends the `EnableCloudProfilerForSidecarConst` as a mount option, and the sidecar waits until the sockets have been populated by the CSI driver before doing anything.

In shared mount, you're making the mounter pod immediately start cloud profiler on startup, which introduces more complexity (hence the extra logic you added for the env vars and feature flag)

To make it consistent, why can't we just start cloud profiler until  `EnableCloudProfilerForSidecarConst` is passed as a mount option via Mount gRPC call? This way, it'd only start until NodeStageVolume is called?

This also has the benefit of that, since you're waiting for Mount to be called before starting cloud profiler, you can just re-use pod name and pod UID, which you are passing anyway as mount options here (and not using in shared mount, you're just populating them for no real reaason):
```
args.fuseMountOptions = s.appendCloudProfilerOptions(podImage, args.enableCloudProfilerForSidecar, podName, string(pod.UID), args.fuseMountOptions)
```
This way, you can also get rid of the POD_NAME and POD_UID env vars you created in the createMounterPod function, and also get rid of the EnableCloudProfilerForSidecarConst check you have in controller.go, since it's also redundant because you only populate `enable-cloud-profiler-for-sidecar` with  EnableCloudProfilerForSidecarConst anyway during NodeStageVolume, so all of this logic seems redundant?

uriel-guzman (changes requested):

cmd/sidecar_mounter/main.go
chrisThePattyEater: removed

cmd/sidecar_mounter/main.go
```diff
@@ -116,7 +113,6 @@ func main() {
 	if err != nil {
 		klog.Fatalf("failed to read defaulting-flag file: %v", err)
 	}
```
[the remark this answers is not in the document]
uriel-guzman: Please update PR description, since it has an obsolete description.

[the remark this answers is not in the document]
chrisThePattyEater: updated the description since this no longer is valid

uriel-guzman (approved):

PR #1560, opened 2026-07-23
Implement start token server for host network in shared mount

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Extract the HostNetwork token server startup logic from `Mount()` into a shared helper `startHostNetworkTokenServer()`, and call it in `MountToNode()` to implement host network KSA support for the shared node mount path.

This resolves the TODO in MountToNode() that previously left the host network path unimplemented.

Fixes #

```release-note

```

### DIFF
--- a/pkg/sidecar_mounter/sidecar_mounter.go
+++ b/pkg/sidecar_mounter/sidecar_mounter.go
@@ -81,23 +81,19 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	var tokenSource oauth2.TokenSource
 
 	// TODO(amacaskill): Add support for hostnetwork pods on OSS k8s.
-	if mc.HostNetworkKSAOptIn {
-		if mc.TokenServerIdentityProvider != "" {
-			klog.V(4).Infof("Pod has hostNetwork enabled and token server feature is supported and opted in. Starting Token Server on %s/%s", mc.TempDir, TokenFileName)
-			go StartTokenServer(ctx, mc.TempDir, TokenFileName, mc.TokenServerIdentityProvider)
-		} else {
-			return fmt.Errorf("HostNetwork Pod KSA feature is opted in, but token server identity provider is not set. Please set identityProvider in VolumeAttributes")
+	if err := startHostNetworkTokenServer(ctx, mc); err != nil {
+		return err
+	}
+
+	if mc.HostNetworkKSAOptIn && mc.EnableSidecarBucketAccessCheck {
+		// Fetch custom tokensource and audience for host network path. For workload identity, tokenSource is not needed and the audience is hardcoded in TokenSource.FetchIdentityBindingToken().
+		audience, err := getAudienceFromContextAndIdentityProvider(ctx, mc.TokenServerIdentityProvider)
+		if err != nil {
+			return fmt.Errorf("failed to get audience from the context: %w", err)
 		}
-		if mc.EnableSidecarBucketAccessCheck {
-			// Fetch custom tokensource and audience for host network path. For workload identity, tokenSource is not needed and the audience is hardcoded in TokenSource.FetchIdentityBindingToken().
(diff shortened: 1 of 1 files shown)


pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -161,6 +157,22 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
 	return nil
 }
 
+// startHostNetworkTokenServer starts the local token server if hostNetwork KSA feature is opted in.
+func startHostNetworkTokenServer(ctx context.Context, mc *MountConfig) error {
+	if !mc.HostNetworkKSAOptIn {
+		return nil
+	}
```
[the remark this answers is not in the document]
yaozile123: Done

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -172,7 +187,10 @@ func (m *Mounter) MountToNode(ctx context.Context, mountServerContext context.Co
 	}
 	mc.EnsureErrWriter()
 
-	// TODO(yaozile) Implement the host network path for the sidecar mounter, where the token server is started in the sidecar mounter.
+	// Start the token server for HostNetwork enabled mounter pods.
+	if err := startHostNetworkTokenServer(mountServerContext, mc); err != nil {
+		return status.Errorf(codes.InvalidArgument, "failed to start host network token server: %v", err)
```
uriel-guzman: If `mc == nil`, this will wrap that into an `InvalidArgument` and not alert us.

In sidecar mode, the errors returned by `startHostNetworkTokenServer` are directly written to the error file: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/cmd/sidecar_mounter/main.go#L165

Then, they are read by node publish: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L426

And then they are supposed to be categorized into an error code based on string parsing here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/utils.go#L550

However, I don't see how these errors return anything other than `Unknown`.

Can you do a quick verification to check which error codes are returned by NodePublish for the errors in `startHostNetworkTokenServer` for sidecar architecture? Whatever they are, we must make the mounter pod return the same error codes to remain consistent.

uriel-guzman (changes requested):

pkg/sidecar_mounter/sidecar_mounter.go
yaozile123: I agree, the error code should be `Internal` rather than `InvalidArgument`. I have updated the code

pkg/sidecar_mounter/sidecar_mounter.go
```diff
@@ -161,6 +157,25 @@ func (m *Mounter) Mount(ctx context.Context, mc *MountConfig) error {
...
+func startHostNetworkTokenServer(ctx context.Context, mc *MountConfig) error {
+	if mc == nil {
+		return fmt.Errorf("mount config cannot be nil")
+	}
+	if !mc.HostNetworkKSAOptIn {
+		return nil
+	}
+
+	if mc.TokenServerIdentityProvider == "" {
+		return fmt.Errorf("HostNetwork Pod KSA feature is opted in, but token server identity provider is not set. Please set identityProvider in VolumeAttributes")
```
uriel-guzman, line 170: should this be Internal or InvalidArgument? Internal should be a bug within our code, InvalidArgument should be a user mistake who missed providing a required argument. In L192 you are always returning Internal.

uriel-guzman (approved):

PR #1564, opened 2026-07-28
[Shared Mount] Pass default mount options

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

 In the sidecar architecture, the CSI driver manually opens the FUSE device via /bin/mount using a 
  set of default mount arguments (see prepareMountOptions). These arguments, along with stripped user-
  provided options, are passed to mount the FUSE device and are removed from the list of arguments    
  passed to the gcsfuse process itself.                                                               
                                                                                                      
In Shared Node Mount mode, gcsfuse opens its own FUSE device directly because the mounter pod runs
  as root (UID 0). Consequently, we do not need to filter out "o=" mount options or construct low-    
  level kernel mount flags. Instead, we use prepareSharedNodeMountOptions to generate a streamlined   
  list of default parameters (e.g., {"o=allow_other,default_permissions"}).
  
Explanation of FUSE Parameters (Needed vs. Unneeded)[https://docs.google.com/document/d/1RW3tyswe0PWAKk6mN8K2zWU94Ln-xNB87RA5qTrduJs/edit?tab=t.0]

 ### Summary of Changes
  
  1. Options Processing Helper (pkg/csi_driver/utils.go):
      • Created prepareSharedNodeMountOptions() to process mount options for Shared Node Mount mode:  
          • Automatically prepends mandatory FUSE options: o=allow_other,default_permissions.         
          • Strips read_ahead_kb= options (preventing gcsfuse flag validation errors).
  2. Node Server Staging Integration (pkg/csi_driver/node.go):
      • Updated executeNodeStageVolume() to pass "o=" options through prepareSharedNodeMountOptions() 
      before invoking s.mountToNode().
  3. Sidecar Mounter Flag Handling (pkg/sidecar_mounter/sidecar_mounter_config.go):
      • Bypassed the -o flag restriction in DisallowedFlags when SharedMountPoint is active (allowing 
      -o flags in Shared Node Mount mode).
      • Updated flag parsing to concatenate multiple -o options with commas, this prevents o= args from overwriting eachother (e.g., merging            
      o=allow_other,default_permissions and o=ro into -o allow_other,default_permissions,ro instead of o=ro being the final value) 
    4.  rw and ro options are directly passed to the host mount and any pod spec rw/ro is applied at bind mount. This facilitates different levels of access across pods using the same mount.
  
  
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -796,3 +796,21 @@ func getInternalMountOptionValue(options []string, key string) string {
 	}
 	return ""
 }
+
+// prepareSharedNodeMountOptions processes and filters mount options to be directly passed to the gcsfuse process during shared node mounts.
+// Unlike prepareMountOptions (which splits options between those used to open fuse device and those passed to gcsfuse).
+func prepareSharedNodeMountOptions(options []string) []string {
+	// TODO(FUECHR): Investigate if we should only allow "allowedOptions" from prepareMountOptions for shared node mount.
+
+	csiMountOptions := []string{"o=allow_other,default_permissions"}
+	for _, o := range options {
+		// Strip read_ahead_kb and node_fuse_max_request_limit_kb flags so they are not passed to gcsfuse (which would reject them as unknown flags).
+		if strings.HasPrefix(o, "read_ahead_kb=") || strings.HasPrefix(o, "node_fuse_max_request_limit_kb=") {
+			continue
+		}
+
+		csiMountOptions = append(csiMountOptions, o)
+	}
+
+	return csiMountOptions
+}
--- a/pkg/sidecar_mounter/sidecar_mounter_config.go
+++ b/pkg/sidecar_mounter/sidecar_mounter_config.go
@@ -290,9 +290,13 @@ func (mc *MountConfig) prepareMountArgs() {
 
 		flag := argPair[0]
 		if _, ok := DisallowedFlags[flag]; ok {
-			invalidArgs = append(invalidArgs, arg)
-
-			continue
+			// Reject all disallowed flags EXCEPT "-o" when SharedMountPoint is set.
+			// SharedMountPoint is set when the shared mount feature is enabled, during which the "-o" flag is needed to pass mount options to gcsfuse.
+			isAllowedOFlag := flag == "o" && mc.SharedMountPoint != ""
+			if !isAllowedOFlag {
+				invalidArgs = append(invalidArgs, arg)
+				continue
+			}
 		}
 
 		value := ""
@@ -376,7 +380,18 @@ func (mc *MountConfig) prepareMountArgs() {
 			value = GCSFuseAppName + "-" + value
 		}
 
-		flagMap[flag] = value
+		// The "-o" flag can be passed multiple times, and gcsfuse expects them to be comma-separated.
+		if flag == "o" {
+			if value != "" {
+				if existing, ok := flagMap["o"]; ok && existing != "" {
+					flagMap["o"] = existing + "," + value
+				} else {
+					flagMap["o"] = value
+				}
+			}
+		} else {
+			flagMap[flag] = value
+		}
 	}
 
 	if len(invalidArgs) > 0 {
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -1105,8 +1105,13 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		return nil, err
 	}
 
+	// Prepare mount options specifically for shared node mount architecture before sending to mounter pod.
+	preparedMountOptions := prepareSharedNodeMountOptions(args.fuseMountOptions)
+
+	// TODO(FUECHR): Pass read_ahead_kb and node_fuse_max_request_limit_kb to gcsfuse config correctly for shared node mounts.
+
 	// Send GRPC to mounter pod to start GCSFuse.
-	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, args.fuseMountOptions); err != nil {
+	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, preparedMountOptions); err != nil {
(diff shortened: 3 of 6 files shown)


pkg/csi_driver/utils.go
```diff
@@ -796,3 +796,34 @@ func getInternalMountOptionValue(options []string, key string) string {
...
+			continue
+		}
+
+		// Strip read_ahead_kb mount flags so they are not passed to gcsfuse (which would reject them as unknown flags).
+		// TODO(FUECHR): Pass read_ahead_kb to gcsfuse config correctly for shared node mounts.
+		// These need to be stripped and passed via config file (see sidecar achitecture mount for more info csi_mounter.go/mount)
+		if strings.HasPrefix(o, "read_ahead_kb=") {
+			continue
+		}
+
```
yaozile123, line 811: Do we need to strip `node_fuse_max_request_limit_kb` as well? It's handled the same way as `read_ahead_kb in `prepareMountOptions()`. It would also rejected by gcsfuse as unknown flags if it got passed directly to gcsfuse.

https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L445-L463

chrisThePattyEater: yup, this was added 2 days ago. I think it should be added in the follow up pr with other kernel params like sysfsBDI. Ill add a todo for implementation and add a strip.

chrisThePattyEater: this is the pr https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/commit/6e6f431e92a8aad8a1e02fa20330a2211d5af3b9

pkg/csi_driver/utils.go
```diff
@@ -796,3 +796,21 @@ func getInternalMountOptionValue(options []string, key string) string {
 	}
 	return ""
 }
+
+// prepareSharedNodeMountOptions processes and filters mount options to be directly passed to the gcsfuse process during shared node mounts.
+// Unlike prepareMountOptions (which splits options between those used to open fuse device and those passed to gcsfuse).
+func prepareSharedNodeMountOptions(options []string) []string {
```
yaozile123, line 802: In your PR description, you mentioned:
> • Converts standard ro flags into FUSE option syntax (`o=ro`).


However, I don't see this implemented in `prepareSharedNodeMountOptions`. Currently, raw `"ro"` and `"rw"` are passed through unmodified, and the unit test in `pkg/csi_driver/utils_test.go:241-244` also expects them as raw strings. Could you clarify or add the conversion logic?

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -290,9 +290,11 @@ func (mc *MountConfig) prepareMountArgs() {
 
 		flag := argPair[0]
 		if _, ok := DisallowedFlags[flag]; ok {
-			invalidArgs = append(invalidArgs, arg)
-
-			continue
+			// Reject all disallowed flags EXCEPT "-o" when SharedMountPoint is set.
+			if flag != "o" || mc.SharedMountPoint == "" {
```
uriel-guzman: 1. According to the [Go Style Guide (go/style/guide#clarity)](https://g3doc.corp.google.com/go/g3doc/style/guide.md), complex inline negations and binary logic have a high cognitive load and are prone to logic errors.

Instead of forcing the reader to parse negated binary logic, it is much easier to read if we extract the positive exception case into a descriptive boolean variable and then negate that single variable.

```
	if _, ok := DisallowedFlags[flag]; ok {
		// Reject all disallowed flags EXCEPT "-o" when SharedMountPoint is set.
		isAllowedOFlag := flag == "o" && mc.SharedMountPoint != ""
		if !isAllowedOFlag {
			invalidArgs = append(invalidArgs, arg)
			continue
		}
	}
```

2. Also, this comment says what it does, but not why, improve the comment to explain why we allow the -o flag in shared mount mode, but not in sidecar mode.

uriel-guzman (changes requested):

pkg/sidecar_mounter/sidecar_mounter_config.go
```diff
@@ -376,7 +378,18 @@ func (mc *MountConfig) prepareMountArgs() {
 			value = GCSFuseAppName + "-" + value
 		}
 
-		flagMap[flag] = value
+		// The "-o" flag can be passed multiple times, and gcsfuse expects them to be comma-separated.
```
uriel-guzman, line 383: Do we actually need to do this? In https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L429, I don't see this sort of logic being applied. It seems like if the customer passed `-o=exec,async`, you would just get a warning and the mount options wouldn't be passed: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L434, why would the behavior be different in shared mount?

The only reason I can think of is because you need this is because you are passing the default mount options as:
`[]string{"o=allow_other,default_permissions"}`, however, this is different than how the CSI Driver currently, passes it, which just passes a slice of strings into https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L121 `MountSensitiveWithoutSystemdWithMountFlags`, so why is this also handled differently?

chrisThePattyEater: Discussion offline led us to understand that this is needed since the second value of o would overwrite the first in this go map

pkg/sidecar_mounter/sidecar_mounter_config.go
chrisThePattyEater: fixed

uriel-guzman (approved):

PR #1565, opened 2026-07-28
update gcs fuse test setup to support prow job runs

Sneha-at:
<!--  Thanks for sending a pull request!  Here are some tips for you:

1. If this is your first time, please read our contributor guidelines: https://git.k8s.io/community/contributors/guide#your-first-contribution and developer guide https://git.k8s.io/community/contributors/devel/development.md#development-guide
2. Follow the instructions for writing a release note: https://git.k8s.io/community/contributors/guide/release-notes.md
3. If the PR is unfinished, see how to mark it: https://git.k8s.io/community/contributors/guide/pull-requests.md#marking-unfinished-pull-requests
-->

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

 /kind failing-test
> /kind feature
> /kind flake

## Decoupling Cluster Lifecycle and Boskos from `--run-in-prow`

### 1. New Flags Added
* **`--manage-cluster-lifecycle`** (`bool`, default: `false`)
  * **Description:** Controls whether the test framework automatically provisions (`gcloud container clusters create`) and tears down (`gcloud container clusters delete`) a GKE cluster for the duration of the test suite. This creates the with same configuration as the prow job, the configuration is controlled through flags [here](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/6fb9447c161be5ccab14d9fea1c91ec7fd9137f5/test/e2e/utils/cluster.go#L94-L107)
* **`--use-boskos`** (`bool`, default: `false`)
  * **Description:** Controls whether the test framework acquires an ephemeral GCP project dynamically from a Boskos resource server (`boskos.Acquire`) over HTTP.
* **`--project-id`** (`string`, default: `""`)
  * **Description:** Specifies the explicit GCP project ID where the cluster and test workloads reside when running without Boskos. If omitted (`""`), the runner automatically discovers the project ID from the active gcloud configuration (`gcloud config get-value project`).

---

### 2. Why They Were Added (Rationale)
Previously, the single flag `--run-in-prow` tightly coupled three separate infrastructure concerns:
1. Dynamic project leasing via Boskos.
2. Automated GKE cluster lifecycle creation and deletion.
3. Prow-specific API overrides and environment variable injection.

This tight coupling prevented modular setups—such as managing cluster creation inside a fixed developer project without Boskos, or testing Boskos leasing outside of Prow pods. Decoupling these behaviors provides major architectural benefits:
* **Modular Infrastructure Control:** Allows any combination of cluster lifecycle management and project acquisition strategies.
* **Elimination of Race Conditions:** Explicitly threading `testParams.ProjectID` into all `gcloud container clusters` commands guarantees operations execute against the intended project without relying on global gcloud configuration state.

---

### 3. 100% Backward Compatibility Guarantee for Older Prow Jobs
A core design requirement of this PR is zero disruption to older or existing Prow jobs across GoB and open-source CI environments.

#### How Backward Compatibility is Enforced
All existing Prow jobs invoke `test/e2e/run-e2e-ci.sh`, which runs `e2e-test-ci` passing `--run-in-prow=true`.
In `test/e2e/main.go`, the framework checks which CLI flags were explicitly provided by the caller (`flag.Visit`):

```go
    if *inProw {
        if !explicitFlags["manage-cluster-lifecycle"] {
            *manageClusterLifecycle = true
        }
        if !explicitFlags["use-boskos"] {
            *useBoskos = true
        }
    }
```

Because existing Prow jobs do not pass `--manage-cluster-lifecycle` or `--use-boskos`, passing `--run-in-prow=true` automatically overrides and defaults both flags to `true`.

#### Impact on Existing Jobs
* **Zero Job YAML / Script Modifications Required:** Older Prow jobs do not need any updates to their job definitions, shell scripts, or command-line arguments.
* **Out-of-the-Box Behavior:** They will continue to automatically lease ephemeral projects from Boskos and provision/destroy GKE clusters exactly as before.

---

### 4. Supported Test Setups & Execution Configurations

| Setup | Flags | Description & Workflow |
| :--- | :--- | :--- |
| **A. Standard Prow CI** | `--run-in-prow=true` | **Default CI Workflow:** Automatically sets `--manage-cluster-lifecycle=true` and `--use-boskos=true`. Leases an ephemeral project from Boskos, creates a fresh GKE cluster, executes the e2e test suite, deletes the cluster, and releases the project back to Boskos. |
| **B. Fixed Project Automation** | `--manage-cluster-lifecycle=true`<br>`--project-id=<project>` | **Automated Non-Boskos Runs:** Ideal for nightly benchmarks or dedicated CI pipelines. Automatically provisions and tears down a fresh GKE cluster inside a specific, pre-allocated GCP project without talking to Boskos. |
| **C. Fast Local Iteration** | `--manage-cluster-lifecycle=false`<br>`--project-id=<project>` *(optional)* | **Day-to-Day Developer Testing:** Executes tests immediately against an existing, already-running GKE cluster using local `gcloud`/`kubectl` credentials. Skips cluster creation and teardown overhead. |
| **D. Boskos Debugging** | `--use-boskos=true`<br>`--manage-cluster-lifecycle=true` | **Infrastructure Debugging:** Allows developers or infrastructure engineers to test Boskos leasing and release logic directly from a workstation or script outside of a Prow pod container. |


Full commands for local runs
PROJECT_ID is optional, it is picked up from the current cluster context if not provided.

B. Fixed Project Automation 
```
make e2e-test \                                                                                                                                                                                           
      REGISTRY=us-central1-docker.pkg.dev/$PROJECT_ID/gcs-fuse-csi-driver \                                                                                                                           
      STAGINGVERSION=prow-gob-internal-boskos-1 \                                                                                                                                                             
      E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true \                                                                                                                                                                
      E2E_TEST_PROJECT_ID=$PROJECT_ID
```

C.  Fast Local Iteration
Non-Managed Driver:                                                                                                                                                                                         
```                                                                                                                                                                                                          
    make e2e-test \                                                                                                                                                                                           
      REGISTRY=us-central1-docker.pkg.dev/$PROJECT_ID/gcs-fuse-csi-driver \                                                                   
      STAGINGVERSION=prow-gob-internal-boskos-1 \                                                
      E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=false                                                                                                                                                     
 ```                                                                                                                                                                                                        
  Managed Driver:                                                                                                                                                                                             
 ```                                                                                                                                                                                                        
    make e2e-test \                                                                                                                                                                                           
      E2E_TEST_USE_GKE_MANAGED_DRIVER=true \                                                        
      E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=false                                                   
```
D. Boskos Debugging
```
make e2e-test \                                                                                                                                                                                           
      E2E_TEST_USE_BOSKOS=true \                                                                                                                                                                              
      E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true \                                             
      STAGINGVERSION=prow-gob-internal-boskos-1 
```

**Logging access**

I have created a google group to manage the GCS Bucket access for PR logs to non-googlers. Anyone contributing to the GCS Fuse CSI driver can be added to `gcs-fuse-csi-contributors@google.com` group. Googlers should have access to the logs by default. I haven't been able to test this yet with any non-google account.


**Testing**

I ran the whole testsuite but 2 of the cloud_profiler tests failed so did not get the summary, analyzed the logs through Jetski
```
   make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false STAGINGVERSION=$STAGINGVERSION REGISTRY=$REGISTRY GKE_CLUSTER_REGION=$GKE_CLUSTER_REGION E2E_SKIP="should.succeed.in.performance.test|Feature:.workload-identity-federation|Feature:.OIDC|Feature:.GCSFuse-PDCSI"                                                                                                                                                             
  • Total tests run: 473                                                                                                                                        
  • Succeeded (Passed): 471                                                                                                                                     
  • Failed: 2 (Both were the Cloud Profiler integration tests: one for CPU profiling and one for Heap profiling).   
```

<!--
*Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
_If PR is about `failing-tests or flakes`, please post the related issues/tests in a comment and do not use `Fixes`_*
-->
Fixes #

<!--
If no, just write "NONE" in the release-note block below.
If yes, a release note is required:
Enter your extended release note in the block below. If the PR requires additional action from users switching to the new release, include the string "action required".
-->
```release-note
NONE
```


test/e2e/utils/cluster.go
```diff
@@ -62,9 +63,27 @@ func gcloudCommand(testParams *TestParameters, args ...string) *exec.Cmd {
...
+	fullArgs = append(fullArgs, args...)
+
+	//nolint:gosec
+	return exec.CommandContext(ctx, gcloudBin, fullArgs...)
+}
+
 func clusterDownGKE(testParams *TestParameters) error {
 	//nolint:gosec
-	cmd := exec.Command("gcloud", "container", "clusters", "delete", testParams.GkeClusterName, "--region", testParams.GkeClusterRegion, "--quiet")
+	cmd := gcloudCommand(testParams, "container", "clusters", "delete", testParams.GkeClusterName, "--region", testParams.GkeClusterRegion, "--project", testParams.ProjectID, "--quiet")
```
Sneha-at, line 67: This was updated due to an older Gemini review comment
>Use the helper function gcloudCommand instead of directly calling exec.Command("gcloud", ...) to ensure that any custom gcloud binary (GkeGcloudCommand) or additional arguments (GkeGcloudArgs) configured in TestParameters are respected.

test/e2e/run-e2e-local.sh
```diff
@@ -19,9 +19,12 @@ set -o xtrace
 set -o nounset
 set -o errexit
 
+export JOB_NAME="${JOB_NAME:-local-e2e-test}"
+
 readonly PKGDIR=$(realpath "$( dirname -- "$0"; )/../..")
 readonly gke_cluster_region=${GKE_CLUSTER_REGION:-us-central1}
-readonly gke_cluster_version=$(kubectl version 2>/dev/null | grep 'Server Version:' | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+(-gke\.[0-9]+)?$')
+gke_cluster_version_detected=$(kubectl version 2>/dev/null | grep 'Server Version:' | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+(-gke\.[0-9]+)?$' || echo "")
```
Sneha-at, line 26: Earlier local tests were only run in existing clusters so we assumes cluster exists, but this PR allows us to create and delete a cluster for local run just like prow so we have to fetch the kubectl version.

test/e2e/run-e2e-ci.sh
```diff
@@ -28,10 +28,13 @@ readonly use_gke_autopilot=${USE_GKE_AUTOPILOT:-false}
 readonly cloudsdk_api_endpoint_overrides_container=${CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER:-https://container.googleapis.com/}
 readonly use_gke_managed_driver="${USE_GKE_MANAGED_DRIVER:-true}"
 readonly ginkgo_focus="${TEST_FOCUS:-}"
-readonly ginkgo_skip="${TEST_SKIP:-}"
+raw_skip="${TEST_SKIP:-}"
```
Sneha-at, line 31: This appends any skips added in this file with skips added through prow jobs

test/e2e/testsuites/gcsfuse_oidc_auth.go
```diff
@@ -540,7 +540,7 @@ func (t *gcsFuseCSIOIDCTestSuite) DefineTests(driver storageframework.TestDriver
 // Helper functions for GCP operations
 
 func getProjectNumber(projectID string) string {
-	cmd := exec.Command("gcloud", "projects", "describe", projectID, "--format=value(projectNumber)")
+	cmd := exec.Command("gcloud", "projects", "describe", projectID, "--format=value(projectNumber)", "--billing-project="+projectID)
```
Sneha-at: OIDC tests refer to this flag so we pass it to ensure correct billing project i.e. boskos leased project is used.

test/e2e/run-e2e-local.sh
```diff
@@ -91,5 +104,5 @@ base_cmd="${PKGDIR}/bin/e2e-test-ci \
             --enable-sidecar-bucket-access-check=${enable_sidecar_bucket_access_check} \
             --enable-gcsfuse-profiles=${enable_gcsfuse_profiles} \
             --enable-gcsfuse-kernel-params=${enable_gcsfuse_kernel_params} \
-            --gcsfuse-pr-number=${gcsfuse_pr_number}"
+            ${gcsfuse_pr_number:+--gcsfuse-pr-number=${gcsfuse_pr_number}}"
```
amacaskill, line 107: is this a typo?

test/README.md
```diff
@@ -113,6 +116,11 @@ make e2e-test E2E_TEST_USE_GKE_MANAGED_DRIVER=false E2E_TEST_BUILD_DRIVER=true \
   GCSFUSE_PR_NUMBER=4651 STAGINGVERSION=v999.999.999 REGISTRY=my-registry
 # Run the test with customized Ginkgo flags.
 make e2e-test E2E_TEST_FOCUS=gcsfuseIntegration E2E_TEST_SKIP=failedMount E2E_TEST_GINKGO_PROCS=3 E2E_TEST_GINKGO_TIMEOUT=20m E2E_TEST_GINKGO_FLAKE_ATTEMPTS=1
+
+# Boskos Debugging: Test Boskos leasing and cluster lifecycle management locally.
```
amacaskill, line 121: When would you use this and how do you set up the boskos server?  

> 4. Supported Test Setups & Execution Configurations

The PR description outlines several great test setups. Since we already have a lot of documentation for manual testing, could we make sure these configurations are added there too? Or distinguish when we add what / when?

test/README.md
```diff
@@ -100,6 +100,9 @@ You can control the test through the following make parameters, eg `make e2e-tes
 - `E2E_TEST_GINKGO_FLAKE_ATTEMPTS`: default value is `2`. The value will be passed to `ginkgo run --flake-attempts` flag.
 - `ENABLE_ZB`: default value is `false`. Change it to `true` if you want the bucket used during testing to be a Zonal Bucket created in the zone -> "$(location provided)" + "-c".
 - `GCSFUSE_CLIENT_PROTOCOL`: default value is 'http1'. Change to 'grpc' to alter the type of protocol gcsfuse uses to communicate with gcs
+- `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE`: default value is `false`. Change it to `true` if you want the test runner to create and destroy the GKE cluster for the test.
```
amacaskill, line 103: Can you please explain what type of cluster this creates, what project it creates it in , etc.   Is a flag saying create GKE cluster more clear?

Also what about the E2E_TEST_GKE_CLUSTER_VERSION flag?, what does that do?

amacaskill:
Did we confirm the tests are still passing locally? What we did before to test our changes to make sure we didn't add a regression?

test/e2e/run-e2e-local.sh
Sneha-at: My tests were failing as the `gcsfuse_pr_number` flag was getting appended to the command without any value like `--gcsfuse-pr-number=""`. This modification basically adds the flag to the command only if a GCS Fuse PR number is passed.

test/README.md
Sneha-at: I am adding more detailed steps on how to run and what to run exactly (https://docs.google.com/document/d/1jRom6W_fl0vyyLestGFZbpNjZgflChUan2yAcYtEO7Y/edit?pli=1&tab=t.byg3izdtxoxz). this PR only adds the flags to the README. May be it is better to remove any readme changes for now and followup with detail documentation.

test/e2e/testsuites/gcsfuse_oidc_auth.go
amacaskill: Why didn't we need this before?

test/e2e/utils/project.go
```diff
@@ -29,7 +29,15 @@ import (
...
-var boskos, _ = boskosclient.NewClient(os.Getenv("JOB_NAME"), "http://boskos", "", "")
+func getEnvWithDefault(key, defaultValue string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return defaultValue
+}
+
+var boskosURL = getEnvWithDefault("BOSKOS_URL", "http://boskos")
+var boskos, _ = boskosclient.NewClient(os.Getenv("JOB_NAME"), boskosURL, "", "")
```
amacaskill, line 40: Why do we want to ignore the error?

test/e2e/utils/handler.go
```diff
@@ -112,9 +115,12 @@ func Handle(testParams *TestParameters) error {
 
 	// Always ensure ProjectID is set and PROJECT env var is exported for all test runs.
 	if testParams.ProjectID == "" {
-		output, err := exec.Command("gcloud", "config", "get-value", "project").CombinedOutput()
+		klog.Infof("Getting default gcloud project...")
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
amacaskill: Should this be passed from caller instead of creating a new context.Background () here?

amacaskill: Also why is this needed/ why does it hang?

test/e2e/testsuites/gcsfuse_oidc_auth.go
Sneha-at: Prow jobs currently skip OIDC tests https://gke-internal.googlesource.com/test-infra/+/refs/heads/master/prow/gob/config_gen/gke_gcs_fuse_e2e_gen.py#150 while local testing use only 1 cluster. The issue we had here was that the build project `oss-cloud-kubernetes-test` was considered for quota check instead of the leased project. This clarifies the billing-project specifically, it doesn't change any test behavior.

test/e2e/utils/project.go
Sneha-at: This is inline with the old implementation, if boskos client fails we will fail to fetch the resource, I just didn't want to update the older logic here.

test/e2e/utils/cluster.go
```diff
@@ -136,10 +156,15 @@ func clusterUpGKE(testParams *TestParameters) error {
 	if !testParams.UseGKEAutopilot {
 		cmdParams = append(cmdParams, standardClusterFlags...)
 
-		// Update gcloud to latest version.
-		cmd = exec.Command("gcloud", "components", "update")
-		if err := runCommand("Updating gcloud to the latest version", cmd); err != nil {
-			return fmt.Errorf("failed to update gcloud to latest version: %w", err)
+		// Update gcloud to latest version in Prow.
+		if testParams.InProw {
+			cmd = exec.Command("gcloud", "components", "update")
```
yaozile123, line 142: Can we also use `gcloudCommand(testParams...)` at here?

test/e2e/utils/project.go
```diff
@@ -84,7 +92,7 @@ func setEnvProject(project string) error {
 }
 
 func setEnvProjectNumberUsingID(projectID string) (string, error) {
-	cmd := exec.Command("gcloud", "projects", "describe", projectID, "--format=value(projectNumber)")
+	cmd := exec.Command("gcloud", "projects", "describe", projectID, "--format=value(projectNumber)", "--billing-project="+projectID)
```
yaozile123: Should we use `gcloudCommandContext` at here?

test/e2e/testsuites/gcsfuse_oidc_auth.go
yaozile123: Same here, can we use `gcloudCommandContext` instead?

test/README.md
Sneha-at: Boskos server is already up an running in the build host project `oss-cloud-kubernetes-test` and cluster ` gcs-fuse-e2e-build-cluster-regional` the steps to connect to the cluster and the server are added in the document above

test/README.md
Sneha-at: This creates a cluster with same configuration we use for prow jobs which is set by the flags we have [here](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/test/e2e/main.go#L34-L39). For now this will create a regional cluster with 9 nodes https://screenshot.googleplex.com/5V5RBFiePTeRtpq

test/e2e/utils/handler.go
Sneha-at: There is no context created in main.go currently, we can update the logic though to create it in main.go and then pass down. As per why it was needed : The tests would sometimes hang while fetch the project details if the metadata server connection is not yet up, in that case it doesn't log anything.

test/e2e/testsuites/gcsfuse_oidc_auth.go
Sneha-at: I am actually skipping the OIDC tests in presubmits too so I will remove this for now.  https://github.com/GoogleCloudPlatform/oss-test-infra/blob/b4434aa7888eedf22f95df6997712059bd3603a7/prow/prowjobs/GoogleCloudPlatform/gcs-fuse-csi-driver/gcs-fuse-csi-driver-config.yaml#L33

Sneha-at: Is this to ensure none of the gcloud commands hang indefinetely?

test/e2e/utils/project.go
Sneha-at: Many of these functions do not have testParams inherited in these functions, and these commands are pretty straight forward so I think it is alright if we do not change to gcloudCommand for some of these commands

test/e2e/utils/handler.go
Sneha-at: Removed the context here.

test/e2e/utils/handler.go
amacaskill, line 182: The defer clusterDownGKE routine is registered after clusterUpGKE has returned successfully. If clusterUpGKE fails (for example, if the GKE create API call succeeds but the subsequent update command in  clusterUpGKE  times out OR if 
	if err := os.Setenv(ClusterLocationEnvVar, testParams.GkeClusterRegion); err != nil {

fails, then we are leaking clusters. Please change it to 

```
	// 4. Tear down the cluster afterwards.
		defer func() {
			if err := clusterDownGKE(testParams); err != nil {
				klog.Errorf("failed to cluster down: %v", err)
			}
		}()
		if err := clusterUpGKE(testParams); err != nil {
			return fmt.Errorf("failed to cluster up: %w", err)
		}

```

I know you didn't add this, but presubmit will make this run more, and I don't want to leak a lot of clusters.

test/e2e/utils/handler.go
```diff
@@ -129,39 +131,42 @@ func Handle(testParams *TestParameters) error {
...
+			defer func() {
+				if err := setEnvProject(oldProject); err != nil {
+					klog.Errorf("failed to set project environment to %s: %v", oldProject, err)
+				}
+			}()
+		} else {
+			if testParams.ImageRegistry == "" {
+				testParams.ImageRegistry = fmt.Sprintf("gcr.io/%s/gcs-fuse-csi-driver", strings.TrimSpace(testParams.ProjectID))
+			}
+		}
```
amacaskill, line 165: Jetski : 

Issue: Prior to this PR, the ImageRegistry fallback (gcr.io/ProjectID/gcs-fuse-csi-driver) was executed whenever InProw was false. In the new PR, this fallback was moved inside the if testParams.ManageClusterLifecycle block. By default, tests run locally (via run-e2e-local.sh) have manage-cluster-lifecycle set to false. Without the fallback executing, ImageRegistry remains completely empty unless explicitly exported as $REGISTRY.

Recommendation: This fallback logic should still apply whenever Boskos isn't overriding the behavior, regardless of whether you're provisioning a new cluster. Consider moving the default registry assignment out of the ManageClusterLifecycle conditional tree entirely.

test/README.md
amacaskill: There should be a description for E2E_TEST_GKE_CLUSTER_VERSION, I still don't see this . Any flag you can set manually, you should have a description, otherwise just don't allow it to be set manually

test/e2e/run-e2e-local.sh
amacaskill: @yaozile123 Does this make sense to you?

test/README.md
amacaskill: > This creates a cluster with same configuration we use for prow jobs which is set by the flags we have here. For now this will create a regional cluster with 9 nodes https://screenshot.googleplex.com/5V5RBFiePTeRtpq

Can you please explain this somewhere? I don't see it explained anywhere . You can say  if you want the test runner to create a GKE cluster and destroy the GKE cluster for the test. It uses the same default configuration as prow. or something like that

test/README.md
amacaskill: Ok sounds good, I think we will want this doc on open source, but a follow up PR is fine.

test/e2e/utils/handler.go
```diff
@@ -129,39 +131,42 @@ func Handle(testParams *TestParameters) error {
...
+	// For inProw tests, both ManageClusterLifecycle and UseBoskos are set to true.
+	// If we need to manage the cluster lifecycle (create and destroy cluster)
+	if testParams.ManageClusterLifecycle {
+		if testParams.UseBoskos {
+			// 1. Get the old project ID.
+			output, err := gcloudCommand(testParams, "config", "get-value", "project").CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("failed to get gcloud project, output: %v, err: %w", string(output), err)
+			}
+			oldProject := string(output)
```
amacaskill: Don't we also need the workaround from above here? Or why is it not needed? 


```
		// CombinedOutput captures both stdout and stderr, so Cloud Shell may prepend
		// "Your active configuration is: [...]" to the project ID. Take the last line.
		lines := strings.Split(strings.TrimSpace(string(output)), "\n")
		testParams.ProjectID = strings.TrimSpace(lines[len(lines)-1])
	}
	if err := os.Setenv(ProjectEnvVar, testParams.ProjectID); err != nil {
		klog.Fatalf("failed to set %s env var: %v", ProjectEnvVar, err)
	}

```

In my case, I see 

```
gcloud config get-value project
Your active configuration is: [staging]
amacaskill-gke-dev
```


So will this fail to restore our old project if we run 

```
make e2e-test E2E_TEST_USE_BOSKOS=true 
```

test/e2e/utils/handler.go
```diff
@@ -129,39 +131,42 @@ func Handle(testParams *TestParameters) error {
...
-		// 4. After the test, tear down the cluster, and switch back to the old project.
-		defer func() {
-			if err := setEnvProject(oldProject); err != nil {
-				klog.Errorf("failed to set project environment to %s: %v", oldProject, err)
+			if err := setEnvProject(newProject); err != nil {
+				return fmt.Errorf("failed to set project environment to %s: %w", newProject, err)
 			}
-		}()
+			testParams.ProjectID = newProject
+			testParams.ImageRegistry = fmt.Sprintf("gcr.io/%s/gcs-fuse-csi-driver", strings.TrimSpace(newProject))
```
amacaskill: testParams.ImageRegistry = fmt.Sprintf("gcr.io/%s/gcs-fuse-csi-driver", strings.TrimSpace(newProject)) is duplicated in the if and in the else. 

Can you please de-duplicate?

amacaskill (approved):

test/e2e/run-e2e-local.sh
Sneha-at: Some more context from jetski:
```
  ### 2. Why did the gcsfuse-pr-number change not fail for other Prow jobs or local changes?                                                                                                                  
                                                                                                                                                                                                              
  Before PR #1565, the scripts always appended the flag: --gcsfuse-pr-number=${gcsfuse_pr_number}.                                                                                                            
  If GCSFUSE_PR_NUMBER was unset or empty, the shell expanded this to --gcsfuse-pr-number=.                                                                                                                   
                                                                                                                                                                                                              
  This did not fail for other runs because:                                                                                                                                                                   
                                                                                                                                                                                                              
  1. Go Flag Parsing: Go's standard flag package handles empty values (e.g., --flag=) without error. It sets the flag's value to "" (empty string).                                                           
  2. Empty String Handling: The E2E test Go code checks if testParams.GCSFusePRNumber != "" before doing any PR-specific actions (like downloading GCSFuse from source). If it is empty, it safely skips these
  steps.                                                                                                                                                                                                      
  3. Shell Evaluation: The shell eval command handles --gcsfuse-pr-number= as a single valid argument.                                                                                                        
                                                                                                                                                                                                              
  #### Potential reasons why it failed in your specific local run:                                                                                                                                            
                                                                                                                                                                                                              
  • Literal Quotes in Env Var: If GCSFUSE_PR_NUMBER was set to literal quotes (e.g., export GCSFUSE_PR_NUMBER='""'), the command would expand to --gcsfuse-pr-number="". Go would receive "" (with quotes) as 
  the value, causing tests to fail when trying to git clone/fetch using that invalid PR number.                                                                                                               
  • Space in Env Var: If it was set to a space (export GCSFUSE_PR_NUMBER=" "), Go would receive " " as the PR number, which would also bypass the empty check and fail during git operations.                 
                                                                                                                                                                                                              
  The change to ${gcsfuse_pr_number:+--gcsfuse-pr-number=${gcsfuse_pr_number}} is a cleaner approach because it omits the flag entirely when empty, preventing potential misinterpretations by the shell or   
  the Go binary
```

test/e2e/utils/handler.go
Sneha-at: Done, this is fixed with de-duplication mentioned in above comment

test/e2e/utils/handler.go
Sneha-at: Done

test/README.md
Sneha-at: Added this in README

test/e2e/utils/handler.go
Sneha-at: Makes sense, our janitor would clean up any resources from clusters but this will leak clusters in local testing

test/e2e/utils/handler.go
Sneha-at: You are correct, not sure why this wasn't done earlier or how did we not fail here. I have created a separate function to parse the project Name from this output

test/e2e/utils/cluster.go
Sneha-at: Replied in a different comment, this needs passing extra parameters to these functions which doesn't seem necessary for basic gcloud commands

PR #1577, opened 2026-08-03
Skip cleanupSocket on staging paths and remove shared mount metrics s…

yaozile123:
…ymlink

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR did two things:

- **Socket cleanup:** Decoupled `CleanupSocket` from general filesystem unmounting (`Mounter.Unmount` / `Mounter.UnmountWithForce`) and invoked it explicitly in `NodeUnpublishVolume` before unmounting Pod target paths. This prevents `NodeUnstageVolume` (which unstages PV Shared Node Mount staging paths) from ever triggering socket cleanup, avoiding invalid Pod ID parsing and erroneous `_` socket removal attempts.
- **Symlink cleanup:** While standard sidecar mount symlinks continue to be removed by `CleanupSocket(targetPath)` during `NodeUnpublishVolume`, Shared Mount metric collector symlinks are now explicitly removed immediately after calling `UnregisterMetricsCollector` in `NodeUnstageVolume` to prevent leakage without conditional staging path checks.

Fixes #

I have verified locally by deploying a workload pod with shared mount volume and delete it, and Node Unstage Volume went successfully.

```release-note

```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -591,6 +591,10 @@ func (s *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpub
 		s.volumeStateStore.Delete(targetPath)
 	}
 
+	if c, ok := s.mounter.(interface{ CleanupSocket(string) }); ok {
+		c.CleanupSocket(targetPath)
+	}
+
 	// Check if the target path is already mounted
 	if mounted, err := s.isDirMounted(targetPath); mounted || err != nil {
 		if err != nil {
@@ -1208,6 +1212,19 @@ func (s *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstage
 		// It is idempotent to unregister the same collector.
 		if s.driver.config.MetricsManager != nil {
 			s.driver.config.MetricsManager.UnregisterMetricsCollector(stagingPath, s.driver.config.NodeID, vs.MounterPodUID, vs.VolumeName)
+			if s.driver.config.FeatureOptions != nil && s.driver.config.FeatureOptions.SharedMountOptions != nil {
+				fuseSocketDir := s.driver.config.FeatureOptions.SharedMountOptions.FuseSocketDir
+				if vs.MounterPodUID != "" && vs.VolumeName != "" && fuseSocketDir != "" {
+					socketBasePath := util.GetSocketBasePath(vs.MounterPodUID, vs.VolumeName, fuseSocketDir)
+					if err := os.Remove(socketBasePath); err != nil && !os.IsNotExist(err) {
+						klog.Errorf("failed to remove metrics collector symlink %q: %v", socketBasePath, err)
+					}
+				} else {
+					// TODO(yaozile): Implement a long-term solution to prevent leaking metrics collector
+					// symlinks when NodeUnstageVolume is called after a driver restart
+					klog.Errorf("failed to remove metrics collector symlink for staging path %q: empty podUID, volumeName, or fuseSocketDir, symlink leaked", stagingPath)
+				}
+			}
 		}
 
 		// Stop GCSFuse Kernel Params monitoring.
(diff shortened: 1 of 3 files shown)


pkg/csi_mounter/csi_mounter.go
```diff
@@ -348,9 +348,15 @@ func (m *Mounter) createSocket(target string, logPrefix string) (net.Listener, e
 }
 
 func (m *Mounter) cleanupSocket(target string) {
+	if util.IsStagingPath(target) {
```
uriel-guzman: Instead of checking if it's a staging path inside `cleanupSocket`, can we just remove `cleanupSocket` from `Unmount`? https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L251, and then just call `cleanupSocket` right before NodeUnpublishVolume calls `Unmount`? I feel like this is a cleaner way of doing this.

pkg/metrics/metrics.go
```diff
@@ -174,6 +174,15 @@ func (mm *manager) UnregisterMetricsCollector(mountPath, nodeName, podUID, volum
 		mm.volumeMountPathRegistered.Delete(mountPath)
 		klog.Infof("successfully unregistered a metrics collector for mount path %q. There's %d collectors registered on node %q.", mountPath, mm.volumeMountPathRegistered.Len(), nodeName)
 	}
+
+	if util.IsStagingPath(mountPath) {
```
uriel-guzman: Instead of putting this `IsStagingPath` check here, can we just remove the socket immediately after we call `UnregisterMetricsCollector` during `NodeUnstageVolume`? https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_driver/node.go#L1218, I feel like this is also a cleaner way of doing this, reducing conditional logics.

uriel-guzman (changes requested):

pkg/csi_mounter/csi_mounter.go
yaozile123: Done

pkg/metrics/metrics.go
yaozile123: Done

pkg/csi_driver/node.go
```diff
@@ -1216,6 +1212,14 @@ func (s *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstage
 		// It is idempotent to unregister the same collector.
 		if s.driver.config.MetricsManager != nil {
 			s.driver.config.MetricsManager.UnregisterMetricsCollector(stagingPath, s.driver.config.NodeID, vs.MounterPodUID, vs.VolumeName)
+			if s.driver.config.FeatureOptions != nil && s.driver.config.FeatureOptions.SharedMountOptions != nil {
+				socketBasePath := util.GetSocketBasePath(vs.MounterPodUID, vs.VolumeName, s.driver.config.FeatureOptions.SharedMountOptions.FuseSocketDir)
+				if socketBasePath != "" {
```
uriel-guzman: I don't think `socketBasePath` will ever be empty, it will be `"_"`, which happens when podUID and volumeName are both empty: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/82f0a38d97d9b69f37c7aa57f6ee267a81171fd6/pkg/util/util.go#L225.

uriel-guzman: I think we'd also like to log an error if that ever happens, since it means the socketBasePath is leaked.

yaozile123: You are right, thanks for catching that. I updated the code and logged an error there if leak happens

uriel-guzman (approved):

PR #1578, opened 2026-08-03
Restrict private sidecar images in shared mount mode.

uriel-guzman:
**What type of PR is this?**
/kind feature

# Why this is a risk now (compared to gke-gcsfuse-sidecar)
 1. Standard Sidecar (gke-gcsfuse-sidecar) is Securely Jailed:
      - In pkg/webhook/sidecar_spec.go (GetSecurityContext), the standard sidecar container runs under the Restricted Pod Security Standard:
        - RunAsNonRoot: true
        - RunAsUser: NobodyUID (65534)
        - Capabilities.Drop: ["ALL"]
        - ReadOnlyRootFilesystem: true
      - Because it is severely restricted, allowing an arbitrary/custom image is safe: the user can only run custom code within their own isolated namespace boundaries under a restricted
        user profile. They cannot access host devices or other pods' volumes.

   2. Shared Mounter Pod (gcsfusecsi-mount) is Fully Privileged:
      - In pkg/csi_driver/shared_mount.go (createMounterPodSpec), the shared mounter container named gcsfusecsi-mount runs with:
        - Privileged: ptr.To(true)
      - Running a container as privileged gives it full host kernel capabilities, raw access to all host devices (/dev), and makes it trivial to escape to the host node (e.g., mounting
        raw disks, using nsenter, or injecting host cronjobs).
      - If pkg/csi_driver/controller.go (line 223) accepts a custom user-supplied image override from the PVC's annotated PodTemplate, the driver will create this privileged pod with
        that arbitrary image.

# How to addresss the security concern
We should implement a balanced option to eliminate the privilege escalation risk while preserving necessary flexibility:

   1. pkg/csi_driver/gcs_fuse_driver.go:
     Added AllowCustomMounterImages bool field to the SharedMountOptions struct to pass the configuration parameter into the CSI controller.
   2. cmd/csi_driver/main.go:
     Introduced a new flag --allow-custom-mounter-images (defaulting to false) to control whether untrusted images are permitted.
   3. pkg/csi_driver/controller.go:
     Added strict validation for the mounter container image override. If a custom image is provided:
      * It is allowed if it is a GKE-managed image (verified using isManagedSidecarImage). This allows users to quickly pin/adopt official GKE fixes or mitigations before a full driver
        rollout.
      * It is allowed if the cluster administrator explicitly opted-in by setting --allow-custom-mounter-images=true.
      * Otherwise, the publish request is rejected with codes.InvalidArgument to prevent execution of arbitrary untrusted binaries as root.
   4. pkg/csi_driver/controller_test.go:
      * Update the existing image override test case to include allowCustomMounterImages: true so that advanced/admin override scenarios continue to pass.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -226,7 +226,12 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
 		if container.Image != "" && container.Image != mounterPodManagedImageKeyword {
 			// If the image isn't the placeholder managed keyword, the user is trying to
 			// override the mounter pod's image.
-			// TODO(urielguzman): Somehow validate image so that only trustworthy repositories are allowed.
+			// Validate that the image is trustworthy (either it is an official GKE-managed image,
+			// or the cluster administrator has explicitly allowed custom mounter images).
+			imageOverrideAllowed := s.features.SharedMountOptions.AllowCustomMounterImages || isStrictManagedSidecarImage(container.Image)
+			if !imageOverrideAllowed {
+				return nil, status.Errorf(codes.InvalidArgument, "image %q is not allowed. Only GKE managed images are allowed", container.Image)
+			}
 			containerImage = container.Image
 		}
 		break
--- a/cmd/csi_driver/main.go
+++ b/cmd/csi_driver/main.go
@@ -81,7 +81,8 @@ var (
 	enableGCSFuseKernelParams = flag.Bool("enable-gcsfuse-kernel-params", false, "Enable gcsfuse kernel params feature.")
 
 	// GCSFuse shared mount flags.
-	enableSharedMount = flag.Bool("enable-shared-mount", false, "Enable the shared mount feature.")
+	enableSharedMount        = flag.Bool("enable-shared-mount", false, "Enable the shared mount feature.")
+	allowCustomMounterImages = flag.Bool("allow-custom-mounter-images", false, "Allow arbitrary non-GKE-managed images to be used for mounter pods.")
 
 	// GCSFuse profiles flags.
 	enableGCSFuseProfiles         = flag.Bool("enable-gcsfuse-profiles", false, "Enable the gcsfuse profiles feature.")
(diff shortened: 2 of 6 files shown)


pkg/csi_driver/controller.go
```diff
@@ -226,7 +226,12 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
 		if container.Image != "" && container.Image != mounterPodManagedImageKeyword {
 			// If the image isn't the placeholder managed keyword, the user is trying to
 			// override the mounter pod's image.
-			// TODO(urielguzman): Somehow validate image so that only trustworthy repositories are allowed.
+			// Validate that the image is trustworthy (either it is an official GKE-managed image,
+			// or the cluster administrator has explicitly allowed custom mounter images).
+			imageOverrideAllowed := s.features.SharedMountOptions.AllowCustomMounterImages || isStrictManagedSidecarImage(container.Image)
+			if !imageOverrideAllowed {
+				return nil, status.Errorf(codes.InvalidArgument, "image %q is not allowed. Only GKE managed images are allowed", container.Image)
```
chrisThePattyEater, line 233: Can we make this more actionable to users by mentioning the new flag maybe - "image %q is not allowed. Only GKE managed images are allowed unless --allow-custom-mounter-images is enabled."

pkg/csi_driver/controller.go
```diff
@@ -226,7 +226,12 @@ func (s *controllerServer) ControllerPublishVolume(ctx context.Context, req *csi
 		if container.Image != "" && container.Image != mounterPodManagedImageKeyword {
 			// If the image isn't the placeholder managed keyword, the user is trying to
 			// override the mounter pod's image.
-			// TODO(urielguzman): Somehow validate image so that only trustworthy repositories are allowed.
+			// Validate that the image is trustworthy (either it is an official GKE-managed image,
+			// or the cluster administrator has explicitly allowed custom mounter images).
+			imageOverrideAllowed := s.features.SharedMountOptions.AllowCustomMounterImages || isStrictManagedSidecarImage(container.Image)
```
yaozile123, line 231: Why do we use `isStrictManagedSidecarImage` at here but use `isManagedSidecarImage` elsewhere?

pkg/csi_driver/controller.go
uriel-guzman: I don't think we should emit this message in the error, since the flag can't even be actionized by the user, since it's a CSI Driver flag. The customer might be confused, thinking they are missing some sort of mount option / volume attribute. If a cluster administrator wants to turn the flag on, they could look at the codebase instead, and turn the flag on in node.yaml.

pkg/csi_driver/controller.go
uriel-guzman: `isManagedSidecarImage` allows an attacker to bypass the regex with something like: `docker.io/attacker/gke.gcr.io/gcs-fuse-csi-driver-sidecar-mounter:v1.23.0-gke.1`. `isStrictManagedSidecarImage` is much more strict. You can look at the unit tests to check which things are not allowed anymore. I didn't want to update `isManagedSidecarImage` to make it more strict, since customers could be broken on upgrades if they were relying on the previous regex and fail to match the new one.

yaozile123 (approved):

PR #1579, opened 2026-08-03
[Shared Mount] Add User Allowed  o= MO

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

In the sidecar we have a list of allowed flags which users can send to the linux kernel. This pr moves shared node mount to be in sync by performing the same filtering on user provided mo.

  1. Extracted Shared Helpers into pkg/util/util.go:
      • ProcessAllowedMountOptions: Filters and extracts FUSE -o options (like exec, rw, ro, allow_other), keeping non-o=
      flags in order.
  2. Unified Option Pipeline:
      • Updated prepareMountOptions (csi_mounter.go) and prepareSharedNodeMountOptions (utils.go) to use these shared    
      helpers instead of duplicate inline loops.
  3. Live Cluster Verification:
      • Verified on GKE that mountOptions: "o=ro" strictly enforces read-only access for root and non-root users (Read-  
      only file system on writes).
      • Verified that passing multiple o= options (e.g. o=rw,o=allow_other) properly concatenates into gcsfuse -o rw,    
      allow_other.
      
      
Fixes #

```release-note
NONE
```


pkg/util/util.go
```diff
@@ -393,3 +393,26 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+			switch v {
+			case "exec", "noexec", "atime", "noatime", "sync", "async", "dirsync", "rw", "ro":
+				allowed = append(allowed, v)
+			default:
+				klog.Warningf("got invalid mount option %q. Will discard invalid options and continue to mount.", v)
+			}
+			continue
+		}
+		remainingOptions = append(remainingOptions, o)
+	}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is intended

pkg/util/util.go
```diff
@@ -393,3 +393,26 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+func ProcessAllowedMountOptions(
+	options []string,
+) ([]string, []string) {
+	allowed := make([]string, 0, len(options))
+	remainingOptions := make([]string, 0, len(options))
+	for _, o := range options {
+		if strings.HasPrefix(o, "o=") {
+			v := o[2:]
+			switch v {
+			case "exec", "noexec", "atime", "noatime", "sync", "async", "dirsync", "rw", "ro":
```
yaozile123: Did you test with mount options other than `rw` and `ro`?
In [validMountOptions](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/84acf4d8c7b6ca90e33f172585511fb772196204/pkg/csi_mounter/csi_mounter.go#L408-L417), it only extracts `rw` and `ro` into `csiMountOptions`. Other options (e.g., `noexec`, `atime`) are missed and get passed to `gcsfuse` as CLI flags, which might cause gcsfuse crash.

uriel-guzman: Why was this transformed into a switch statement? Can't we just keep the allowedOptions map (and the rest of the code and just copy paste it into a helper function) we already had? What does this improve?

uriel-guzman (changes requested):

pkg/csi_driver/utils.go
```diff
@@ -800,10 +800,11 @@ func getInternalMountOptionValue(options []string, key string) string {
 // prepareSharedNodeMountOptions processes and filters mount options to be directly passed to the gcsfuse process during shared node mounts.
 // Unlike prepareMountOptions (which splits options between those used to open fuse device and those passed to gcsfuse).
 func prepareSharedNodeMountOptions(options []string) []string {
-	// TODO(FUECHR): Investigate if we should only allow "allowedOptions" from prepareMountOptions for shared node mount.
-
 	csiMountOptions := []string{"o=allow_other,default_permissions"}
-	for _, o := range options {
+
+	allowed, remainingOptions := util.ProcessAllowedMountOptions(options)
+	csiMountOptions = append(csiMountOptions, allowed...)
```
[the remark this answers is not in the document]
uriel-guzman: +1 how did this even pass manual verification?

uriel-guzman (changes requested):

pkg/util/util.go
chrisThePattyEater: It was a recommendation from gemini that was a micro enhancement. Seemed innocent, but I can revert to keep same structure as before.

pkg/util/util.go
```diff
@@ -393,3 +405,25 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+				allowed = append(allowed, v)
+			} else {
+				klog.Warningf("got invalid mount option %q. Will discard invalid options and continue to mount.", v)
+			}
+			continue
+		}
+		remainingOptions = append(remainingOptions, o)
+	}
+	return allowed, remainingOptions
+}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is expected since its the existing sidecars logic

pkg/util/util.go
```diff
@@ -393,3 +405,25 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+	for _, o := range options {
+		if strings.HasPrefix(o, "o=") {
+			v := o[2:]
+			if allowedMountOptions[v] {
+				allowed = append(allowed, v)
+			} else {
+				klog.Warningf("got invalid mount option %q. Will discard invalid options and continue to mount.", v)
+			}
+			continue
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: this is not going to be supported to stay in sync with sc

pkg/csi_driver/utils_test.go
```diff
@@ -1325,6 +1325,11 @@ func TestPrepareSharedNodeMountOptions(t *testing.T) {
 			options:  []string{"o=rw", "implicit-dirs"},
 			expected: append(append([]string{}, defaultOpts...), "o=rw", "implicit-dirs"),
 		},
+		{
+			name:     "mixed options with existing o= flags",
+			options:  []string{"o=noexec", "rw", "ro", "enable-cloud-profiler=true"},
+			expected: append(append([]string{}, defaultOpts...), "o=noexec", "rw", "ro", "enable-cloud-profiler=true"),
+		},
```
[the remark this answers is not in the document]
chrisThePattyEater: na

pkg/util/util.go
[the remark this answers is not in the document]
chrisThePattyEater: this is expected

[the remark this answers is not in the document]
chrisThePattyEater: the point being to stay in sync with sidecar

pkg/util/util.go
```diff
@@ -91,6 +91,18 @@ var (
...
+
+	allowedMountOptions = map[string]bool{
+		"exec":    true,
+		"noexec":  true,
+		"atime":   true,
+		"noatime": true,
+		"sync":    true,
+		"async":   true,
+		"dirsync": true,
+		"rw":      true,
```
uriel-guzman: `rw` and `ro` were not previously in this map. You added them in this PR. This used to mean that `o=rw` and `o=ro` were disallowed in sidecar architecture mode. Since you added these flags here for some reason, now `o=rw` and `o=ro` will be allowed in sidecar architecture mode, which I don't think we should modify since it's a user facing change that is unrelated to shared mount.

uriel-guzman (changes requested):

pkg/csi_driver/utils.go
```diff
@@ -808,10 +800,17 @@ func getInternalMountOptionValue(options []string, key string) string {
...
-	// TODO(FUECHR): Investigate if we should only allow "allowedOptions" from prepareMountOptions for shared node mount.
-
 	csiMountOptions := []string{"o=allow_other,default_permissions"}
-	for _, o := range options {
+
+	allowed, remainingOptions := util.ProcessAllowedMountOptions(options)
+	for _, opt := range allowed {
+		csiMountOptions = append(csiMountOptions, "o="+opt)
+	}
+	for _, o := range remainingOptions {
```
uriel-guzman: `prepareMountOptions` used to oput the mount options into a set: `optionSet := sets.NewString(remainingOptions...)`, but `prepareSharedMountOptions` is not, which means that it will allow duplicate mount options, while sidecar architecture doesn't, which is inconsistent behavior.

I don't actually understand why the implementation is so different in both functions, if the only difference is the default `csiMountOptions`? Or am I missing something here? Can't you just make `prepareMountOptions` accept a new parameter for `defaultMountOptions`, which are passed differently in sidecar architecture and in shared mode architecture, and then the only different is that for shared mount, the caller of `prepareMountOptions` prepends the `csiMountOptions` with the `-o` flag, and then merges them together?

uriel-guzman (changes requested):

pkg/util/util.go
```diff
@@ -393,3 +403,102 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+			if nodeFuseMaxRequestLimitKB > maxAllowedSizeKB {
+				return nil, nil, nil, 0, fmt.Errorf("invalid value for %s mount flag %q: exceeds maximum allowed limit of %d KiB on this host", nodeFuseMaxRequestLimitKBMountFlag, o, maxAllowedSizeKB)
+			}
+			// Convert the requested size in KB to pages. We use integer ceiling division
+			// to round up to the next page if the requested size is not a multiple of the
+			// page size. This ensures we can accommodate the full requested buffer size in
+			// a single fuse request.
+			fuseMaxPagesLimit = CeilDiv64(nodeFuseMaxRequestLimitKB, pageSizeKB)
+			continue
+		}
```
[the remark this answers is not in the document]
chrisThePattyEater: lets leave this alone since its pre existing code from sidecar

pkg/util/util.go
```diff
@@ -91,6 +92,15 @@ var (
...
 	)
 	procMountsPath = "/proc/mounts"
+	allowedOptions = map[string]bool{
+		"exec":    true,
+		"noexec":  true,
+		"atime":   true,
+		"noatime": true,
+		"sync":    true,
+		"async":   true,
+		"dirsync": true,
```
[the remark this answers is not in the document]
chrisThePattyEater: this is existing logic, the sidecar currently only takes "rw" and "ro".

pkg/util/util.go
```diff
@@ -393,3 +403,102 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+	// maxFuseMaxPagesLimit defines the maximum pages limit supported by the Linux FUSE kernel (2^16 - 1).
+	maxFuseMaxPagesLimit = 65535
+)
+
+var (
+	readAheadKBMountFlagRegex               = regexp.MustCompile(readAheadKBMountFlagRegexPattern)
+	nodeFuseMaxRequestLimitKBMountFlagRegex = regexp.MustCompile(nodeFuseMaxRequestLimitKBMountFlagRegexPattern)
+)
+
+func PrepareMountOptions(options []string, csiMountOptions []string) ([]string, []string, map[string]int64, int64, error) {
```
uriel-guzman: Why did you have to move all of this to util.go? If you had left it in csi_mounter.go, you wouldn't have needed to migrate all of this stuff here, including unrelated stuff like the consts above?

uriel-guzman: I can't seem to figure out why you moved everything to util.go, if you're just exporting the function `PrepareMountOptions` and using it in csi_mounter.go anyway?

pkg/csi_mounter/csi_mounter.go
```diff
@@ -37,29 +35,13 @@ import (
 	sidecarmounter "github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/sidecar_mounter"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/webhook"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 )
 
 const (
 	// Note: All variables here are in KiB instead of KB but they are added like this to ensure consistency in codebase.
```
uriel-guzman: You left an orphaned comment here.

pkg/util/util.go
```diff
@@ -393,3 +403,102 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+	pageSizeKB := max(minPageSizeKB, int64(os.Getpagesize()/KiB))
+	// Calculate default FUSE max pages limit. We use integer ceiling division to ensure we round
+	// up to the next page if the default size 16MB is not a perfect multiple of the page size on the machine.
+	fuseMaxPagesLimit := CeilDiv64(defaultNodeFuseMaxRequestLimitKB, pageSizeKB)
+
+	for _, o := range options {
+		if strings.HasPrefix(o, "o=") {
+			v := o[2:]
+			if allowedOptions[v] {
+				csiMountOptions = append(csiMountOptions, v)
```
uriel-guzman, line 483: This logic just blindly removed the optionSet, meaning now there will be duplicate mount options? https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/main/pkg/csi_mounter/csi_mounter.go#L447.

Can we just not add any logical changes here? This is also shared by sidecar architecture, so I don't really understand why you're adding new implementation details that change how the logic works?

chrisThePattyEater: I have updated the code to copy and paste the existing function

pkg/csi_mounter/csi_mounter.go
chrisThePattyEater: removed

pkg/util/util.go
```diff
@@ -393,3 +403,102 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+			"default_permissions",
+			"rootmode=40000",
+			fmt.Sprintf("user_id=%d", os.Getuid()),
+			fmt.Sprintf("group_id=%d", os.Getgid()),
+		}
+	} else {
+		csiMountOptions = append([]string(nil), csiMountOptions...)
+	}
+
+	var remainingOptions []string
```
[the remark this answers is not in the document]
chrisThePattyEater: this is a micro fix resolving

pkg/csi_mounter/csi_mounter.go
uriel-guzman: I didn't say to remove it... I meant you copied it to util.go, since you moved the consts from csi_mounter.go to util.go, but just left the comment there without any context, it had important context about the consts, so please add it back, in util.go, on top of the consts.

pkg/util/util.go
```diff
@@ -393,3 +404,92 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+	// maxFuseMaxPagesLimit defines the maximum pages limit supported by the Linux FUSE kernel (2^16 - 1).
+	maxFuseMaxPagesLimit = 65535
+)
+
+var (
+	readAheadKBMountFlagRegex               = regexp.MustCompile(readAheadKBMountFlagRegexPattern)
+	nodeFuseMaxRequestLimitKBMountFlagRegex = regexp.MustCompile(nodeFuseMaxRequestLimitKBMountFlagRegexPattern)
+)
+
+func PrepareMountOptions(options []string, csiMountOptions []string) ([]string, []string, map[string]int64, int64, error) {
```
uriel-guzman: Exported functions should have comments. Please add a comment at the top (copy and paste to Gemini and paste whatever it returns on the top).

pkg/util/util_test.go
```diff
@@ -781,3 +791,189 @@ func TestWaitForPathMounted(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			t.Logf("test case: %s", tc.name)
+
+			initialOpts := tc.initialCsiMountOptions
+			if initialOpts == nil {
+				initialOpts = defaultCsiMountOptions
+			}
+
```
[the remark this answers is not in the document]
uriel-guzman: Address comment.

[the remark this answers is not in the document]
chrisThePattyEater: fixed

pkg/util/util.go
chrisThePattyEater: added

pkg/util/util.go
```diff
@@ -393,3 +404,96 @@ func WaitForPathMounted(ctx context.Context, path string) error {
 func CeilDiv64(a, b int64) int64 {
 	return (a + b - 1) / b
 }
+
+const (
```
uriel-guzman: You can't just copy and paste consts from one file to another and leave them in the middle of the file. Consts should go on the top of the file, and there's actually alraedy a const block declared at the top., and a var block as well. Please put them properly.

chrisThePattyEater: moved them to be at the top

pkg/csi_driver/node.go
```diff
@@ -1126,12 +1126,21 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	}
 
 	// Prepare mount options specifically for shared node mount architecture before sending to mounter pod.
-	preparedMountOptions := prepareSharedNodeMountOptions(args.fuseMountOptions)
+	defaultCSIMountOptions := []string{"allow_other,default_permissions"}
```
uriel-guzman: To improve readability, create a helper method in util.go called `PrepareSharedMountOptions` that returns `gcsfuseMO, sysfsBDI, fuseMaxPagesLimit, err` andcopy / paste the code from L1129 to L1139 so that you can just call `gcsfuseMO, _, _, err := util.PrepareSharedMountOptions` (and handle the error).

Similarly, in util.go, create another helper method called `PrepareSidecarMountOptions` and put:
```
	defaultOptions := []string{
		"nodev",
		"nosuid",
		"allow_other",
		"default_permissions",
		"rootmode=40000",
		fmt.Sprintf("user_id=%d", os.Getuid()),
		fmt.Sprintf("group_id=%d", os.Getgid()),
	}
```
inside the helper function, so that in csi_mounter.go you just call `util.PrepareSidecarMountOptions`.

Also, you can unexport `PrepareMountOptions` -> `prepareMountOptions` in util.go since now it's just called by your two helper functions `PrepareSharedMountOptions` and `PrepareSidecarMountOptions`.

This helps moving `defaultOptions` in csi_mounter.go and `defaultCSIMountOptions` in node.go to be both in util.go, where they should be (together with allowedMountOptions), and improves readability since you just call one single function with a clear and descriptive name / purpose. It also captures exactly what the difference is between both implementations.

uriel-guzman (changes requested):

pkg/csi_driver/node.go
chrisThePattyEater: fixed in latest push https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/compare/ce375a7b449970209c9ff61154e3e8ca4e909a02..79c7b9b02d5a91da1ef1c95ac8b371a2e9f6e2a1

pkg/util/util_test.go
```diff
@@ -781,3 +791,191 @@ func TestWaitForPathMounted(t *testing.T) {
...
+	for _, o := range options {
+		dict[o]++
+	}
+
+	return dict
+}
+
+func ptr[T any](v T) *T {
+	return &v
+}
```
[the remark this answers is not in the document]
uriel-guzman: Please add this unit test.

pkg/util/util.go
```diff
@@ -393,3 +420,107 @@ func WaitForPathMounted(ctx context.Context, path string) error {
...
+	linuxMO, sidecarMountOptions, sysfsBDI, fuseMaxPagesLimit, err := prepareMountOptions(options, defaultCSIMountOptions)
+	if err != nil {
+		return nil, nil, fuseMaxPagesLimit, err
+	}
+	for i, opt := range linuxMO {
+		linuxMO[i] = "o=" + opt
+	}
+	// For shared node mounts we pass both linux mo and sidecar mo, with linux mo being prefixed with "o=" to be passed to gcsfuse.
+	gcsfuseMO := append(linuxMO, sidecarMountOptions...)
+	return gcsfuseMO, sysfsBDI, fuseMaxPagesLimit, nil
```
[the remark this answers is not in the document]
uriel-guzman: Ignore.

uriel-guzman (approved):

PR #1580, opened 2026-08-04
Check container termination and OOM for mounter pods

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

In shared mount mode, the CSI driver only checked the `/error` file and `pod.Status.Phase`.
When a mounter container terminated because of an OOM event:
- The container did not write an error to the `/error` file.
- The driver returned `codes.Internal`.
- This behavior triggered false SLO alerts.
- When the container restarted, `pod.Status.Phase` remained `PodRunning`. The driver did not detect the terminated GCSFuse process.

To solve this issue, this PR did the following things:
- Refactored GCSFuse container inspection into `checkContainerStatusErr` for shared use.
- Added container status inspection to `waitForMounterServer`, `executeNodeStageVolume`, and `NodePublishVolumeForSharedMount`.
- Returns `codes.ResourceExhausted` when the container terminates with `OOMKilled` or exit code 137.
- Returns `codes.Internal` with the exit code and reason when the container terminates with a non-zero exit code or restarts.

Fixes #

```release-note

```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -173,6 +173,15 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Error(code, err.Error())
 	}
 
+	cs, err := getMounterPodContainerStatus(mounterPod)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	code, err = checkContainerStatusErr(cs)
+	if code != codes.OK {
+		return nil, status.Error(code, err.Error())
+	}
+
 	// We ignore the pod UID parsed from the target path, since we register the mounter pod's UID instead.
 	_, volumeName, err := util.ParsePodIDVolumeFromTargetpath(targetPath)
 	if err != nil {
@@ -461,7 +470,11 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 		}
 
 		// Check if there is any error from the sidecar container
-		code, err = checkSidecarContainerErr(isInitContainer, pod)
+		cs, err := getSidecarContainerStatus(isInitContainer, pod)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		code, err = checkContainerStatusErr(cs)
 		if code != codes.OK {
 			return nil, status.Error(code, err.Error())
 		}
@@ -1120,8 +1133,25 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 	// Send GRPC to mounter pod to start GCSFuse.
 	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, preparedMountOptions); err != nil {
 		klog.Errorf("Failed to mount volume %q to staging path %q: %v", volumeID, stagingPath, err)
-		if code, err := checkMounterPodErrorFile(emptyDirBasePath); err != nil {
-			return nil, status.Error(code, err.Error())
+		if code, fileErr := checkMounterPodErrorFile(emptyDirBasePath); fileErr != nil {
+			return nil, status.Error(code, fileErr.Error())
+		}
+		cs, csErr := func() (*corev1.ContainerStatus, error) {
+			// Re-fetch the mounter pod to get the updated container termination status
+			// (e.g. OOMKilled or crash) that occurred during mountToNode.
+			mounterPod, getErr := s.k8sClients.GetPod(podNamespace, podName)
+			if getErr != nil {
+				return nil, fmt.Errorf("failed to get mounter pod %s/%s: %w", podNamespace, podName, getErr)
+			}
+			if mounterPod == nil {
+				return nil, fmt.Errorf("mounter pod %s/%s cannot be nil", podNamespace, podName)
+			}
+			return getMounterPodContainerStatus(mounterPod)
+		}()
+		if csErr != nil {
+			klog.Warningf("Failed to inspect mounter pod container status: %v", csErr)
+		} else if code, cErr := checkContainerStatusErr(cs); code != codes.OK {
+			return nil, status.Error(code, cErr.Error())
 		}
 		return nil, err
 	}
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -605,13 +605,12 @@ func extractErrorFromGcsFuseErrorFile(errMsg []byte) (codes.Code, error) {
 	return codes.OK, nil
 }
 
-func checkSidecarContainerErr(isInitContainer bool, pod *corev1.Pod) (codes.Code, error) {
-	code := codes.Internal
-	cs, err := getSidecarContainerStatus(isInitContainer, pod)
-	if err != nil {
-		return code, err
+func checkContainerStatusErr(cs *corev1.ContainerStatus) (codes.Code, error) {
+	if cs == nil {
+		return codes.Internal, errors.New("container status is nil")
 	}
 
+	code := codes.Internal
 	var reason string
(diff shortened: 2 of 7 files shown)


pkg/csi_driver/utils.go
```diff
@@ -627,12 +626,24 @@ func checkSidecarContainerErr(isInitContainer bool, pod *corev1.Pod) (codes.Code
...
+	if pod == nil {
+		return nil, errors.New("pod is nil")
+	}
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == util.MounterPodNamePrefix {
+			return &cs, nil
+		}
+	}
+	return nil, errors.New("the mounter pod container was not found")
+}
```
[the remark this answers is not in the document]
yaozile123: Done, I updated `getSidecarContainerStatus` as well

pkg/csi_driver/utils.go
```diff
@@ -627,13 +626,28 @@ func checkSidecarContainerErr(isInitContainer bool, pod *corev1.Pod) (codes.Code
 			code = codes.ResourceExhausted
 		}
 
-		return code, fmt.Errorf("the sidecar container terminated due to %v, exit code: %v", reason, exitCode)
+		return code, fmt.Errorf("the container terminated due to %v, exit code: %v", reason, exitCode)
```
uriel-guzman: After the refactoring, it might not be clear to the customer which container terminated (they could think it was their own application container). I think we need to pass the name of the container through a parameter and log that, e.g. the "gke-gcsfuse-sidecar" or "gcsfusecsi-mount" container terminated due to ...

pkg/csi_driver/node.go
```diff
@@ -1123,6 +1136,13 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
 		if code, err := checkMounterPodErrorFile(emptyDirBasePath); err != nil {
 			return nil, status.Error(code, err.Error())
 		}
+		if mounterPod, getErr := s.k8sClients.GetPod(podNamespace, podName); getErr == nil && mounterPod != nil {
```
uriel-guzman: I'm not sure why we need to call `GetPod` again, if it's already in L1011.

pkg/csi_driver/utils.go
uriel-guzman: You can also consider not passing it as a parameter, but instead, the caller can add the name of the right container using klog.Errorf in node.go.

pkg/csi_driver/node.go
uriel-guzman: getErr is ignored. Consider wrapping all of these errors with `errors.Join` into a single, unified, error. We just need to be careful about how this is going to affect the SLO error categorization.

uriel-guzman (changes requested):

pkg/csi_driver/node.go
```diff
@@ -173,6 +173,15 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Error(code, err.Error())
 	}
 
+	cs, err := getMounterPodContainerStatus(mounterPod)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	code, err = checkContainerStatusErr(cs)
```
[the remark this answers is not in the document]
yaozile123: `checkContainerStatusErr` did not have field for isSidecar

pkg/csi_driver/node.go
yaozile123: The `pod` object at L1011 is fetched before `waitForMounterServer` and `mountToNode` execute. If the mounter container crashes or gets OOMKilled during `mountToNode`, the L1011 snapshot does not have the updated `ContainerStatuses`.

pkg/csi_driver/utils.go
yaozile123: `corev1.ContainerStatus` already includes `cs.Name`. I updated `checkContainerStatusErr` to use `cs.Name` directly in the error string, so we don't need to add a new parameter or change caller signatures.

pkg/csi_driver/node.go
yaozile123: I noticed that`errors.Join` returns a `*joinError` which [does not implement GRPCStatus()](https://github.com/golang/go/blob/5d7583fe1f947d3746dc9b66da38b62ba9902759/src/errors/join.go#L29-L44) which would causing `grpc-go` to [fall back](https://github.com/grpc/grpc-go/blob/fa603ec26e697cf94db719db21a38d7dde24c033/status/status.go#L96-L127) to `codes.Unknown`.

Since `getErr` is only a failure of the diagnostic lookup, should we log `getErr`/`csErr` with `klog.Warningf` instead?

uriel-guzman: Ah, good catch. This is fine, since the we're just getting the pod from the informer's memory cache. Can you just add a comment explain that? Someone might want to optimize it later, causing a bug.

yaozile123: Done, I have added the comment

pkg/csi_driver/node.go
```diff
@@ -1120,8 +1133,21 @@ func (s *nodeServer) executeNodeStageVolume(ctx context.Context, req *csi.NodeSt
...
-			return nil, status.Error(code, err.Error())
+		if code, fileErr := checkMounterPodErrorFile(emptyDirBasePath); fileErr != nil {
+			return nil, status.Error(code, fileErr.Error())
+		}
+		// Re-fetch the mounter pod to get the updated container termination status
+		// (e.g. OOMKilled or crash) that occurred during mountToNode.
+		mounterPod, getErr := s.k8sClients.GetPod(podNamespace, podName)
+		if getErr != nil {
+			klog.Warningf("Failed to get mounter pod %s/%s to check container status: %v", podNamespace, podName, getErr)
+		} else if mounterPod != nil {
```
uriel-guzman: These if / else nestings can be eliminated by returning early and using a function to improve readability:
```
	if err := s.mountToNode(ctx, podUID, stagingPath, volumeID, preparedMountOptions); err != nil {
		klog.Errorf("Failed to mount volume %q to staging path %q: %v", volumeID, stagingPath, err)
		if code, fileErr := checkMounterPodErrorFile(emptyDirBasePath); fileErr != nil {
			return nil, status.Error(code, fileErr.Error())
		}
        cs, csErr := func() (*ContainerStatus, error) {
           // Re-fetch the mounter pod to get the updated container termination status
		  // (e.g. OOMKilled or crash) that occurred during mountToNode.
		  mounterPod, getErr := s.k8sClients.GetPod(podNamespace, podName)
		  if getErr != nil {
			 return nil, fmt.Errorf("failed to get mounter pod %s/%s to check container status: %v", podNamespace, podName, getErr)
		  } 
          if mounterPod == nil {
             return nil, fmt.Errorf("mounter pod %s/%s can't be nil", podNamespace, podName)
          }
	      if cs, csErr := getMounterPodContainerStatus(mounterPod); err != nil {
	        return nil, fmt.Errorf("failed to get mounter pod container status for %s/%s: %v", podNamespace, podName, csErr)
	      } 
        }(); 
        if csErr != nil {
          klog.Warningf(csErr.Error())
        }
       if code, cErr := checkContainerStatusErr(cs); code != codes.OK {
	     return nil, status.Error(code, cErr.Error())
	   }
	   return nil, err
	}

yaozile123: Done, I have updated to use an anonymous function and kept the `else if` on `checkContainerStatusErr` so that if `csErr != nil` (and `cs` is nil), it falls back to logging the warning and returning the original mount error rather than failing with "container status is nil".

uriel-guzman (approved):

PR #1582, opened 2026-08-06
Verify fsGroup and SA name in NPV for shared mount

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

During concurrent resource creation, the mutating webhook skips validation if the referenced PodTemplate is not yet found, allowing workload pods with mismatched permissions to bypass webhook checks. To prevent unauthorized shared mount access, this change enforces defense-in-depth ServiceAccount and FSGroup validation directly in `NodePublishVolumeForSharedMount` before performing the bind-mount. Any mismatch between the workload and the target mounter Pod is rejected with `PermissionDenied`.

### Changes
- **`pkg/csi_driver/controller.go`**: Propagate `PodTemplate.Template.Spec.SecurityContext` to the mounter Pod in `ControllerPublishVolume`.
- **`pkg/csi_driver/shared_mount.go`**: Add `securityContext` to `mounterPodConfig`, populate `spec.Spec.SecurityContext` in `createMounterPodSpec`, and add `validateFSGroup` helper.
- **`pkg/csi_driver/node.go`**: Validate workload pod's SA name and fsGroup against the mounter Pod in `NodePublishVolumeForSharedMount`.
- **`pkg/cloud_provider/clientset/clientset.go`**: Retain `SecurityContext` and `ServiceAccountName` in `ConfigurePodLister` transform so the node-local informer cache preserves mounter pod security settings.

Fixes #

```release-note

```

### DIFF
--- a/pkg/cloud_provider/clientset/fake.go
+++ b/pkg/cloud_provider/clientset/fake.go
@@ -49,6 +49,8 @@ type FakePodConfig struct {
 	PodStatus          *corev1.PodStatus
 	NodeName           string
 	IsMounterPod       bool
+	SecurityContext    *corev1.PodSecurityContext
+	ServiceAccountName string
 }
 
 type FakePVConfig struct {
@@ -209,6 +211,14 @@ func (c *FakeClientset) CreatePod(podConfig FakePodConfig) {
 
 	if podConfig.NodeName != "" {
 		c.fakePod.Spec.NodeName = podConfig.NodeName
+	}
+
+	if podConfig.SecurityContext != nil {
+		c.fakePod.Spec.SecurityContext = podConfig.SecurityContext
+	}
+
+	if podConfig.ServiceAccountName != "" {
+		c.fakePod.Spec.ServiceAccountName = podConfig.ServiceAccountName
 	}
 
 	c.fakePods = append(c.fakePods, c.fakePod)
--- a/pkg/csi_driver/controller_test.go
+++ b/pkg/csi_driver/controller_test.go
@@ -38,17 +38,19 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/fake"
 	k8stesting "k8s.io/client-go/testing"
+	"k8s.io/utils/ptr"
 )
 
 const (
-	testVolumeID           = "test-volume-id"
-	testNodeID             = "test-node-id"
-	testPV                 = "test-pv"
-	testPVC                = "test-pvc"
-	testMounterPodTemplate = "test-mounter-pod-template"
-	testNamespace          = "test-namespace"
-	testPod                = "test-pod"
-	testImage              = "k8s.gcr.io/pause"
+	testVolumeID                 = "test-volume-id"
+	testNodeID                   = "test-node-id"
+	testPV                       = "test-pv"
+	testPVC                      = "test-pvc"
+	testMounterPodTemplate       = "test-mounter-pod-template"
+	testNamespace                = "test-namespace"
+	testPod                      = "test-pod"
+	testImage                    = "k8s.gcr.io/pause"
+	testFSGroup            int64 = 1000
 )
 
 func initTestController(t *testing.T, clientset clientset.Interface) csi.ControllerServer {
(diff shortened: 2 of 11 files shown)


pkg/csi_driver/node.go
```diff
@@ -159,6 +159,18 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
...
 	}
 
+	// Verify workload service account name matches mounter pod.
+	vc := req.GetVolumeContext()
+	if vc[VolumeContextKeyServiceAccountName] != mounterPod.Spec.ServiceAccountName {
+		return nil, status.Errorf(codes.PermissionDenied, "workload serviceAccountName %q does not match mounter serviceAccountName %q", vc[VolumeContextKeyServiceAccountName], mounterPod.Spec.ServiceAccountName)
+	}
+
+	// Verify workload fsGroup matches mounter pod.
+	workloadFSGroup := req.GetVolumeCapability().GetMount().GetVolumeMountGroup()
```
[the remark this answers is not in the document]
yaozile123: `GetVolumeMountGroup()` already had defensive `nil` check 
https://github.com/container-storage-interface/spec/blob/v1.11.0/lib/go/csi/csi.pb.go#L6485-L6490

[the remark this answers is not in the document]
yaozile123: `GetVolumeMountGroup()` already had defensive `nil` check 
https://github.com/container-storage-interface/spec/blob/v1.11.0/lib/go/csi/csi.pb.go#L6485-L6490

pkg/csi_driver/node.go
```diff
@@ -159,6 +159,18 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
 		return nil, status.Errorf(codes.Internal, "mounter pod %s/%s cannot be nil", mounterPodNamespace, mounterPodName)
 	}
 
+	// Verify workload service account name matches mounter pod.
+	vc := req.GetVolumeContext()
+	if vc[VolumeContextKeyServiceAccountName] != mounterPod.Spec.ServiceAccountName {
```
uriel-guzman: What happens when there is no service account specified? Will the `vc[VolumeContextKeyServiceAccountName]` be empty, and the `mounterPod.Spec.ServiceAccountName` be empty too and the validation will still work? If so, let's add that as a comment so anyone wondering about the same thing also gets assured that it works as intended. Also, manually verify that it works so that we are sure if not manually validated yet.

pkg/csi_driver/node.go
```diff
@@ -159,6 +159,18 @@ func (s *nodeServer) NodePublishVolumeForSharedMount(_ context.Context, req *csi
...
+		return nil, status.Errorf(codes.PermissionDenied, "workload serviceAccountName %q does not match mounter serviceAccountName %q", vc[VolumeContextKeyServiceAccountName], mounterPod.Spec.ServiceAccountName)
```
uriel-guzman: `mounter` -> `mounter pod`

pkg/csi_driver/shared_mount.go
```diff
@@ -546,3 +552,26 @@ func checkMounterPodErrorFile(emptyDirPath string) (codes.Code, error) {
 
 	return extractErrorFromGcsFuseErrorFile(errMsg)
 }
+
+// validateFSGroup checks if the workload fsGroup from the CSI request matches the mounter Pod's fsGroup.
+func validateFSGroup(workloadFSGroup string, mounterSC *corev1.PodSecurityContext) error {
```
uriel-guzman: Wasn't there also a similar thing in webhook package? Can't we just put it somewhere and use it for both instead of having duplicate function? Probably exporting it from webhook and doing something like webhook.ValidateFSGroup?

uriel-guzman (changes requested):

pkg/csi_driver/node.go
yaozile123: Yes, when unspecified K8s defaults serviceAccountName to "default" on both the workload and mounter pods , so the equality check passes as expected. I have added an explanatory comment

pkg/csi_driver/node.go

pkg/csi_driver/shared_mount.go
yaozile123: Done. I exported ValidateFSGroup and ValidateServiceAccountName from webhook and reused them across both webhook and node publish.

pkg/csi_driver/controller_test.go
```diff
@@ -38,17 +38,19 @@ import (
...
-	testImage              = "k8s.gcr.io/pause"
+	testVolumeID                 = "test-volume-id"
+	testNodeID                   = "test-node-id"
+	testPV                       = "test-pv"
+	testPVC                      = "test-pvc"
+	testMounterPodTemplate       = "test-mounter-pod-template"
+	testNamespace                = "test-namespace"
+	testPod                      = "test-pod"
+	testImage                    = "k8s.gcr.io/pause"
+	testFSGroup            int64 = 1000
```
uriel-guzman, line 53: Is this valid golang code?

pkg/cloud_provider/clientset/fake.go
```diff
@@ -49,6 +49,8 @@ type FakePodConfig struct {
 	PodStatus          *corev1.PodStatus
 	NodeName           string
 	IsMounterPod       bool
+	SecurityContext    *corev1.PodSecurityContext
```
uriel-guzman, line 52: these are all the fields that can go into security context:

runAsUser: The User ID (UID) to run the entrypoint of the container process.
runAsGroup: The Primary Group ID (GID) to run the entrypoint of the container process.
runAsNonRoot: A boolean indicating that the container must run as a non-root user. If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root).
fsGroup: A special supplemental group ID applied to all volumes mounted by the Pod. This allows containers to read and write to mounted volumes.
fsGroupChangePolicy: Defines the behavior for changing ownership and permissions of volumes before being exposed inside a Pod. Allowed values are OnRootMismatch and Always.
supplementalGroups: A list of group IDs applied to the first process run in each container, in addition to the container's primary GID.
seLinuxOptions: The SELinux context to be applied to all containers (includes level, role, type, user).
seccompProfile: The seccomp profile to use for the containers in this Pod (includes type and localhostProfile).
sysctls: A list of namespaced kernel parameters (sysctls) to set in the Pod.
windowsOptions: Windows-specific settings applied to all containers (includes gMSACredentialSpecName, gMSACredentialSpec, runAsUserName, hostProcess).
appArmorProfile: The AppArmor profile to apply to the Pod's containers (Added in newer K8s versions, replaces older annotations).

I think it's risky to allow the customer to modify all of these in the mounter pod. It'd probably be best to only allow the customer to specify fsGroup, since we don't really know if the other features work / if we could get an Equilibrium bug due to unknown vulnerabilities.

pkg/csi_driver/controller_test.go
yaozile123: Yeah it is a typed constant

uriel-guzman (changes requested):

pkg/cloud_provider/clientset/fake.go
yaozile123: Done, I only kept fsGroup in security context

uriel-guzman (approved):

PR #1603, opened 2026-08-25
use static vendor mode in sanity-test target

chrisThePattyEater:
**What type of PR is this?**
/kind bug

Auto-VC configuration manifests drive two independent vulnerability remediation pipelines per repository:
* **Go Module Patches** (Enabled globally via `goModVulnerabilityUpdate: enabled: true`): Scans dependencies via `govulncheck` to patch CVEs.
* **Base Image Updates** (Defined via custom hooks under `commits: - name: baseimages`): Bumps compilers and OS bases in Dockerfiles.

Because pipelines are decoupled, they concurrently generate independent PRs. Previously, the `sanity-test` Makefile target  executed `go mod tidy` dynamically on every test run. When base image remediation hooks ran sanity tests to verify Dockerfile updates,`test/go.mod` and `test/go.sum` were automatically modified and bundled into the base image PR. Merging one PR made the other stale,  
  causing merge conflicts and requiring manual reconciliation.                                                                          
                                                                                                                                        
    Additionally, in Go 1.14+, executing `go` commands inside `test/` automatically defaults to `-mod=vendor` due to the presence of    
  `test/vendor/modules.txt`. Under `-mod=vendor`, Go ignores the `replace ... => ../` directive in `test/go.mod` and compiles against   
  stale static snapshots in `test/vendor/` rather than the live `pkg/` codebase.                                                        
                                                                                                                                        
    This PR:                                                                                                                            
    1. Removes dynamic `go mod tidy` from `make sanity-test` to prevent unintended dependency file mutations during test runs.          
    2. Explicitly specifies `-mod=readonly` across all `test/` targets (`sanity-test`, `build-e2e-test`, `test/e2e/run-e2e-local.sh`,   
  and `test/e2e/run-e2e-ci.sh`). This forces Go to respect the local `replace` directive and compile against the live `pkg/` code.      
    3. Updates `docs/development.md` with the workflow for updating `test/go.mod` dependencies.     

Fixes #

After making the code change I ran go mod vendor in the test dir and ran sanity tests at root. Also verified e2e runs as expected.
```
cd test && go mod vendor
make sanity-test
Ran 28 of 92 Specs in 0.407 seconds
SUCCESS! -- 28 Passed | 0 Failed | 1 Pending | 63 Skipped
--- PASS: TestSanity (0.42s)
```

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -389,11 +389,15 @@ unit-test:
 	go test -v -mod=vendor -timeout 30s "./pkg/..." -cover
 
 sanity-test:
-	cd test && go mod tidy && go test -mod=readonly -v -timeout 30s "./sanity/" -run TestSanity
+	# Note: Any go commands inside the test/ directory must explicitly specify -mod=readonly.
+	# This prevents Go from defaulting to -mod=vendor and compiling against stale snapshots.
+	cd test && go test -mod=readonly -v -timeout 30s "./sanity/" -run TestSanity
 
+# Note: Any go commands inside the test/ directory must explicitly specify -mod=readonly.
+# This prevents Go from defaulting to -mod=vendor and compiling against stale snapshots.
 build-e2e-test:
-	cd test && go build -o ../bin/e2e-test-ci ./e2e
-	cd test && go test -c -o ../bin/e2e-test ./e2e
+	cd test && go build -mod=readonly -o ../bin/e2e-test-ci ./e2e
+	cd test && go test -mod=readonly -c -o ../bin/e2e-test ./e2e
 
 e2e-test:
 	./test/e2e/run-e2e-local.sh
--- a/docs/development.md
+++ b/docs/development.md
@@ -145,6 +145,8 @@ Refer to [Test](../test/README.md) documentation.
 
 Follow the following steps to update go modules:
 
+### Root Module (`go.mod`)
+
 1. Open the [go.mod](../go.mod) file in Visual Studio Code.
 2. Click "Check for upgrades" above the first `require` block.
 3. Hover over the module with available upgrade. Click "Quick Fix" to apply the upgrade.
@@ -152,6 +154,17 @@ Follow the following steps to update go modules:
 5. Run `go mod tidy` to ensure that the listed dependencies are really still needed.
 6. Run `go mod vendor` to update vendor directory.
 7. Resolve any issues that may be introduced by the new modules.
+
+### Test Module (`test/go.mod`)
+
+The `test/` directory maintains its own Go module. Tests run with `-mod=readonly`:
+
+1. Add or update dependencies in [`test/go.mod`](../test/go.mod).
+2. Tidy the test module dependencies:
+   ```bash
+   cd test && go mod tidy
+   ```
+3. Commit the updated `test/go.mod` and `test/go.sum`.
 
 ## Troubleshooting
 
--- a/test/e2e/run-e2e-ci.sh
+++ b/test/e2e/run-e2e-ci.sh
@@ -80,7 +80,7 @@ cd "${PKGDIR}"
 
 # Build e2e-test CLI
 pushd test
-go build -o ${PKGDIR}/bin/e2e-test-ci ./e2e
+go build -mod=readonly -o ${PKGDIR}/bin/e2e-test-ci ./e2e
 popd
 chmod +x ${PKGDIR}/bin/e2e-test-ci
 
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -389,7 +389,7 @@ unit-test:
 	go test -v -mod=vendor -timeout 30s "./pkg/..." -cover
 
 sanity-test:
-	cd test && go mod tidy && go test -mod=readonly -v -timeout 30s "./sanity/" -run TestSanity
+	cd test && go test -mod=vendor -v -timeout 30s "./sanity/" -run TestSanity
```
amacaskill: Because `test/go.mod` uses `replace => ../`, the `test/vendor/` directory acts as a static, copied snapshot of our root codebase. The `-mod=vendor` flag does **not** update this snapshot—it just forces `go test` to blindly evaluate whatever is currently sitting in that folder.

For example, if we accidentally break `pkg/csi_driver/utils.go` but forget to manually update the snapshot beforehand, I think the CI will silently test the older, working copy located in `test/vendor/`, hiding issues. 

To fix the original issue without breaking CI testing, I'm wondering if we can just remove the command that mutates files (`go mod tidy`), but leave `go test` pointing at the live code (`-mod=readonly`):

`cd test && go test -mod=readonly -v -timeout 30s "./sanity/" -run TestSanity`  	

Note that `-mod=readonly` will strictly fail CI if `test/go.mod` falls out of sync with `go: updates to go.mod needed, disabled by -mod=readonly`.  To prevent this, just ensure your dedicated Go module auto-bumping bot is configured to execute `go mod tidy` in both the root and `test/` directories when it generates its dependency PRs.

If you don't have any objections, can you please try this out to see if this suggestion works? Try out some cases to make sure we do see `go: updates to go.mod needed, disabled by -mod=readonly`` / no other major issues with it?

chrisThePattyEater: Yes this is right I tried breaking an example file "pkg/csi_driver/utils.go" and ran sanity tests but it passes.

I applied the fix and see the failures as expected.

I updated the make file and documentation to match this.

amacaskill: Thank you! I think now we can  revert your commits touching test/vendor/..

Also, I think we have this in other places in the Makefile. IIUC, if a `vendor/` folder physically exists, any Go command that omits a `-mod` flag will automatically default to `-mod=vendor` (correct me if I'm wrong). Because commands like `make build-e2e-test` don't explicitly specify a flag, we are currently ignoring our live workspace code and compiling the E2E tests against the stagnant snapshot sitting inside `test/vendor/`. We will hit this issue for any new commands added too.

Without a  `test/vendor/` folder , .we would have to download the frameworks from scratch which we don't want to do. I think for now, lets explicitly add `-mod=readonly` to the `build-e2e-test` and `go test -c` targets in the Makefile wherever it is needed, then add a comment in the makefile somewhere saying  any new Go commands referencing the `test/` directory must explicitly use `-mod=readonly`, to ensure they evaluate the live driver code via the `replace => ../` directive rather than falling back to the stagnant vendor snapshot.

amacaskill: I think: 

```
  sanity-test:
       +# Note: Any go commands inside the test/ directory must explicitly specify -mod=readonly.
        +# This prevents Go from defaulting to -mod=vendor and compiling against stale snapshots.
  	cd test && go test -mod=readonly -v -timeout 30s "./sanity/" -run TestSanity
  
+# Note: Any go commands inside the test/ directory must explicitly specify -mod=readonly.
+# This prevents Go from defaulting to -mod=vendor and compiling against stale snapshots.
  build-e2e-test:
- 	cd test && go build -o ../bin/e2e-test-ci ./e2e
- 	cd test && go test -c -o ../bin/e2e-test ./e2e
+ 	cd test && go build -mod=readonly -o ../bin/e2e-test-ci ./e2e
+ 	cd test && go test -mod=readonly -c -o ../bin/e2e-test ./e2e
```

but please confirm.  AFter you test it works, feel free to merge

amacaskill (approved):

Makefile
chrisThePattyEater: Yeah based on [this](https://go.dev/ref/mod#:~:text=At%20go%201.14%20or%20higher%2C%20automatic%20vendoring%20may%20be%20enabled.%20If%20the%20file%20vendor/modules.txt%20is%20present%20and%20consistent%20with%20go.mod%2C%20there%20is%20no%20need%20to%20explicitly%20use%20the%20%2Dmod%3Dvendor%20flag.) this is correct.

I also tested with another broken file 
`go build -o /dev/null ./e2e` Passes bc compiles against stale vendored files
`go build -mod=readonly -o /dev/null ./e2e` Fails bc compiles local files

So I applied the fix you mentioned also I noticed we do the same thing in test/e2e/run-e2e-ci.sh    and test/e2e/run-e2e-local.sh so I applied it there.

chrisThePattyEater: updated comment

PR #1607, opened 2026-08-25
feat(metrics): migrate GCSFuse volume metrics to k8s_gcs_fuse_volume schema with bidirectional rollback support

alleaditya:
### Description

This PR migrates GCSFuse operational metrics to the new Monitored Resource schema (**`k8s_gcs_fuse_volume`**) while introducing an explicit boolean CLI feature flag (**`--enable-gcsfuse-volume-metrics-schema`**, default `false`) to support safe, zero-downtime live rollbacks to the legacy `k8s_pod` Monitored Resource without restarting workload pods.

### Key Changes
1. **Feature Flag (`--enable-gcsfuse-volume-metrics-schema`, default `false`)**:
   - Added to `cmd/csi_driver/main.go` and plumbed through `pkg/csi_driver/gcs_fuse_driver.go` and `pkg/csi_driver/node.go`.
   - When `false` (default/rollback), `MetricSchemaPolicy.BuildConstLabels` restores the exact legacy `k8s_pod` label set (`pod_name`, `namespace_name`, `volume_name`, `bucket_name`, `pod_uid: ""`) and omits `k8s_controller_type`/`k8s_controller_name`.
   - When `true`, emits `k8s_gcs_fuse_volume` labels (`pod_name`, `namespace_name`, `volume_name`, `bucket_name`, `k8s_controller_type`, `k8s_controller_name`) and omits `pod_uid`.
2. **Bidirectional Metric Name Normalization (`MetricSchemaPolicy.NormalizeMetricName`)**:
   - Handles all 4 combinations of `(flag=false|true) x (sidecar_binary=legacy_singular|new_plural)` for `fs_ops_latency` <-> `fs_ops_latencies`.
   - Enables live DaemonSet flag rollbacks via kubelet `requiresRepublish: true` without restarting running workload mounter sidecar containers.
3. **Kubernetes Controller Extraction (`ExtractK8sController`)**:
   - Strips 10-character alphanumeric `pod-template-hash` suffixes from ReplicaSets to resolve `("deployment", "<deployment-name>")` directly without extra API server round-trips.
4. **Defensive & Concurrency Hardening**:
   - Defensive map cloning in `NewMetricsCollector`.
   - Mutex protection in `FakeClientset.GetPod` for race-free concurrent testing.

### Verification
- Unit tests with race detection: `go test -race ./pkg/metrics/... ./pkg/csi_driver/...` -> **100% PASSED (0 race conditions)**
- JIT Mutation Testing -> **100% Mutant Kill Rate (3/3 semantic faults killed)**
- End-to-End Unmanaged GKE cluster verification (`alleaditya-test-cluster`) -> **Verified both Mode A (`flag=true`) and Mode B (Live rollback `flag=false` via `requiresRepublish` with 0 pod restarts)**


amacaskill:
We won't be able to merge this without a feature flag until we start the customer outreach . See my email. 

Let's put this behind a feature flag so that we can get the changes ready to go, then just flip the feature on from the component yaml for when we are ready to release it. This also lets us roll back easier. See all the feature flags here: https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/cd894e0b83ac2478e1ae90199b59b79e476b5da1/cmd/csi_driver/main.go#L73-L76. TO start , you will default to false, then when we are ready, we enable it in the managed component yaml


Here is an example PR that gates new feature behind enable-auto-gomemlimit flag (though more complicated since it has changes in webhook AND sidecar mounter, your PR doesn't need this, and just needs the changes in : 
- cmd/csi_driver/main.go, Flag should be false in this. 
- deploy/base/node/node.yaml, this can be true so we can start CI testing on your feature
- pkg/csi_driver/node.go, all changes in node.go should be hidden behind this flag. 


https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1364.

cmd/csi_driver/main.go
```diff
@@ -96,6 +96,9 @@ var (
 	// Stream metrics export
 	streamMetricsExport = flag.Bool("stream-metrics-export", false, "Stream metrics export instead of downloading the entire content")
 
+	// GCSFuse volume metrics schema flag
+	enableGcsFuseVolumeMetricsSchema = flag.Bool("enable-gcsfuse-volume-metrics-schema", true, "Enable gke.googleapis.com/GcsFuseVolume Monitored Resource metrics schema for GCSFuse metrics.")
```
amacaskill: default should be false.

pkg/metrics/metrics.go
```diff
@@ -114,13 +116,36 @@ func (mm *manager) RegisterMetricsCollector(mountPath, podNamespace, podName, bu
...
+		constLabels = map[string]string{
+			"pod_name":       podName,
+			"namespace_name": podNamespace,
+			"volume_name":    volumeName,
+			"bucket_name":    bucketName,
+			"jobset_name":    jobsetName,
+		}
+	} else {
+		constLabels = map[string]string{
+			"pod_uid":     podUID,
```
amacaskill, line 131: Prior to this PR, the NewMetricsCollector explicitly defined pod_uid as "" to reduce cardinality.   Your else block brings back the true podUID string and actively drops the other standard labels. If the feature flag is disabled by a user to intentionally use the old stable metrics, things will not work as they did before.

pkg/metrics/metrics.go
```diff
@@ -114,13 +116,36 @@ func (mm *manager) RegisterMetricsCollector(mountPath, podNamespace, podName, bu
...
-		"namespace_name": podNamespace,
-		"volume_name":    volumeName,
-		"bucket_name":    bucketName,
-		"pod_uid":        "", // podUID is emptied to avoid infinite cardinality in the metric labels
-	}, mm.clientset, mm.streamMetrics)
+	var constLabels map[string]string
+	if mm.enableGcsFuseVolumeMetricsSchema {
+		jobsetName := ""
+		if mm.clientset != nil {
+			if pod, err := mm.clientset.GetPod(podNamespace, podName); err == nil && pod != nil && pod.Labels != nil {
```
amacaskill: Making a GetPod API server call here is redundant  NodePublishVolume in pkg/csi_driver/node.go already holds the fully resolved Pod object. 

 Calculate the jobsetName safely inside pkg/csi_driver/node.go using the existing pod.Labels and pass jobsetName directly into RegisterMetricsCollector as a string parameter.

test/e2e/testsuites/metrics.go
```diff
@@ -44,15 +44,15 @@ import (
 var expectedMetricNames = []string{
 	"fs_ops_count",
 	"fs_ops_error_count",
-	"fs_ops_latency",
+	"fs_ops_duration_seconds",
```
amacaskill: We need to make sure we communicate these changes during the customer outreach. Please note this in the customer outreach doc you are planning so we can track somewhere / make sure it is mentioned in the design doc

pkg/csi_driver/node.go
```diff
@@ -1357,3 +1355,17 @@ func (s *nodeServer) checkWINodeLabel(node *corev1.Node, isHostNetwork bool) err
 
 	return nil
 }
+
+func countGCSFuseVolumes(pod *corev1.Pod) int {
```
amacaskill: Code Duplication with countGCSFuseVolumes In pkg/csi_driver/node.go, you added:

In the original node.go code, there was a method receiver s.countGcsFuseVolumes(pod) (int, error) utilized on lines you patched over. If you didn't explicitly delete the original countGcsFuseVolumes in this PR (it wasn't in the patch removals), you are establishing duplicated dead code. You should either reuse the original s.countGcsFuseVolumes(pod) method or rip it out

test/e2e/testsuites/metrics.go
amacaskill: Please let me know where you are going to track this before you resolve this comment.

pkg/csi_driver/node.go
```diff
@@ -444,15 +444,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
-
-		if err != nil {
-			klog.Errorf("Metrics collection is disabled for Pod %s/%s as counting the number of GCS FUSE volumes failed with error: %v", pod.Namespace, pod.Name, err)
-		} else if gcsFuseVolumeCount > maxGCSFuseVolumesForMetrics {
-			klog.Warningf("Metrics collection is disabled for Pod %s/%s as the number of GCS FUSE volumes is %d, which is greater than the limit of %d.", pod.Namespace, pod.Name, gcsFuseVolumeCount, maxGCSFuseVolumesForMetrics)
-		} else if podUID != "" && volumeName != "" && emptyDirBasePath != "" {
-			klog.V(4).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
-			s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, args.bucketName, s.driver.config.NodeID, emptyDirBasePath, podUID, volumeName)
+		if podUID != "" && volumeName != "" && emptyDirBasePath != "" {
+			if s.driver.config.FeatureOptions != nil && !s.driver.config.FeatureOptions.EnableGcsFuseVolumeMetricsSchema && countGCSFuseVolumes(pod) > maxGCSFuseVolumesForMetrics {
```
amacaskill: If s.driver.config.FeatureOptions is nil for any reason (e.g. during certain tests, partial initialization, or future code paths), the entire expression evaluates to false (due to short-circuiting on != nil)! Because it is false, it drops down to the else block and registers the metrics collector, bypassing the 10-volume cap completely.

 If the metrics manager is running the legacy schema, this re-introduces the cardinality issue. 

If the config isn't explicitly enabling the new schema, enforce the old schema's 10-volume safety ceiling

pkg/csi_driver/node.go
```diff
@@ -444,15 +444,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+				klog.V(4).Infof("NodePublishVolume metrics collection disabled for target path %q because pod %q has > %d volumes", targetPath, pod.Name, maxGCSFuseVolumesForMetrics)
```
amacaskill: why did we downgrade the warning to an info log?

pkg/csi_driver/node.go
```diff
@@ -444,15 +444,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
-		gcsFuseVolumeCount, err := s.countGcsFuseVolumes(pod)
-
-		if err != nil {
-			klog.Errorf("Metrics collection is disabled for Pod %s/%s as counting the number of GCS FUSE volumes failed with error: %v", pod.Namespace, pod.Name, err)
-		} else if gcsFuseVolumeCount > maxGCSFuseVolumesForMetrics {
-			klog.Warningf("Metrics collection is disabled for Pod %s/%s as the number of GCS FUSE volumes is %d, which is greater than the limit of %d.", pod.Namespace, pod.Name, gcsFuseVolumeCount, maxGCSFuseVolumesForMetrics)
-		} else if podUID != "" && volumeName != "" && emptyDirBasePath != "" {
-			klog.V(4).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
-			s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, args.bucketName, s.driver.config.NodeID, emptyDirBasePath, podUID, volumeName)
+		if podUID != "" && volumeName != "" && emptyDirBasePath != "" {
```
amacaskill: From AI: 


**3. Inverted Conditional Precedence**
In the old code, the sequence was:
1. Are there $> 10$ volumes? If yes, Log warning.
2. `else if podUID != "" && volumeName != ""` -> enable metrics.

In the *new* code, the check for `podUID != ""` and `volumeName != ""` comes **first** (the outer `if`). 
If the volume lacks an `emptyDirBasePath` or `podUID`, it silently evaluates to false and totally bypasses the `countGCSFuseVolumes(pod) > 10` cap check altogether, meaning the warning is never logged.

### How it should be written:
The code should extract the boolean securely, restore the `Warningf` level, and match the original evaluation order:

```go
// Safely evaluate if the schema is explicitly enabled
schemaEnabled := s.driver.config.FeatureOptions != nil && s.driver.config.FeatureOptions.EnableGcsFuseVolumeMetricsSchema
volumeCount := countGCSFuseVolumes(pod)

if !schemaEnabled && volumeCount > maxGCSFuseVolumesForMetrics {
	// Restored to Warningf
	klog.Warningf("Metrics collection is disabled for Pod %s/%s as the number of GCS FUSE volumes is %d, which is greater than the limit of %d.", pod.Namespace, pod.Name, volumeCount, maxGCSFuseVolumesForMetrics)
} else if podUID != "" && volumeName != "" && emptyDirBasePath != "" {
	klog.V(4).Infof("NodePublishVolume enabling metrics collector for target path %q", targetPath)
	s.driver.config.MetricsManager.RegisterMetricsCollector(targetPath, pod.Namespace, pod.Name, args.bucketName, s.driver.config.NodeID, emptyDirBasePath, podUID, volumeName)
}
```

amacaskill:
You will need to squash your commits into 1 before merging

pkg/csi_driver/node_test.go
amacaskill, line 1: The existing logic still relies on the 10-volume cap if the EnableGcsFuseVolumeMetricsSchema flag is disabled (false). By flipping all existing mock tests to true, there is now zero test coverage for the 10-volume cap enforcement.

Please do not modify the old tests. You should instead  duplicate the table-driven test runner and run full set of tests for both cases: 

Run the test matrix with EnableGcsFuseVolumeMetricsSchema = false (where >10 volumes expects false).
Run the test matrix with EnableGcsFuseVolumeMetricsSchema = true (where >10 volumes expects true).

pkg/metrics/metrics_test.go
```diff
@@ -148,6 +151,148 @@ func TestRegisterUnregisterMetricsCollector(t *testing.T) {
 	}
 }
 
+func TestMetricsCollectorGcsFuseVolumeLabels(t *testing.T) {
```
amacaskill: This unit test invokes `NewMetricsCollector(...)`, passes it a hardcoded `map[string]string` of labels, and then tests if the returned object has that exact map attached to it. It does not test `RegisterMetricsCollector`, which means it isn't actually testing the mapping logic or the feature flag; its basically just testing if Go maps can be assigned to a struct. 

Please fix this test to correctly test `RegisterMetricsCollector`

Additionally you are missing `jobset_name`.

amacaskill: EDIT: 

Because you already wrote TestEndToEndGcsFuseVolumeMetricsScraping (which spins up the registry, calls RegisterMetricsCollector, and gathers the actual Prometheus metrics output), I think TestMetricsCollectorGcsFuseVolumeLabels is redundant and should just be deleted.

You should turn TestEndToEndGcsFuseVolumeMetricsScraping into a table-driven test to verify the labels for both branches (when the flag is on, and when it is off).

pkg/metrics/metrics_test.go
```diff
@@ -148,6 +151,148 @@ func TestRegisterUnregisterMetricsCollector(t *testing.T) {
...
+		if got := c.constLabels[k]; got != want {
+			t.Errorf("constLabel[%q] = %q, want %q for gke.googleapis.com/GcsFuseVolume resource target label propagation", k, got, want)
+		}
+	}
+	if val, ok := c.constLabels["pod_uid"]; ok {
+		t.Errorf("pod_uid constant label should not be present for gke.googleapis.com/GcsFuseVolume MR target, but got %q", val)
+	}
+}
+
+func TestEndToEndGcsFuseVolumeMetricsScraping(t *testing.T) {
```
amacaskill, line 161: TestEndToEndGcsFuseVolumeMetricsScraping ignores legacy regressions In Patch 3, the E2E scraper test initializes NewMetricsManager(..., enableGcsFuseVolumeMetricsSchema=true).

As mentioned in other comment, we should have table driven test with full test coverage for all cases when enableGcsFuseVolumeMetricsSchema=true vs enableGcsFuseVolumeMetricsSchema=false. 

- There is no test case checking what labels are scraped when enableGcsFuseVolumeMetricsSchema is initialized to false. I
- The test explicitly injects jobset.sigs.k8s.io/jobset-name on the mocked Pod. There should be a condition or separate test asserting that an empty string "" correctly propagates when the pod lacks the label, avoiding crashes.

pkg/metrics/metrics_test.go
```diff
@@ -18,18 +18,21 @@ package metrics
...
+	"github.com/googlecloudplatform/gcs-fuse-csi-driver/pkg/util"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 func TestNewMetricsManager(t *testing.T) {
```
amacaskill, line 39: Here you are  hardcoding the new boolean to true across all permutations. The table-driven test parameters (the struct) should be updated to pass true/false and verify that the manager.enableGcsFuseVolumeMetricsSchema field is correctly populated in each case.

test/e2e/testsuites/metrics.go
amacaskill, line 1: Don't we need to update the E2E tests need to assert that the jobset_name label is correctly populated in the scraped /metrics payload when the new schema feature flag is true?  If you add this , You will need to make sure this test only runs for a min GKE version and/or GCSFuse version so the test doesn't start running incorrectly on past versions

test/e2e/testsuites/metrics.go
amacaskill: Similarly, we cannot just change the expectedMetricNames slice to expect fs_ops_duration_seconds instead of fs_ops_latency without any version check.  Without a version check, this will immeditaly start failing on the managed testgrid.

pkg/csi_driver/node_test.go
alleaditya: This PR is currently a work in progress. Please bear with me as I finish implementing the requested changes, complete the testing, and update the PR for a final review.

PR #1612, opened 2026-08-28
Add pre-flight Capacity Advisor region evaluation to mitigate cluster stockouts

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

After introducing GCE Capacity Advisor to select the best zone, we still encounter stockout issues (`GCE_STOCKOUT` / `ZONE_RESOURCE_POOL_EXHAUSTED`) in `us-central1` when creating 9-node `n2-standard-8` clusters because the entire region frequently runs out of available N2 capacity.
This PR addresses this by:
1. **Pre-flight Capacity Advisor evaluation across candidate regions**: Before initiating cluster creation (for both Standard and Autopilot clusters), queries the GCE Capacity Advisor across candidate fallback regions when Zonal Buckets (`EnableZB`) are disabled.
2. **Stack-ranked region & zone selection**: Stack-ranks all evaluated candidate regions by obtainability score and selects the highest-scoring region for a single cluster creation attempt. For Standard clusters, pins `--node-locations` to the advisor-recommended single zone.
3. Updates `modifyPVForProfiles` in `profiles.go` to dynamically derive Anywhere Cache zones (`<region>-c`) from `CLUSTER_LOCATION` instead of hardcoding `us-central1-c`.

Fixes #

I verified manually in a boskos project and  successfully provisioned a 9-node `n2-standard-8` cluster in `us-east4` (`us-east4-c`) without stockouts

```release-note

```

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -101,6 +101,7 @@ You can control the test through the following make parameters, eg `make e2e-tes
 - `ENABLE_ZB`: default value is `false`. Change it to `true` if you want the bucket used during testing to be a Zonal Bucket created in the zone -> "$(location provided)" + "-c".
 - `GCSFUSE_CLIENT_PROTOCOL`: default value is 'http1'. Change to 'grpc' to alter the type of protocol gcsfuse uses to communicate with gcs
 - `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE`: default value is `false`. Change it to `true` if you want the test runner to create and destroy the GKE cluster for the test.
+- `USE_CAPACITY_ADVISOR`: default value is `false` (defaults to `true` in Prow CI). Whether to use GCE Capacity Advisor to select node locations and fallback regions. Only used when `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true`; ignored when running on an existing cluster.
 - `E2E_TEST_GKE_CLUSTER_VERSION`: It denotes the GKE cluster master and node version. If `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE` is `true`, it specifies the version of the cluster to be created (defaults to `latest` if not provided). If `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE` is `false`, it auto-detects the version of the existing cluster (can be overridden by setting this parameter). The version is used by the test framework to check for feature compatibility.
 - `E2E_TEST_USE_BOSKOS`: default value is `false`. Change it to `true` if you want to use Boskos to acquire a project. Useful for debugging Boskos infrastructure locally.
 - `E2E_TEST_PROJECT_ID`: default value is empty. GCP project ID to run E2E tests. Required when managing cluster lifecycle without Boskos.
@@ -130,9 +131,11 @@ When `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE` is set to `true`, the test runner provi
 - **Cluster Type**: Standard GKE Cluster (default) or Autopilot (if `E2E_TEST_USE_GKE_AUTOPILOT=true`).
 - **Release Channel**: `rapid` (can be overridden by `GKE_RELEASE_CHANNEL` environment variable).
 - **Version**: Determined by `E2E_TEST_GKE_CLUSTER_VERSION` (local) or `GKE_CLUSTER_VERSION` (CI).
-- **Region**: Determined by `GKE_CLUSTER_REGION` (default: `us-central1`).
+- **Region & Capacity Selection**: Determined by `GKE_CLUSTER_REGION` (default: `us-central1`).
+  - **Dynamic Fallback & Capacity Checks**: When `USE_CAPACITY_ADVISOR=true`, as enabled by default in Prow, the test runner pre-flights compute availability across candidate fallback regions using the Compute Engine Capacity Advisor API. It evaluates obtainability scores, stack-ranks candidate regions from highest to lowest score, and attempts cluster creation in the top-ranked region to mitigate regional stockouts. Note that `USE_CAPACITY_ADVISOR` is only used when `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true`, and is ignored when running tests against an existing cluster.
+  - **Bucket Co-location**: Selecting a fallback region also implicitly updates the `--test-bucket-location` passed to Ginkgo, ensuring test GCS buckets are co-located in the same region as the cluster to prevent cross-region data transfer.
 - **Node Pool Configuration** (Standard cluster only):
-  - **Node Count**: 3 nodes (can be overridden by `E2E_TEST_NUM_NODES` or `NUMBER_NODES` env var).
+  - **Node Count**: 9 nodes (can be overridden by `E2E_TEST_NUM_NODES` or `NUMBER_NODES` env var).
   - **Machine Type**: `n2-standard-8` (can be overridden by `MACHINE_TYPE` env var).
   - **Image Type**: `cos_containerd` (Container-Optimized OS with containerd).
 - **Workload Identity**: Enabled by default using the GCP project's workload pool (`<project-id>.svc.id.goog`).
--- a/test/e2e/main.go
+++ b/test/e2e/main.go
@@ -31,12 +31,12 @@ var (
 	pkgDir = flag.String("pkg-dir", "", "the package directory")
 
 	// Kubernetes cluster flags.
-	gkeClusterRegion               = flag.String("gke-cluster-region", "", "region that gke regional cluster should be created in")
+	gkeClusterRegion               = flag.String("gke-cluster-region", "", "region that gke regional cluster should be created in, which also sets --test-bucket-location for Ginkgo")
 	gkeClusterVersion              = flag.String("gke-cluster-version", "", "GKE cluster worker master and node version")
 	gkeReleaseChannel              = flag.String("gke-release-channel", "rapid", "GKE cluster release channel")
 	gkeNodeVersion                 = flag.String("gke-node-version", "", "GKE cluster worker node version")
 	nodeMachineType                = flag.String("node-machine-type", "n2-standard-8", "GKE cluster worker node machine type")
-	numNodes                       = flag.Int("number-nodes", 3, "number of nodes in the test cluster")
+	numNodes                       = flag.Int("number-nodes", 9, "number of nodes in the test cluster")
 	useGKEAutopilot                = flag.Bool("use-gke-autopilot", false, "use GKE Autopilot cluster for the tests")
 	apiEndpointOverride            = flag.String("api-endpoint-override", "https://container.googleapis.com/", "CloudSDK API endpoint override to use for the cluster environment")
 	nodeImageType                  = flag.String("node-image-type", "cos_containerd", "image type to use for the cluster")
@@ -51,7 +51,7 @@ var (
 	skipCSIDriverInstall           = flag.Bool("skip-csi-driver-install", false, "skips the install of the driver. You must have manually deployed the driver and webhook.")
 	gcsFusePrNumber                = flag.String("gcsfuse-pr-number", "", "PR number for gcsfuse to test against")
 	// Only works for zonal clusters by pinning nodes to a single advised zone, set to false if want to create regional cluster.
-	useCapacityAdvisor = flag.Bool("use-capacity-advisor", false, "whether to use GCE Capacity Advisor to select node locations")
+	useCapacityAdvisor = flag.Bool("use-capacity-advisor", false, "whether to use GCE Capacity Advisor to select node locations. Only used when --manage-cluster-lifecycle=true; ignored for existing clusters.")
 
 	// Test infrastructure flags.
 	inProw                 = flag.Bool("run-in-prow", false, "whether or not to run the test in PROW")
@@ -106,6 +106,11 @@ func main() {
 			*gkeClusterVersion = "latest"
 		}
 		utils.EnsureVariable(gkeReleaseChannel, true, "'gke-release-channel' must be set when managing cluster lifecycle")
+	}
+
+	if !*manageClusterLifecycle && *useCapacityAdvisor {
+		klog.Infof("Ignoring --use-capacity-advisor because --manage-cluster-lifecycle is false (tests are running on an existing cluster)")
+		*useCapacityAdvisor = false
 	}
 
 	if *useBoskos {
--- a/test/e2e/testsuites/profiles.go
+++ b/test/e2e/testsuites/profiles.go
@@ -538,14 +538,25 @@ func modifyPVForProfiles(testScenario string, pv *v1.PersistentVolume, sc string
 		mo = append(mo, fmt.Sprintf("%s:%s", fileCacheSizeMiBMountOptionKey, "30"))
 
 		if sc == "gcsfusecsi-serving" {
-			// We are hardcoding the zones for ac to be us-central1-c because one of the available zones is an ai
-			// specific zone which has limited quota, we should not make caches in this zone.
+			// TODO(yaozile): Create the ANyC in the same zone as the compute (or skip the test
+			// if the zone is not supported for AnyC), rather than hardcoding to <cluster-region>-c.
+			// We are hardcoding the zones for ac to be <cluster-region>-c because some available zones are AI-specific
+			// zones which have limited quota, we should not make caches in those zones.
 			va[anywhereCacheTTLKey] = "2h"
 			va[anywhereCacheAdmissionPolicyKey] = "admit-on-second-miss"
 			isZBEnabled := os.Getenv(utils.IsZBEnabledEnvVar)
 			if isZBEnabled == "false" {
 				// AC is not supported for zb, so we only use AC for non zb tests.
-				va[anywhereCacheZonesKey] = "us-central1-c"
+				clusterLocation := os.Getenv(utils.ClusterLocationEnvVar)
+				if clusterLocation == "" {
+					clusterLocation = "us-central1"
+				}
+				// If clusterLocation is a zone (e.g. "us-central1-a"), extract the region prefix ("us-central1").
+				parts := strings.Split(clusterLocation, "-")
+				if len(parts) > 2 && len(parts[len(parts)-1]) == 1 {
+					clusterLocation = strings.Join(parts[:len(parts)-1], "-")
+				}
+				va[anywhereCacheZonesKey] = clusterLocation + "-c"
 			}
 
 		}
--- a/test/e2e/utils/cluster.go
+++ b/test/e2e/utils/cluster.go
@@ -20,6 +20,7 @@ package utils
 import (
 	"fmt"
 	"os/exec"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -75,9 +76,9 @@ func clusterDownGKE(testParams *TestParameters) error {
 
 // queryRegionalStandardZones retrieves standard compute zones for a region from 'gcloud compute regions describe',
 // which naturally excludes AI-only zones, so they can be passed to Capacity Advisor via --zones.
-func queryRegionalStandardZones(testParams *TestParameters) string {
+func queryRegionalStandardZones(testParams *TestParameters, region string) string {
 	regionArgs := []string{
-		"compute", "regions", "describe", testParams.GkeClusterRegion,
+		"compute", "regions", "describe", region,
 		"--format=value(zones.basename())",
 	}
 	if testParams.ProjectID != "" {
@@ -86,7 +87,7 @@ func queryRegionalStandardZones(testParams *TestParameters) string {
 
 	regionOut, err := gcloudCommand(testParams, regionArgs...).Output()
 	if err != nil {
-		klog.Warningf("Failed to query standard regional compute zones for %s: %v", testParams.GkeClusterRegion, err)
+		klog.Warningf("Failed to query standard regional compute zones for %s: %v", region, err)
 		return ""
 	}
 
@@ -97,24 +98,24 @@ func queryRegionalStandardZones(testParams *TestParameters) string {
 	return strings.Join(zones, ",")
 }
 
-// queryCapacityAdvisedZone calls queryRegionalStandardZones to get all standard GKE zones,
-// and queries Capacity Advisor to select the best zone with sufficient compute capacity.
-func queryCapacityAdvisedZone(testParams *TestParameters) (string, error) {
+// queryCapacityAdvice calls queryRegionalStandardZones to get standard GKE zones in the given region,
+// and queries Capacity Advisor to return the recommended zone and obtainability score.
+func queryCapacityAdvice(testParams *TestParameters, region string) (string, float64, error) {
 	// Note: Capacity Advisor requires --provisioning-model (supports only SPOT or FLEX_START).
 	// We use SPOT as a proxy for regional resource availability.
 	// See: https://cloud.google.com/sdk/gcloud/reference/beta/compute/advice/capacity
 	cmdArgs := []string{
 		"beta", "compute", "advice", "capacity",
-		"--region=" + testParams.GkeClusterRegion,
+		"--region=" + region,
 		"--provisioning-model=SPOT",
 		"--size=" + strconv.Itoa(testParams.NumNodes),
 		"--instance-selection-machine-types=" + testParams.NodeMachineType,
 		"--target-distribution-shape=any-single-zone",
-		"--format=value(recommendations[0].shards[0].zone.basename())",
+		"--format=value(recommendations[0].scores.obtainability,recommendations[0].shards[0].zone.basename())",
 	}
 
 	// Restrict capacity search to standard regional compute zones to avoid non-GKE AI zones.
-	if zonesFilter := queryRegionalStandardZones(testParams); zonesFilter != "" {
+	if zonesFilter := queryRegionalStandardZones(testParams, region); zonesFilter != "" {
 		cmdArgs = append(cmdArgs, "--zones="+zonesFilter)
 	}
 	if testParams.ProjectID != "" {
@@ -124,15 +125,21 @@ func queryCapacityAdvisedZone(testParams *TestParameters) (string, error) {
 	out, err := gcloudCommand(testParams, cmdArgs...).Output()
 	if err != nil {
 		if exitErr, ok := err.(*exec.ExitError); ok {
-			return "", fmt.Errorf("failed to query capacity advice: %w, stderr: %s", err, string(exitErr.Stderr))
+			return "", 0, fmt.Errorf("failed to query capacity advice: %w, stderr: %s", err, string(exitErr.Stderr))
 		}
-		return "", fmt.Errorf("failed to query capacity advice: %w", err)
+		return "", 0, fmt.Errorf("failed to query capacity advice: %w", err)
 	}
-	zone := strings.TrimSpace(string(out))
-	if zone == "" {
-		return "", fmt.Errorf("no zone returned in capacity advice")
+	klog.Infof("Capacity Advisor output for region %s: %s", region, string(out))
+	fields := strings.Fields(string(out))
+	if len(fields) < 2 {
+		return "", 0, fmt.Errorf("no zone or score returned in capacity advice: %q", string(out))
 	}
-	return zone, nil
+	score, err := strconv.ParseFloat(fields[0], 64)
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to parse obtainability score %q: %w", fields[0], err)
+	}
+	zone := fields[1]
+	return zone, score, nil
 }
 
 func clusterUpGKE(testParams *TestParameters) error {
@@ -149,10 +156,91 @@ func clusterUpGKE(testParams *TestParameters) error {
 		klog.Infof("Skipping gcloud components update for local run.")
 	}
 
+	// Fall back across candidate regions to mitigate stockouts; Zonal Buckets
+	// are pinned to testParams.GkeClusterRegion (which defaults to us-central1).
+	// For standard clusters, testParams.GkeClusterRegion is prioritized first,
+	// followed by candidate fallback regions with typically lower contention.
+	candidateRegions := []string{testParams.GkeClusterRegion}
+	if !testParams.EnableZB {
+		fallbackRegions := []string{
+			"us-east4",
+			"us-east1",
+			"us-west1",
+			"us-west4",
+			"europe-west1",
+			"europe-west3",
+		}
+		for _, r := range fallbackRegions {
+			if r != testParams.GkeClusterRegion {
+				candidateRegions = append(candidateRegions, r)
+			}
+		}
+	}
+
+	var nodeLocations string
+
+	// Use Capacity Advisor to select an unconstrained region for both Standard and Autopilot clusters.
+	if testParams.UseCapacityAdvisor {
+		type capacityResult struct {
+			region string
+			zone   string
+			score  float64
+		}
+		var results []capacityResult
+
+		for _, region := range candidateRegions {
+			zone, score, err := queryCapacityAdvice(testParams, region)
+			if err != nil {
+				klog.Warningf("Capacity Advisor query failed for region %q: %v", region, err)
+				continue
+			}
+			klog.Infof("Capacity Advisor for region %q: zone=%q, obtainability=%.2f", region, zone, score)
+			results = append(results, capacityResult{region: region, zone: zone, score: score})
+		}
+
+		if len(results) > 0 {
+			// Sort descending by obtainability score. Stable sort preserves priority for earlier candidate regions on ties.
+			sort.SliceStable(results, func(i, j int) bool {
+				return results[i].score > results[j].score
+			})
+
+			best := results[0]
+			klog.Infof("Stack-ranked candidate regions:")
+			for idx, r := range results {
+				klog.Infof("  #%d: region=%q zone=%q score=%.2f", idx+1, r.region, r.zone, r.score)
+			}
+			klog.Infof("Selecting highest-scoring region %q (zone: %q, score: %.2f)", best.region, best.zone, best.score)
+
+			// Note: Updating GkeClusterRegion here implicitly updates the
+			// --test-bucket-location passed to Ginkgo. This ensures that the
+			// GCS buckets created during tests are co-located in this fallback
+			// region, preventing cross-region data transfer.
+			testParams.GkeClusterRegion = best.region
+			// Autopilot manages node placement dynamically across zones within the region,
+			// so --node-locations is only applied to Standard clusters.
+			if !testParams.UseGKEAutopilot {
+				nodeLocations = best.zone
+			}
+		} else {
+			klog.Warningf("All Capacity Advisor queries failed, falling back to default regional node allocation in %s", testParams.GkeClusterRegion)
+		}
+	}
+
+	// gcloud interprets --num-nodes as a per-zone count. When node-locations pins a
+	// single zone, numNodes is the total count across the cluster. Without pinned node-locations,
+	// GKE defaults to allocating nodes across all zones in the region (typically 3 zones),
+	// so scale numNodes down (minimum 1) to keep the total cluster node count consistent.
+	numNodes := testParams.NumNodes
+	if !testParams.UseGKEAutopilot && nodeLocations == "" {
+		numNodes = max(1, testParams.NumNodes/3)
+	}
+
+	klog.Infof("Attempting to create cluster %q in region %q (zone: %q)...", testParams.GkeClusterName, testParams.GkeClusterRegion, nodeLocations)
+
 	//nolint:gosec
-	out, err := gcloudCommand(testParams, "container", "clusters", "list", "--region", testParams.GkeClusterRegion, "--project", testParams.ProjectID, "--verbosity", "none", "--filter", "name="+testParams.GkeClusterName).CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("failed to check for previous test cluster: output: %v, err: %w", out, err)
+	out, listErr := gcloudCommand(testParams, "container", "clusters", "list", "--region", testParams.GkeClusterRegion, "--project", testParams.ProjectID, "--verbosity", "none", "--filter", "name="+testParams.GkeClusterName).CombinedOutput()
+	if listErr != nil {
+		return fmt.Errorf("failed to check for previous test cluster: output: %v, err: %w", out, listErr)
 	}
 	if len(out) > 0 {
 		klog.Infof("Detected previous cluster %s. Deleting it so a new one can be created...", testParams.GkeClusterName)
@@ -177,24 +265,8 @@ func clusterUpGKE(testParams *TestParameters) error {
 		cmdParams = append(cmdParams, "--cluster-version", testParams.GkeClusterVersion)
 	}
 
-	// Query Capacity Advisor to select a stockout-free zone for Standard clusters.
-	// gcloud interprets --num-nodes as a per-zone count. When the advisor pins a
-	// single zone, NumNodes is the total; on failure we fall back to GKE's default
-	// 3-zone regional layout, so scale NumNodes down (minimum 1) to keep the total consistent.
-	var nodeLocations string
-	if !testParams.UseGKEAutopilot && testParams.UseCapacityAdvisor {
-		advisedZone, err := queryCapacityAdvisedZone(testParams)
-		if err != nil {
-			klog.Warningf("Capacity Advisor query failed, falling back to default regional node allocation: %v", err)
-			testParams.NumNodes = max(1, testParams.NumNodes/3)
-		} else {
-			klog.Infof("Using Capacity Advised zone %q for cluster node locations", advisedZone)
-			nodeLocations = advisedZone
-		}
-	}
-
 	standardClusterFlags := []string{
-		"--num-nodes", strconv.Itoa(testParams.NumNodes), "--image-type", testParams.NodeImageType,
+		"--num-nodes", strconv.Itoa(numNodes), "--image-type", testParams.NodeImageType,
 		"--machine-type", testParams.NodeMachineType,
 		"--workload-pool", testParams.ProjectID + ".svc.id.goog",
 	}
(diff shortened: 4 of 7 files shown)


test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,116 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+
+	for i, region := range candidateRegions {
+		klog.Infof("Attempting to create cluster %q in region %q (%d/%d)...", testParams.GkeClusterName, region, i+1, len(candidateRegions))
+		testParams.GkeClusterRegion = region
+
+		//nolint:gosec
+		out, err := gcloudCommand(testParams, "container", "clusters", "list", "--region", testParams.GkeClusterRegion, "--project", testParams.ProjectID, "--verbosity", "none", "--filter", "name="+testParams.GkeClusterName).CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("failed to check for previous test cluster: output: %v, err: %w", out, err)
+		}
```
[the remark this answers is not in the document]
yaozile123: Done

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,116 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-	cmd = gcloudCommand(testParams, cmdParams...)
-	if err := runCommand("Starting e2e Cluster on GKE", cmd); err != nil {
-		return fmt.Errorf("failed to bring up kubernetes e2e cluster on GKE: %w", err)
-	}
+		cmd = gcloudCommand(testParams, cmdParams...)
+		// Retry in the next candidate region if cluster creation fails (e.g. stockout).
+		if err = runCommand("Starting e2e Cluster on GKE", cmd); err != nil {
+			klog.Warningf("Cluster creation in region %q (zone %q) failed: %v", region, nodeLocations, err)
+			continue
+		}
```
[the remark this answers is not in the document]
yaozile123: All the test cluster in prow will be cleaned up by janitor

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,116 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+		var nodeLocations string
+		if !testParams.UseGKEAutopilot && testParams.UseCapacityAdvisor {
+			advisedZone, err := queryCapacityAdvisedZone(testParams)
+			if err != nil {
+				klog.Warningf("Capacity Advisor query failed, falling back to default regional node allocation: %v", err)
+				numNodes = max(1, numNodes/3)
+			} else {
+				klog.Infof("Using Capacity Advised zone %q for cluster node locations", advisedZone)
+				nodeLocations = advisedZone
+			}
```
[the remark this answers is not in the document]
yaozile123: Done

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,116 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-	exclusionDuration, err := time.ParseDuration(testParams.GinkgoTimeout)
-	if err != nil {
-		klog.Warningf("failed to parse ginkgo timeout %q, using default 4h for maintenance exclusion: %v", testParams.GinkgoTimeout, err)
-		exclusionDuration = 4 * time.Hour
-	}
+		exclusionDuration, err := time.ParseDuration(testParams.GinkgoTimeout)
+		if err != nil {
+			klog.Warningf("failed to parse ginkgo timeout %q, using default 4h for maintenance exclusion: %v", testParams.GinkgoTimeout, err)
+			exclusionDuration = 4 * time.Hour
+		}
```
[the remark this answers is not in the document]
yaozile123: Done

test/e2e/utils/cluster.go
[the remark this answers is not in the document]
yaozile123: Added cleanup logic if cluster creation failed

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,120 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+		cmd = gcloudCommand(testParams, "container", "clusters", "update", testParams.GkeClusterName, "--region", testParams.GkeClusterRegion, "--project", testParams.ProjectID,
+			"--add-maintenance-exclusion-name", "no-upgrades-during-test",
+			"--add-maintenance-exclusion-start", startExclusionTime.Format(time.RFC3339),
+			"--add-maintenance-exclusion-end", startExclusionTime.Add(exclusionDuration).Format(time.RFC3339),
+			"--add-maintenance-exclusion-scope", "no_upgrades")
+		if err := runCommand("Updating Cluster with maintenance window", cmd); err != nil {
+			return fmt.Errorf("failed to update cluster with maintenance window: %w", err)
+		}
+
+		return nil
```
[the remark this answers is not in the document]
yaozile123: Done

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,124 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+	candidateRegions := []string{testParams.GkeClusterRegion}
+	if !testParams.EnableZB {
+		candidateRegions = []string{"us-east4", "us-east1", "us-west1", "us-west4", testParams.GkeClusterRegion}
 	}
-	if len(out) > 0 {
-		klog.Infof("Detected previous cluster %s. Deleting it so a new one can be created...", testParams.GkeClusterName)
-		if err := clusterDownGKE(testParams); err != nil {
-			return err
+
+	var err error
```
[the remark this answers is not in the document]
yaozile123: `testParams.GkeClusterRegion` are usually pinned to us-central1

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,124 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+		if err := runCommand("Updating Cluster with maintenance window", cmd); err != nil {
+			return fmt.Errorf("failed to update cluster with maintenance window: %w", err)
+		}
+
+		testParams.NumNodes = numNodes
+		return nil
 	}
 
-	return nil
+	return fmt.Errorf("failed to bring up kubernetes e2e cluster on GKE: %w", err)
```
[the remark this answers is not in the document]
yaozile123: Failed creations are already cleaned up in-place inside the loop

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,124 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to check for previous test cluster: output: %v, err: %w", out, err)
+	// Fall back across candidate regions to mitigate stockouts; Zonal Buckets
+	// are pinned to testParams.GkeClusterRegion which defaults to us-central1.
+	// us-east4, us-east1, and us-west1/4 are prioritized because they typically
+	// have lower quota and capacity contention than us-central1 in our test projects.
+	candidateRegions := []string{testParams.GkeClusterRegion}
+	if !testParams.EnableZB {
+		candidateRegions = []string{"us-east4", "us-east1", "us-west1", "us-west4", testParams.GkeClusterRegion}
 	}
```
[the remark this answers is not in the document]
yaozile123: Done, prioritized the custom region

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,124 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-		return fmt.Errorf("failed to update cluster with maintenance window: %w", err)
+		//nolint:gosec
+		cmd = gcloudCommand(testParams, "container", "clusters", "update", testParams.GkeClusterName, "--region", testParams.GkeClusterRegion, "--project", testParams.ProjectID,
+			"--add-maintenance-exclusion-name", "no-upgrades-during-test",
+			"--add-maintenance-exclusion-start", startExclusionTime.Format(time.RFC3339),
+			"--add-maintenance-exclusion-end", startExclusionTime.Add(exclusionDuration).Format(time.RFC3339),
+			"--add-maintenance-exclusion-scope", "no_upgrades")
+		if err := runCommand("Updating Cluster with maintenance window", cmd); err != nil {
+			return fmt.Errorf("failed to update cluster with maintenance window: %w", err)
+		}
```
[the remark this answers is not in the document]
yaozile123: It will be cleaned up at here https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/de0ae22291435e6700a0de9238f90fbc5fbf8d8c/test/e2e/utils/handler.go#L174-L178

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,124 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to check for previous test cluster: output: %v, err: %w", out, err)
+	// Fall back across candidate regions to mitigate stockouts; Zonal Buckets
+	// are pinned to testParams.GkeClusterRegion (which defaults to us-central1).
+	// For standard clusters, testParams.GkeClusterRegion is prioritized first,
+	// followed by candidate fallback regions with typically lower contention.
+	candidateRegions := []string{testParams.GkeClusterRegion}
+	if !testParams.EnableZB {
+		candidateRegions = append(candidateRegions, "us-east4", "us-east1", "us-west1", "us-west4")
 	}
```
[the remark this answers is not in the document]
yaozile123: Done

amacaskill:
For "I verified manually in a boskos project and successfully provisioned a 9-node n2-standard-8 cluster in us-east4 (us-east4-c) without stockouts" CAn you please check we have the necessary quota for 9 nodes of n2-standard-8 in all the zones in each region that it could land in. I have seen in teh past, that our prow projects don't have the correct quota for some regions or zones for a particular machine type. We should check this before we merge it (just check 1 project across all the zones). otherwise, the tests will fail when you land in that zone

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,128 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-		klog.Infof("Detected previous cluster %s. Deleting it so a new one can be created...", testParams.GkeClusterName)
-		if err := clusterDownGKE(testParams); err != nil {
-			return err
+	// Fall back across candidate regions to mitigate stockouts; Zonal Buckets
+	// are pinned to testParams.GkeClusterRegion (which defaults to us-central1).
+	// For standard clusters, testParams.GkeClusterRegion is prioritized first,
+	// followed by candidate fallback regions with typically lower contention.
+	candidateRegions := []string{testParams.GkeClusterRegion}
+	if !testParams.EnableZB {
+		for _, r := range []string{"us-east4", "us-east1", "us-west1", "us-west4"} {
```
amacaskill, line 164: How did we choose these? And what was the reasoning for keeping within us multi-region? Is it because you didn't want to change the way we create buckets?  If this is still causing stockouts, we might need to expand even further but for now I think we can try us- only. 

How did you come up with these region with us-?

```
gcloud compute machine-types list --filter="zone ~ ^us- AND name = n2-standard-8"
NAME           ZONE           CPUS  MEMORY_GB  DEPRECATED
n2-standard-8  us-central1-a  8     32.00
n2-standard-8  us-central1-b  8     32.00
n2-standard-8  us-central1-c  8     32.00
n2-standard-8  us-central1-f  8     32.00
n2-standard-8  us-central2-b  8     32.00
n2-standard-8  us-central2-a  8     32.00
n2-standard-8  us-central2-c  8     32.00
n2-standard-8  us-central2-d  8     32.00
n2-standard-8  us-west1-a     8     32.00
n2-standard-8  us-west1-b     8     32.00
n2-standard-8  us-west1-c     8     32.00
n2-standard-8  us-east1-b     8     32.00
n2-standard-8  us-east1-c     8     32.00
n2-standard-8  us-east1-d     8     32.00
n2-standard-8  us-east4-a     8     32.00
n2-standard-8  us-east4-b     8     32.00
n2-standard-8  us-east4-c     8     32.00
n2-standard-8  us-west2-c     8     32.00
n2-standard-8  us-west2-b     8     32.00
n2-standard-8  us-west2-a     8     32.00
n2-standard-8  us-west3-a     8     32.00
n2-standard-8  us-west3-b     8     32.00
n2-standard-8  us-west3-c     8     32.00
n2-standard-8  us-west4-c     8     32.00
n2-standard-8  us-west4-a     8     32.00
n2-standard-8  us-west4-b     8     32.00
n2-standard-8  us-east7-a     8     32.00
n2-standard-8  us-east7-b     8     32.00
n2-standard-8  us-east7-c     8     32.00
n2-standard-8  us-east5-c     8     32.00
n2-standard-8  us-east5-b     8     32.00
n2-standard-8  us-east5-a     8     32.00
n2-standard-8  us-south1-c    8     32.00
n2-standard-8  us-south1-a    8     32.00
n2-standard-8  us-south1-b    8     32.00
n2-standard-8  us-west8-c     8     32.00
n2-standard-8  us-west8-a     8     32.00
n2-standard-8  us-west8-b     8     32.00
```

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,128 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-	if !testParams.UseGKEAutopilot && testParams.UseCapacityAdvisor {
-		advisedZone, err := queryCapacityAdvisedZone(testParams)
-		if err != nil {
-			klog.Warningf("Capacity Advisor query failed, falling back to default regional node allocation: %v", err)
-			testParams.NumNodes = max(1, testParams.NumNodes/3)
-		} else {
-			klog.Infof("Using Capacity Advised zone %q for cluster node locations", advisedZone)
-			nodeLocations = advisedZone
+		createCmd := "create"
+		if testParams.UseGKEAutopilot {
```
amacaskill: Please add a comment for why we skip zone advisor for AP

test/e2e/utils/cluster.go
```diff
@@ -149,98 +150,128 @@ func clusterUpGKE(testParams *TestParameters) error {
...
-	}
+	var err error
 
-	cmdParams := []string{
-		"container", "clusters", createCmd, testParams.GkeClusterName,
-		"--region", testParams.GkeClusterRegion, "--quiet",
-		"--release-channel", testParams.GkeReleaseChannel,
-		"--project", testParams.ProjectID,
-	}
+	for i, region := range candidateRegions {
```
amacaskill: This loop iterates over candidate regions and invokes the gcloud container clusters create  command. If the GKE control plane or node provisioning hangs or slowly times out due to underlying resource exhaustion (stockout) (which it usually does after 50 min or so, then we have to wait for that each time). Even if it fails once, then runs on next region and succeeds,  the overall test will just timeout since we can't spare waiting an extra 50min to run tests since we are almost at the 4hr mark. We don't want to extend test time either .

This should be refactored to use a pre-flight capacity check. Before running any gcloud container clusters create command, use the Capacity Advisor or Zone Advisor APIs (e.g., gcloud  beta compute advice capacity ) to query obtainability scores or capacity availability for all candidate regions/zones.If the advisor explicitly returns "no capacity" or an obtainability score below a certain threshold for a specific region, the loop should immediately continue to the next candidate region before invoking the expensive cluster creation process.

test/e2e/utils/cluster.go
yaozile123: I ran `gcloud compute regions describe <region>` against our Boskos test project pool to inspect regional quota limits and found these specific regions had abundant N2 CPU quotas. And I agree we might need to expand further if it still causing stock outs

test/e2e/utils/cluster.go
yaozile123: Done

test/e2e/utils/cluster.go
yaozile123: Done, I have refactored the code to use pre-flight capacity check. The only concern is that Capacity Advisor evaluates SPOT availability, so we may still encounter on-demand stockouts despite high obtainability scores.

```
gcloud beta compute advice capacity \
  --region=us-central1 \
  --provisioning-model=SPOT \
  --size=9 \
  --instance-selection-machine-types=n2-standard-8 \
  --target-distribution-shape=any-single-zone \
  --format="value(recommendations[0].scores.obtainability,recommendations[0].shards[0].zone.basename())" \
  --zones=us-central1-a,us-central1-b,us-central1-c,us-central1-f \
  --project=yaozile-gke-dev
0.9	us-central1-c
```

yaozile123:
> For "I verified manually in a boskos project and successfully provisioned a 9-node n2-standard-8 cluster in us-east4 (us-east4-c) without stockouts" CAn you please check we have the necessary quota for 9 nodes of n2-standard-8 in all the zones in each region that it could land in. I have seen in teh past, that our prow projects don't have the correct quota for some regions or zones for a particular machine type. We should check this before we merge it (just check 1 project across all the zones). otherwise, the tests will fail when you land in that zone

I have verified that we have enough quota for n2-standard-8 for all the zones in our candidate regions
```
gcloud compute machine-types list \
  --project=prow-gob-internal-boskos-xxx \
  --filter="name=n2-standard-8 AND zone ~ ^us-" \
  --format="table(zone, name, guestCpus, memoryMb)"
```

amacaskill:
> I have verified that we have enough quota for n2-standard-8 for all the zones in our candidate regions
> gcloud compute machine-types list \
>   --project=prow-gob-internal-boskos-xxx \
>   --filter="name=n2-standard-8 AND zone ~ ^us-" \
>   --format="table(zone, name, guestCpus, memoryMb)"

This just means that project knows about the machine type but nothing about the project's ability to consume it (i.e., your quota ceiling or current usage). To check quota, you will need to use a different command and/or check in pantheon. Your looking for something like https://screenshot-v2.corp.google.com/6k1gpkiq28kjo, but it looks like you did the correct quota check in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1612#discussion_r3884135712. 

Please check for ARM also if you haven't

test/e2e/utils/cluster.go
```diff
@@ -149,10 +155,53 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+		for _, r := range []string{"us-east4", "us-east1", "us-west1", "us-west4"} {
+			if r != testParams.GkeClusterRegion {
+				candidateRegions = append(candidateRegions, r)
+			}
+		}
+	}
+
+	var nodeLocations string
+
+	// Skip Capacity Advisor for Autopilot because Autopilot automatically manages where nodes are created.
```
amacaskill, line 173: GKE Autopilot manages node placement within a region, but it relies on us to select the region itself. Should we let Autopilot utilize this Capacity Advisor loop to pick an unconstrained region? This would give Autopilot tests the same resilience against regional stockouts as Standard clusters, rather than pinning it to us-central1

test/e2e/utils/cluster.go
```diff
@@ -149,10 +155,53 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+				klog.Warningf("Capacity Advisor query failed for region %q: %v", region, err)
+				continue
+			}
+			klog.Infof("Capacity Advisor for region %q: zone=%q, obtainability=%.2f", region, zone, score)
+
+			// If score meets threshold, pick this region immediately.
+			if score >= minObtainabilityThreshold {
+				testParams.GkeClusterRegion = region
+				nodeLocations = zone
+				break
```
amacaskill, line 188: WE can add a comment here, explaining how testParams.GkeClusterRegion = region is later used to also select bucket location . 

+				// Note: Updating GkeClusterRegion here implicitly updates the 
+				// --test-bucket-location passed to Ginkgo. This ensures that the 
+				// GCS buckets created during tests are co-located in this fallback 
+				// region, preventing cross-region data transfer.

Please confirm this is true. ALso where we define GkeClusterRegion in the documentation, could also use a comment like this if it is true.

amacaskill: Additionally, we would want a short paragraph in  OSS documentation explaining how  the capacity checks works.

test/e2e/utils/cluster.go
amacaskill: Should we also consider adding some non-US regions (e.g., europe-west1 or asia-northeast1) to the candidateRegions fallback list? Because the tests inherently tie --test-bucket-location to the final testParams.GkeClusterRegion, landing the cluster in Europe will automatically provision a strictly regional GCS bucket there as well (IIUC please confirm(. This natively prevents any cross-region data pulls without us needing to write any extra logic, assuming our Boskos network/quota supports it, and will allow us to spread out resources more reducing stockouts.

test/e2e/utils/cluster.go
amacaskill: What you have now is a much cleaner approach, but moving away from the creation retry loop introduces a few new edge cases:

- Strict Thresholding: If the whole network is constrained and the max score across all regions happens to be 0.79, we immediately abandon the fallback list. But a 0.79 region still has a much better chance of succeeding than falling back to the default region which might have no chance of obtaining. 
- State Mutation: testParams.NumNodes = max(1, testParams.NumNodes/3) mutates the shared pointer. If the test framework ever re-invokes this function, the node count will unintentionally shrink again (e.g., 9 -> 3 -> 1. Can we fix this?  Can we use a local variable for the node count calculation instead of mutating testParams.NumNodes).

 Instead of using a hard break threshold, how long would it take to  query all candidate regions, sorting the candidate list by their Capacity Advisor scores (highest to lowest), and putting the cluster creation runCommand back inside a loop over this sorted list? I'm wondering if there was some reason why you didn't go with this approach instead?

yaozile123:
> This just means that project knows about the machine type but nothing about the project's ability to consume it (i.e., your quota ceiling or current usage). To check quota, you will need to use a different command and/or check in pantheon. Your looking for something like https://screenshot-v2.corp.google.com/6k1gpkiq28kjo, but it looks like you did the correct quota check in [#1612 (comment)](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1612#discussion_r3884135712).
> 
> Please check for ARM also if you haven't

I went ahead and checked the ARM (`T2A_CPUS`) quotas across all candidate regions in our Boskos project pool (`prow-gob-internal-boskos-xxx`) using `gcloud compute regions describe <region>`:
- `us-central1`: limit=300.0, usage=0.0
- `us-east4`: limit=300.0, usage=0.0
- `us-east1`: limit=300.0, usage=0.0
- `us-west1`: limit=300.0, usage=0.0
- `us-west4`: limit=300.0, usage=0.0

All candidate regions have 300 T2A CPUs with 0 usage, so there is sufficient quota headroom. 

I also noticed that we recently removed the ARM testgrid. Do we still run or plan to maintain ARM E2E tests?

test/e2e/utils/cluster.go
yaozile123: Done

test/e2e/utils/cluster.go
yaozile123: I have created a local `numNodes` variable to avoid mutating `testParams.NumNodes`. The Capacity Advisor query only takes ~5s for each region but cluster creation can still takes ~40 minutes before timing out, looping through cluster creations would easily hit the Prow timeout. We can query and sort all candidates to pick the highest-scoring region, or only attempt creation in the top 2 regions.

test/e2e/utils/cluster.go
yaozile123: I have added the comment and confirmed Ginkgo is passed `--test-bucket-location` directly from `testParams.GkeClusterRegion` in [handler.go](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1042c696e21d2f89bfcd61174f2c188e8514f53f/test/e2e/utils/handler.go#L373)

test/e2e/utils/cluster.go
yaozile123: I confirmed that `--test-bucket-location` directly takes `testParams.GkeClusterRegion` in [handler.go](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1042c696e21d2f89bfcd61174f2c188e8514f53f/test/e2e/utils/handler.go#L373)

I also checked the N2_CPUS quotas for europe-west1 and asia-northeast1 in our Boskos project pool using `gcloud compute regions describe <region>`:

europe-west1: limit=3000.0, usage=0.0
asia-northeast1: limit=1500.0, usage=0.0

Both regions have sufficient quota, so I added them to candidateRegions as well.

test/e2e/utils/cluster.go
yaozile123: Currently, both the [Ginkgo timeout](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1042c696e21d2f89bfcd61174f2c188e8514f53f/test/e2e/run-e2e-ci.sh#L33) and the Prow job timeout are configured to 4 hours. The test suite itself takes approximately 2.5 hours to run. If we want to iterate over cluster creations, it will likely require bumping the Prow job timeout to 4.5–5 hours.

test/e2e/utils/cluster.go
amacaskill: > Additionally, we would want a few sentences in OSS documentation explaining how the capacity checks works.

Please address this also

And 

> Please confirm this is true. ALso where we define GkeClusterRegion in the documentation, could also use a comment like this if it is true. If there is somewhere else we need

test/e2e/testsuites/profiles.go
```diff
@@ -538,14 +538,18 @@ func modifyPVForProfiles(testScenario string, pv *v1.PersistentVolume, sc string
...
 			va[anywhereCacheAdmissionPolicyKey] = "admit-on-second-miss"
 			isZBEnabled := os.Getenv(utils.IsZBEnabledEnvVar)
 			if isZBEnabled == "false" {
 				// AC is not supported for zb, so we only use AC for non zb tests.
-				va[anywhereCacheZonesKey] = "us-central1-c"
+				clusterRegion := os.Getenv(utils.ClusterLocationEnvVar)
+				if clusterRegion == "" {
+					clusterRegion = "us-central1"
+				}
+				va[anywhereCacheZonesKey] = clusterRegion + "-c"
```
amacaskill: This assumes that utils.ClusterLocationEnvVar always contains a region (e.g., us-central1). If this environment variable happens to contain a zone (e.g., us-central1-a), the resulting string will be malformed (e.g., us-central1-a-c).

What if ClusterLocationEnvVar is a zone?

test/e2e/utils/cluster.go
```diff
@@ -149,10 +155,62 @@ func clusterUpGKE(testParams *TestParameters) error {
...
+			}
+		}
+	}
+
+	var nodeLocations string
+	numNodes := testParams.NumNodes
+
+	// Use Capacity Advisor to select an unconstrained region for both Standard and Autopilot clusters.
+	if testParams.UseCapacityAdvisor {
+		const minObtainabilityThreshold = 0.8
```
amacaskill: what is the reasoning for doing this rather than just not picking the region with highest chance to obtain compute ?

test/e2e/utils/cluster.go
amacaskill: I'm not recommending iterating over cluster create, my suggestion is to query the Capacity API for all region, and create a stack ranked list, and try the one with the best chance  . You only try 1 cluster create, if it fails, fail the test.

test/e2e/utils/cluster.go
amacaskill: europe-west1 or asia-northeast1 was just a random example. Please add all the ones that you see are possible.

test/e2e/utils/cluster.go
yaozile123: I had some confusion earlier regarding whether we wanted a cluster creation retry or fail faster. I will update the code now

test/e2e/testsuites/profiles.go
yaozile123: In our CI script, [utils.ClusterLocationEnvVar](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1042c696e21d2f89bfcd61174f2c188e8514f53f/test/e2e/utils/handler.go#L191) is populated with testParams.GkeClusterRegion which is always a region.  But for local runs CLUSTER_LOCATION can be a zone like us-central1-a.

I will parse CLUSTER_LOCATION to strip the trailing zone suffix if it's a zone, ensuring we always extract the base region before appending -c.

test/e2e/utils/cluster.go
yaozile123: I checked the N2_CPUS quotas, us-central2, europe-west1, and europe-west3 have noticeably higher quotas (2200–3000 CPUs with 0 usage), while the remaining regions only have 1500 or less:

europe-west1: limit = 3000
europe-west3: limit = 3000
us-central2: limit = 2200
asia-east1: limit = 1500, usage = 0
asia-northeast1: limit = 1500, usage = 0

I have added these three regions us-central2, europe-west1, and europe-west3 to candidateRegions

test/e2e/utils/cluster.go
yaozile123: Done

test/e2e/utils/cluster.go
yaozile123: I have updated the code to picking the the region with highest scoring for creating the cluster

test/e2e/utils/cluster.go
yaozile123: Done. I [verified this behavior](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/blob/1042c696e21d2f89bfcd61174f2c188e8514f53f/test/e2e/utils/handler.go#L373), added comments to the definition sites in handler.go and main.go, and updated `test/README.md` with a section explaining the capacity checks and bucket co-location.

test/e2e/main.go
```diff
@@ -31,7 +31,7 @@ var (
 	pkgDir = flag.String("pkg-dir", "", "the package directory")
 
 	// Kubernetes cluster flags.
-	gkeClusterRegion               = flag.String("gke-cluster-region", "", "region that gke regional cluster should be created in")
+	gkeClusterRegion               = flag.String("gke-cluster-region", "", "region that gke regional cluster should be created in, which also sets --test-bucket-location for Ginkgo")
```
amacaskill, line 34: Thank you for adding this!

amacaskill:
> > This just means that project knows about the machine type but nothing about the project's ability to consume it (i.e., your quota ceiling or current usage). To check quota, you will need to use a different command and/or check in pantheon. Your looking for something like https://screenshot-v2.corp.google.com/6k1gpkiq28kjo, but it looks like you did the correct quota check in [#1612 (comment)](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1612#discussion_r3884135712).
> > Please check for ARM also if you haven't
> 
> I went ahead and checked the ARM (`T2A_CPUS`) quotas across all candidate regions in our Boskos project pool (`prow-gob-internal-boskos-xxx`) using `gcloud compute regions describe <region>`:
> 
> * `us-central1`: limit=300.0, usage=0.0
> * `us-east4`: limit=300.0, usage=0.0
> * `us-east1`: limit=300.0, usage=0.0
> * `us-west1`: limit=300.0, usage=0.0
> * `us-west4`: limit=300.0, usage=0.0
> 
> All candidate regions have 300 T2A CPUs with 0 usage, so there is sufficient quota headroom.
> 
> I also noticed that we recently removed the ARM testgrid. Do we still run or plan to maintain ARM E2E tests?

Ah you are correct. I forgot we removed arm. Thanks. We don't need arm then.

test/e2e/utils/cluster.go
amacaskill: thanks. Lets just remove us-central2 as that is a private region, and I'm not sure if we will run into other issues with it.. The rest of the regions looks good . Thanks for confirming.

test/README.md
```diff
@@ -130,7 +130,9 @@ When `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE` is set to `true`, the test runner provi
 - **Cluster Type**: Standard GKE Cluster (default) or Autopilot (if `E2E_TEST_USE_GKE_AUTOPILOT=true`).
 - **Release Channel**: `rapid` (can be overridden by `GKE_RELEASE_CHANNEL` environment variable).
 - **Version**: Determined by `E2E_TEST_GKE_CLUSTER_VERSION` (local) or `GKE_CLUSTER_VERSION` (CI).
-- **Region**: Determined by `GKE_CLUSTER_REGION` (default: `us-central1`).
+- **Region & Capacity Selection**: Determined by `GKE_CLUSTER_REGION` (default: `us-central1`).
+  - **Dynamic Fallback & Capacity Checks**: When `USE_CAPACITY_ADVISOR=true`, as enabled by default in Prow, the test runner pre-flights compute availability across candidate fallback regions using the Compute Engine Capacity Advisor API. It evaluates obtainability scores, stack-ranks candidate regions from highest to lowest score, and attempts cluster creation in the top-ranked region to mitigate regional stockouts.
```
amacaskill: Sorry I wasn't on the original PR, so also leaving comments on how previous PRs implemented things: 

In test/e2e/run-e2e-ci.sh , please move

```
readonly use_capacity_advisor=${USE_CAPACITY_ADVISOR:-true}
if [ "${use_capacity_advisor}" = true ]; then
  default_number_nodes=9
else
  default_number_nodes=3
fi
```

closer to where we create the cluster so we have the context that  gcloud interprets --num-nodes as per-zone, not per-cluster.

Currently, the Bash script has to guess the topology and pass 9 or 3 to the Go code, which is confusing. I think just keep num_nodes as the number of nodes you want to run across the cluster, then do the logic of how to make sure that happens, within the go code.  This is much better encapsulation.

test/README.md
```diff
@@ -130,7 +130,9 @@ When `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE` is set to `true`, the test runner provi
...
+  - **Bucket Co-location**: Selecting a fallback region also implicitly updates the `--test-bucket-location` passed to Ginkgo, ensuring test GCS buckets are co-located in the same region as the cluster to prevent cross-region data transfer.
 - **Node Pool Configuration** (Standard cluster only):
   - **Node Count**: 3 nodes (can be overridden by `E2E_TEST_NUM_NODES` or `NUMBER_NODES` env var).
```
amacaskill: Can you please fix: 
- 
- **Node Count**: 3 nodes (can be overridden by `E2E_TEST_NUM_NODES` or `NUMBER_NODES` env var). 

to say 9 nodes.

test/e2e/testsuites/profiles.go
```diff
@@ -538,14 +538,23 @@ func modifyPVForProfiles(testScenario string, pv *v1.PersistentVolume, sc string
...
+				clusterLocation := os.Getenv(utils.ClusterLocationEnvVar)
+				if clusterLocation == "" {
+					clusterLocation = "us-central1"
+				}
+				// If clusterLocation is a zone (e.g. "us-central1-a"), extract the region prefix ("us-central1").
+				parts := strings.Split(clusterLocation, "-")
+				if len(parts) > 2 && len(parts[len(parts)-1]) == 1 {
+					clusterLocation = strings.Join(parts[:len(parts)-1], "-")
+				}
+				va[anywhereCacheZonesKey] = clusterLocation + "-c"
```
amacaskill, line 559: Why are we doing this? If our compute/pod runs in us-central1-a, but anyc created in us-central1-c, we aren't going to be using the anyc, which defeats the purpose of the test. 

Can you please add a TODO here to instead create the ANyC in the same zone as the compute (or skip the test if the zone is not supported for AnyC), then please file a q4 fix it bug as a p2. We don't need to fix in this PR, but we should fix later. Thanks!

test/README.md
amacaskill: Can you please fix: 
- **Node Count**: 3 nodes (can be overridden by `E2E_TEST_NUM_NODES` or `NUMBER_NODES` env var). 

It should be 9.

test/README.md
amacaskill: Please explain somwhere that USE_CAPACITY_ADVISOR is only used when running  E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true. 

And please make sure this is true in the code. If you run on existing cluster, you want to ignore USE_CAPACITY_ADVISOR.

amacaskill:
Since we have made a lot of changes, please do another run with E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true to make sure the tests are working properly, then another run with E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=false on an existing cluster looks good.   You don't need to run all test cases, maybe just run 1-2 tests and confirm the logic for picking the cluster zone behaves as expected .

test/README.md
amacaskill: Also, consider the case is handled when we move the logic within the go code: 

1. USE_CAPACITY_ADVISOR=false + E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true (Local E2E)
If someone runs these E2E tests manually/locally without USE_CAPACITY_ADVISOR=true, the UseCapacityAdvisor logic is skipped entirely.

In this code path, numNodes remains its default value (e.g., 9), and --node-locations is left empty.
Because --region is used without pinning the zone, gcloud container clusters create silently multiplies --num-nodes by the number of zones in the region (typically 3).
Result: Local developers running the test suite will inadvertently spin up 9 * 3 = 27 nodes and burn through their quota super fast.

test/README.md
yaozile123: Done

yaozile123: Done

test/README.md
yaozile123: Done. Moved the node count calculation into Go and decoupled it from the Bash script.

yaozile123: Done, updated `test/README.md` to clarify this, and added a check in `test/e2e/main.go` to reset and ignore `--use-capacity-advisor` when running against an existing cluster

test/e2e/testsuites/profiles.go
yaozile123: Done, I have added a TODO and created the fix it bug

yaozile123:
> Since we have made a lot of changes, please do another run with E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true to make sure the tests are working properly, then another run with E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=false on an existing cluster looks good. You don't need to run all test cases, maybe just run 1-2 tests and confirm the logic for picking the cluster zone behaves as expected .

I have tested locally with 
```
E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=false \
USE_CAPACITY_ADVISOR=true \
E2E_TEST_PROJECT_ID=yaozile-gke-dev \
GKE_CLUSTER_REGION=us-central1 \
REGISTRY=ignored \
E2E_TEST_USE_GKE_MANAGED_DRIVER=true \
E2E_TEST_GINKGO_PROCS=1 \
E2E_TEST_FOCUS="Ephemeral-volume.*read-only.*should fail when write" \
E2E_TEST_SKIP="Dynamic PV|should.succeed.in.performance.test|oidc" \
./test/e2e/run-e2e-local.sh
```

The log shows that we are ignoring capacity advisor when testing on existing cluster, and 4 test cases passed
```
I0909 22:50:30.378098 171679 main.go:112] Ignoring --use-capacity-advisor because --manage-cluster-lifecycle is false (tests are running on an existing cluster)

Ran 4 of 703 Specs in 871.980 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 699 Skipped
PASS
```
and 
```
E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true \
USE_CAPACITY_ADVISOR=true \
E2E_TEST_PROJECT_ID=yaozile-gke-dev \
GKE_CLUSTER_REGION=us-central1 \
E2E_TEST_GKE_CLUSTER_VERSION=1.36 \
REGISTRY=ignored \
E2E_TEST_USE_GKE_MANAGED_DRIVER=true \
E2E_TEST_GINKGO_PROCS=1 \
E2E_TEST_FOCUS="Ephemeral-volume.*read-only.*should fail when write" \
E2E_TEST_SKIP="Dynamic PV|should.succeed.in.performance.test|oidc" \
./test/e2e/run-e2e-local.sh
```

The capcity advisor works well when specify `E2E_TEST_MANAGE_CLUSTER_LIFECYCLE=true` and all the testcases passed
```
I0909 22:19:51.676140  141292 cluster.go:209] Stack-ranked candidate regions:
I0909 22:19:51.676164  141292 cluster.go:211]   #1: region="us-central1" zone="us-central1-c" score=0.90
I0909 22:19:51.676187  141292 cluster.go:211]   #2: region="us-east4" zone="us-east4-c" score=0.90
I0909 22:19:51.676205  141292 cluster.go:211]   #3: region="us-east1" zone="us-east1-b" score=0.90
I0909 22:19:51.676222  141292 cluster.go:211]   #4: region="us-west1" zone="us-west1-b" score=0.90
I0909 22:19:51.676238  141292 cluster.go:211]   #5: region="us-west4" zone="us-west4-c" score=0.90
I0909 22:19:51.676252  141292 cluster.go:211]   #6: region="us-central2" zone="us-central2-c" score=0.90
I0909 22:19:51.676263  141292 cluster.go:211]   #7: region="europe-west1" zone="europe-west1-c" score=0.90
I0909 22:19:51.676279  141292 cluster.go:211]   #8: region="europe-west3" zone="europe-west3-b" score=0.90
I0909 22:19:51.676298  141292 cluster.go:213] Selecting highest-scoring region "us-central1" (zone: "us-central1-c", score: 0.90)

I0909 22:19:51.676315  141292 cluster.go:239] Attempting to create cluster "gcsfuse1846" in region "us-central1" (zone: "us-central1-c")...

Ran 4 of 703 Specs in 881.511 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 699 Skipped
PASS
```

amacaskill (approved):

test/e2e/utils/cluster.go
```diff
@@ -97,24 +98,24 @@ func queryRegionalStandardZones(testParams *TestParameters) string {
...
 	// We use SPOT as a proxy for regional resource availability.
 	// See: https://cloud.google.com/sdk/gcloud/reference/beta/compute/advice/capacity
 	cmdArgs := []string{
 		"beta", "compute", "advice", "capacity",
-		"--region=" + testParams.GkeClusterRegion,
+		"--region=" + region,
 		"--provisioning-model=SPOT",
 		"--size=" + strconv.Itoa(testParams.NumNodes),
 		"--instance-selection-machine-types=" + testParams.NodeMachineType,
 		"--target-distribution-shape=any-single-zone",
```
amacaskill, line 113: I don't really trust this command, the obtainability is always .90 then will just stockout anyways, which doesn't make sense: 

````
us-central1: 0.9        us-central1-c
us-east4: 0.9   us-east4-c
us-east1: 0.9   us-east1-b
us-west1: 0.9   us-west1-c
us-west4: 0.9   us-west4-c
europe-west1: 0.9       europe-west1-c
europe-west3: 0.9       europe-west3-b
```

PR #1620, opened 2026-09-03
[Shared Mount] In-Place Workload Pod Restart E2E Test

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds an E2E test to verify that in-place container restarts within a workload pod do not disrupt the GCSFuse mount in the shared node mount architecture:

1. Deploys a workload pod referencing a shared-mount PVC with container `restartPolicy: Always`.
2. Confirms that both the workload pod and the external Mounter Pod are created and running with 0 restarts.
3. Writes a pre-restart file to the shared mount and verifies contents.
4. Triggers an in-place container restart by sending `SIGTERM` to PID 1 inside the workload container.
5. Verifies that Kubelet restarts the workload container in-place (restart count increments to 1) while preserving Pod UID and Node assignment.
6. Verifies that the external Mounter Pod remains untouched with 0 restarts.
7. Verifies that the restarted container immediately resumes reading pre-existing data and writing new data without errors.

NONE
NONE

```release-note
NONE
```

### DIFF
--- a/test/e2e/specs/specs.go
+++ b/test/e2e/specs/specs.go
@@ -20,6 +20,7 @@ package specs
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -37,11 +38,13 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/kubernetes/pkg/kubelet/events"
 	"k8s.io/kubernetes/test/e2e/framework"
@@ -230,9 +233,10 @@ func GetMounterPod(ctx context.Context, c clientset.Interface, namespace string,
 }
 
 type TestPod struct {
-	client    clientset.Interface
-	pod       *corev1.Pod
-	namespace *corev1.Namespace
+	client                   clientset.Interface
+	pod                      *corev1.Pod
+	namespace                *corev1.Namespace
+	inplacePodRestartEnabled bool
 }
 
 func NewTestPodModifiedSpec(c clientset.Interface, ns *corev1.Namespace, setAutomountServiceAccountToken bool) *TestPod {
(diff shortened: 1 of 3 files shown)


test/e2e/testsuites/shared_mount.go
```diff
@@ -649,4 +649,79 @@ func (t *gcsFuseCSISharedMountTestSuite) DefineTests(driver storageframework.Tes
...
+		ginkgo.By("Verifying the external Mounter Pod remains untouched with 0 restarts")
+		updatedMounterPod := specs.GetMounterPod(ctx, f.ClientSet, f.Namespace.Name, nodeName)
+		gomega.Expect(updatedMounterPod.UID).To(gomega.Equal(initialMounterPod.UID), "expected mounter pod UID to be unchanged")
+		verifyMounterPodNoRestarts(updatedMounterPod)
+
+		ginkgo.By("Verifying the restarted container immediately resumes reading pre-existing data and writing new data without errors")
+		workloadPod.VerifyRWMount(f, sharedMountPath)
+		workloadPod.VerifyReadFile(f, preRestartFile, preRestartContent)
+
+		postRestartFile := fmt.Sprintf("%s/post-restart-data.txt", sharedMountPath)
```
uriel-guzman: Instead of writing a file before restart, restarting, and then writing a file after restart (all non-intersecting validations), we should make the container write continuous I/O and verify it never runs into an error. The current test just validates that the mount continues to be functional after an in-place restart, but it doesn't verify that the I/O remains functional **during** the in-place restart.

chrisThePattyEater: That makes sense, I updated to continously write:
workloadPod.SetCommand(fmt.Sprintf("trap 'exit 88' 15; while true; do echo \"$(date) - heartbeat\" >> %s; sleep 2; done", heartbeatFile))

We now check before restart to confirm that writing has occured and then we check after to ensure that it keeps writing. Then we check to verify there are no error logs.

test/e2e/specs/specs.go
```diff
@@ -304,11 +308,75 @@ func NewTestPod(c clientset.Interface, ns *corev1.Namespace) *TestPod {
...
-	t.pod, err = t.client.CoreV1().Pods(t.namespace.Name).Create(ctx, t.pod, metav1.CreateOptions{})
-	framework.ExpectNoError(err)
+	if !t.inplacePodRestartEnabled {
+		var err error
+		t.pod, err = t.client.CoreV1().Pods(t.namespace.Name).Create(ctx, t.pod, metav1.CreateOptions{})
+		framework.ExpectNoError(err)
+		return
+	}
+
+	// TODO(k8s-1.35): When k8s.io/api and client-go in test/go.mod are upgraded to >= v0.35.0,
```
uriel-guzman: Can you create a bug for this? Otherwise it will be forgotten after this PR is merged.

test/e2e/testsuites/shared_mount.go
```diff
@@ -649,4 +650,117 @@ func (t *gcsFuseCSISharedMountTestSuite) DefineTests(driver storageframework.Tes
...
+			g.Expect(count).To(gomega.BeNumerically(">", preRestartCount),
+				fmt.Sprintf("expected heartbeat count (%d) to increase beyond pre-restart count (%d)", count, preRestartCount))
+		}, "30s", "2s").Should(gomega.Succeed())
+
+		ginkgo.By("Verifying that no errors occurred during container I/O operations")
+		logs, err := e2epod.GetPodLogs(ctx, f.ClientSet, f.Namespace.Name, workloadPod.GetPodName(), specs.TesterContainerName)
+		framework.ExpectNoError(err, "failed to get workload container logs")
+		gomega.Expect(strings.ToLower(logs)).NotTo(gomega.ContainSubstring("error"), "expected no errors in workload container logs")
+
+		postRestartFile := fmt.Sprintf("%s/post-restart-verify.txt", mountPath)
```
uriel-guzman: Why do we need a post-restart file? Isn't the heartbeat file continuous I/O? Meaning that it r/w before, during, and after restart? So, why do we have a separate post-restart file?

test/e2e/specs/specs.go
chrisThePattyEater: added

test/e2e/testsuites/shared_mount.go
chrisThePattyEater: I thought it would be good to create a new file after restarting but you are right it doesn't add much, removed this

uriel-guzman (approved):

PR #1631, opened 2026-09-16
Fix Istio REGISTRY_ONLY E2E test hang for gRPC client protocol

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test
> /kind feature
> /kind flake

Previously, we excluded port 443 from Istio sidecar interception only for Zonal Buckets (`zbEnabled`), because Istio blocks gRPC connections in `REGISTRY_ONLY` mode and causes gcsfuse mount to hang. Regional gRPC tests passed because gcsfuse would fall back to HTTP. Recently, gcsfuse removed the HTTP fallback when `client-protocol=grpc` is explicitly set. As a result, regional gRPC hits the exact same Istio blackhole issue as Zonal Buckets. 

I have updated the annotation check from `zbEnabled(driver)` to `zbEnabled(driver) OR getClientProtocol(driver) == "grpc"`, so this covers both ZB and gRPC test suites.

Fixes #

I have tested locally on 1.37 cluster for 5 times, and they all passed:
```
export REGISTRY=us-central1-docker.pkg.dev/yaozile-gke-dev/gcs-fuse-csi-driver
export STAGINGVERSION=istio-grpc-1

make build-image-and-push-multi-arch \
    REGISTRY=$REGISTRY \
    BUILD_GCSFUSE_FROM_SOURCE=true

REGISTRY=us-central1-docker.pkg.dev/yaozile-gke-dev/gcs-fuse-csi-driver \
STAGINGVERSION=istio-grpc-1 \
GCSFUSE_CLIENT_PROTOCOL=grpc \
E2E_TEST_FOCUS="should.store.data.with.istio.registry.only.outbound.traffic.policy.mode" \
make e2e-test

Ran 4 of 712 Specs in 37.318 seconds
  SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 708 Skipped
  Ginkgo ran 1 suite in 44.435522495s
  Test Suite Passed
```

```release-note
Fix Istio REGISTRY_ONLY E2E test hang for gRPC client protocol
```


test/e2e/testsuites/istio.go
```diff
@@ -113,7 +113,9 @@ func (t *gcsFuseCSIIstioTestSuite) DefineTests(driver storageframework.TestDrive
 			specs.DeployIstioServiceEntry(f.Namespace.Name)
 		}
 
-		if zbEnabled(driver) {
+		// When using gRPC, Istio blocks outbound connections to GCS and causes gcsfuse to hang on mount.
+		// Excluding port 443 lets gRPC traffic bypass the Istio sidecar proxy.
+		if getClientProtocol(driver) == "grpc" {
```
amacaskill: Please do if zbEnabled(driver) OR getClientProtocol(driver) == "grpc" just to be extra safe in case we change something in the future

amacaskill (approved):

test/e2e/testsuites/istio.go
yaozile123: Done

PR #1640, opened 2026-09-21
Avoid the 1.5s stagger delay before the first gcsfuse mount

### LINKED ISSUE #1639: Sidecar mounter sleeps 1.5s before the first mount
The sidecar mounter sleeps 1.5s at the top of the socket loop, so you pay it before the first gcsfuse even starts:

```go
for _, sp := range socketPaths {
	...
```

With a single volume there's nothing to stagger against, but the pod still starts 1.5s later. With N volumes it's N × 1.5s.

The comment says the sleep is there to keep logs from mixing and to spread the memory peak. Neither applies before the first process starts, so I think it can just be skipped on the first iteration. PR coming.

Two things I noticed while looking at this:

- `Mount()` starts gcsfuse in a goroutine and returns right away, so the mounts are concurrent regardless. The sleep spaces out the launches but doesn't cap how many run at once.
- If log mixing is still a concern, a lock won't help for stdout: `cmd.Stdout = os.Stdout` hands the fd straight to the child, so it's separate processes writing to it. Only stderr goes through a copy goroutine in the sidecar. Prefixing lines with the volume name would probably work better than a time gap anyway.

Would you also take a patch making the interval configurable? A flag on its own wouldn't help on GKE since the webhook hardcodes the sidecar args, so it'd need an annotation.

otto-sellerstam:
**What type of PR is this?**

/kind bug

The sleep is at the top of the socket loop, so it runs before the first gcsfuse starts. A pod with one volume waits 1.5s for nothing.

The comment says the sleep keeps logs from mixing and spreads the memory peak. Neither applies before the first process, so this skips it on the first iteration. Consecutive launches are still spaced out as before.

Fixes #1639

Kept this to the one-line fix. The issue has some notes on making the interval configurable and on the logging, happy to follow up on either.

```release-note
Removed a 1.5 second delay from gcsfuse sidecar startup. The sidecar mounter no longer waits before launching the first gcsfuse process.
```


otto-sellerstam:
Will sign CLA once a [GoogleCloudPlatform](https://github.com/orgs/GoogleCloudPlatform/people) member has verified the patch.

ChaitanyaKN:
Hi Otto, this is already being done in https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/pull/1638

otto-sellerstam:
@ChaitanyaKN Oh wow I checked when I started on this... Funny coincidence!

Thanks for letting me know! Will close the PR and issue.

